PR #3, opened 2019-02-09
Crystal and more

lelongg:
This is a pretty big update and unfortunately it might be a bit too hard to review.

It adds :
- ROS2 Crystal support
- support for nested messages
- support for arrays in message
- generated RCL binding


esteve:
@lelongg whoa, this is so great, thank you so much! Do you think you could split this into smaller and more focused pull requests? Given that your changes started in December, I think it's doable. It'll make it much easier for me to review, I'll make a few comments in this PR, though.

README.md
```diff
@@ -1,65 +1,62 @@
...
 
-### Publisher and subscriber
+## Quickstart
 
-Publisher:
+### Build
 
+```bash
+source /opt/ros/crystal/setup.bash
+colcon build --symlink-install --merge-install --cmake-args -G Ninja
```
esteve: There's no need for Ninja, the default builder should work. Have you tried this with an isolated workspace? (i.e. without --merge-install)

common_interfaces/.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,45 @@
+<!--
```
esteve: Remove all the message packages.

NOTICE
```diff
@@ -1,5 +1,7 @@
 ros2_rust
 Copyright 2018 Esteve Fernandez (esteve@apache.org)
+Copyright 2019 Easymov Robotics (dev@easymov.fr)
```
esteve: I don't think it's reasonable for a first contribution to add yourself as an author. I'm more than happy to add you after a few more contributions, that's how opensource works.

rcl_sys/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+cmake_minimum_required(VERSION 3.5)
```
esteve: Given that `rclrs` is deeply tied to `rcl`, I don't think there's a need for a separate `rcl-sys` package. Moreover, while `bindgen` is pretty cool, when I first wrote the Rust bindings I decided to go the `libc` route because that gave me more control over how to expose `rcl` to Rust.

rclrs/package.xml
```diff
@@ -2,54 +2,18 @@
 <package format="2">
   <name>rclrs</name>
   <version>0.0.3</version>
-  <description>Package containing the Rust client.</description>
-  <author email="esteve@apache.org">Esteve Fernandez</author>
-  <maintainer email="esteve@apache.org">Esteve Fernandez</maintainer>
+  <description>Common files for the Rust client.</description>
+  <maintainer email="dev@easymov.fr">Easymov Robotics</maintainer>
+  <author email="gerald.lelong@easymov.fr">Gérald Lelong</author>
```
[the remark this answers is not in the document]
lelongg: Sorry, I started with a full rewrite (just keeping the crate name) before choosing to stick with your initial API. Obviously I will fix this :)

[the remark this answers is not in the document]
esteve: No worries :-)

rcl_sys/CMakeLists.txt
lelongg: rclrs is indeed deeply tied to rcl but I think that the inverse relation is not necessarily true.
Having rcl-sys as a separate crate might allow for different client interface, for example.
Besides, the two crates (rcl-sys and rclrs) work on two well separated level of abstraction and I feel more confortable with keeping the two separate.

I liked how bindgen reduced the volume of Rust code to maintain in the original implementation, plus the lack of C code to write. I feel like it's a whole burden removed from rclrs so I'm interested in more detailed reasons on why not to use it.
What parts of rcl would you like not to expose to Rust and why ?

common_interfaces/.github/ISSUE_TEMPLATE.md
lelongg: Unfortunately there is an issue with the IDL generation that requires fixing the packages containing message definition and I don't know how to fix this yet (see [https://answers.ros.org/question/315189/fail-at-ros2-idl-generation/](https://answers.ros.org/question/315189/fail-at-ros2-idl-generation/)).

Moreover, when this issue is fixed, message packages might still be required, at least for the example to compile since the ROS2 IDL generation for third party generators will only trigger for messages defined in the current workspace.

esteve: The current version of `ros2_rust` already supports building messages without importing them or making any changes, as long as they are part of the workspace. Sure, it'd be great if in the future out-of-workspace messages work too, but adding the messages to the `ros2_rust` repo is just unmaintainable.

rcl_sys/CMakeLists.txt
esteve: > Having rcl-sys as a separate crate might allow for different client interface, for example.

Given that the only thing that `rcl-sys` does is generating bindings at compile time from `rcl/rcl.h` and `rclutils/error_handling.h`, why not add it as part of `rclrs`? It's not like it's doing much else. I don't think there will be alternative clients based on `rcl-sys`, mostly because it doesn't do much. Moreover, have a look at all the other client libraries: `rclpy`, `rclcpp`, `rcljava`, etc. the code that glues `rcl` and the client APIs are in the wrappers.

> What parts of rcl would you like not to expose to Rust and why?

Allocators, for example. They should be implemented natively in Rust and not exposed via C.

lelongg: > why not add it as part of `rclrs`?

I see your point, I agree that rcl-sys is now empty enough to be merged with rclrs and maybe not that useful on is own. My main motive to split it in the first place was to follow the Rust community guidelines about low level wrapper.

> Allocators, for example. They should be implemented natively in Rust and not exposed via C.

bindgen allows to blacklist specific structs and functions from the generated interface. Do you mean that rcl_get_default_allocator should be blacklisted ? I'm not sure to understand the issue raised by allocators, can you elaborate ?

lelongg:
Most concerns should have been addressed.
Splitting in separate meaningful PR might be difficult since this repository cannot be built in its current state (due to changes in ROS2 dependencies) and there is not much change I can easily remove from this PR and still making it compile.

esteve:
@lelongg thank you so much for taking the time to address my feedback. Don't worry about splitting up the PR, it'd make it easier sure, but the work you've done is excellent. Let me have a couple of days to review it all and if I don't get back to you, feel free to ping me again. Thanks!

tprk77:
Hi @lelongg, I've been trying out your branch. It's very cool and the examples are working great.

But I've run into a pretty serious issue. It seems that iterative builds are using stale source files. I am having to delete the build directory every time I want to compile.

lelongg:
@tprk77 Yes I noticed that too and it's very inconvenient but I haven't solved it yet.

@esteve I've found issues in array generation when working on this project today and I expect to find some more. Will it interfere with your review if I keep adding fixes to this PR? Alternatively, I can try to remove changes to the generator from this PR and fix it separately. It might lighten this PR a bit.

esteve:
@lelongg I think it'd make this PR easier to review if the generator stuff were in a separate PR, but it's your call, whatever works best for you.

tprk77:
@lelongg Sounds good. I am all for getting this in and fixing the issues later.

I've been trying this out some more, and I have one more comment. It seems there are a few types like `Node` which contain `Rc` making them unable to be `Send` or `Sync`. I think it might be useful in some situations to move types like `Node`, `Subscription`, and `Publisher` to other threads. It seems like the underlying `rcl` should(?) be capable of supporting that, but I don't really know. You would know better than me about the feasibility.

Again, maybe something to consider later on. :grin:

rclrs_examples/src/rclrs_publisher.rs
```diff
@@ -1,22 +1,26 @@
...
-        node.create_publisher::<std_msgs::msg::String>("topic", rclrs::qos::QOS_PROFILE_DEFAULT);
+        node.create_publisher::<std_msgs::msg::String>("topic", rclrs::QOS_PROFILE_DEFAULT)?;
 
-    let mut message: std_msgs::msg::String = Default::default();
+    let mut message = std_msgs::msg::String::default();
 
     let mut publish_count: u32 = 1;
 
-    while rclrs::ok() {
+    while context.ok() {
```
esteve: I'd keep rclrs:::ok for symmetry with the other client libraries.

rclrs_examples/src/rclrs_publisher.rs
```diff
@@ -1,22 +1,26 @@
...
+        std::thread::sleep(std::time::Duration::from_millis(500));
```
esteve: The examples for the other client libraries have the sleep call after publishing the message.

rclrs_examples/src/rclrs_subscriber.rs
```diff
@@ -1,19 +1,20 @@
...
+use std_msgs;
 
 fn topic_callback(msg: &std_msgs::msg::String) {
     println!("I heard: '{}'", msg.data);
 }
 
-fn main() {
-    rclrs::init().unwrap();
+fn main() -> rclrs::RclResult {
+    let context = rclrs::Context::default();
```
esteve: rclrs::init() like the other examples.

rosidl_generator_rs/msg/Bool.msg
```diff
@@ -1 +0,0 @@
-bool empty_bool
```
esteve: Why remove the test messages for the generator?

rclrs/package.xml
```diff
@@ -7,49 +7,12 @@
...
   <buildtool_depend>ament_cmake_export_crates</buildtool_depend>
-  <buildtool_depend>rclrs_common</buildtool_depend>
 
-  <buildtool_export_depend>ament_cmake_ros</buildtool_export_depend>
-  <buildtool_export_depend>ament_cmake_export_crates</buildtool_export_depend>
-  <buildtool_export_depend>rclrs_common</buildtool_export_depend>
-
-  <build_depend>builtin_interfaces</build_depend>
-  <build_depend>rcl_interfaces</build_depend>
-  <build_depend>rcl</build_depend>
```
esteve: Even if you don't link rcl and the rest of the libraries via CMake, you still need to mark them as a dependency to let colcon know that they must be available at build time.

esteve:
@tprk77 I'd rather get the pull request fixed or the changes moved to another pull request. This repository works fine right now (albeit only with Bouncy), I'd prefer not to introduce any changes that contain bugs.

rclrs/src/node/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+    where
+        T: rclrs_common::traits::MessageDefinition<T> + Default,
+    {
+        let subscription = Rc::new(Subscription::<T>::new(self, topic, qos, callback)?);
+        self.subscriptions
+            .push(Rc::downgrade(&subscription) as Weak<SubscriptionBase>);
+        Ok(subscription)
+    }
+
+    pub fn spin(&self) -> RclResult {
```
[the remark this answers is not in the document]
lelongg: Moving spin and spin_once to Node allowed to keep Node's fields private but it's true that it prevent multiple node spin, so I will revert.

rclrs/package.xml
lelongg: Resolved in 3f3a205

rclrs_examples/src/rclrs_publisher.rs
lelongg: Resolved in c55dd16.

rclrs/src/node/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+                    RclError::Timeout => continue,
+                    _ => return Err(error),
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    pub fn spin_once(&self, timeout: i64) -> RclResult {
```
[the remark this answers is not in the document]
lelongg: Resolved in f657ce1.

rosidl_generator_rs/msg/Bool.msg
lelongg: I thought they weren't used. Can we depend on the test_msgs package instead ?

rclrs_examples/src/rclrs_publisher.rs
lelongg: From [https://index.ros.org//doc/ros2/Releases/Release-Crystal-Clemmys/](https://index.ros.org//doc/ros2/Releases/Release-Crystal-Clemmys/).

> However, you may just continue to offer a single, global init and shutdown in your client library, and just store a single global context object.

As it is optional I was thinking about adding the global context object later.

rclrs/src/node/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+    pub fn create_subscription<T>(
+        &mut self,
+        topic: &str,
+        qos: QoSProfile,
+        callback: fn(&T),
+    ) -> RclResult<Rc<Subscription<T>>>
+    where
+        T: rclrs_common::traits::MessageDefinition<T> + Default,
+    {
+        let subscription = Rc::new(Subscription::<T>::new(self, topic, qos, callback)?);
```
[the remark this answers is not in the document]
lelongg: I thought it was closer to the rclcpp implementation as I believe publishers and subscribers hold a shared_ptr to the node. Do you suggest to have node's lifetime depending on the lifetime of the context too ?

lelongg:
@tprk77 the iterative compilation issue should be addressed by #7.

rosidl_generator_rs/msg/Bool.msg
esteve: The `test_msgs` package depends on the generator to be installed before getting there (i.e. you can't have `test_msgs` built unless you have built the generators). However when you're testing the generator itself, you need to have the messages available at test time within the same package. The Python and C++ generators have the same issue and that's why they have a collection of messages bundled.

rclrs_examples/src/rclrs_publisher.rs
esteve: It's fine if the context is not global for the time being, we can address that in the future. What I meant is that all the client libraries have an 'init` function that optionally take a context (or use the global one implicitly) and all functions are attached to the root namespace (e.g. rclcpp::init, rclpy.init, etc.)

Moreover, the concept of the context as defined in ROS2 differs slightly from what's presented in this pull request. It's not the context that calls `rcl_init`, but the opposite:

https://github.com/ros2/rcl/blob/master/rcl/src/rcl/init.c#L59

so in order to have mimic the other client libraries, we need to have a `rclrs::init` function that accepts a context and all functions are done at the rclrs level (e.g. spin,  ok, etc.)

rclrs/src/node/mod.rs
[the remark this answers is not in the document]
esteve: That's because C++ doesn't have lifetimes ;-)

Don't worry, just add a TODO comment stating that this will eventually be replaced by lifetimes and we'll address that in a subsequent pull request.

rosidl_generator_rs/msg/Bool.msg
lelongg: That makes sense ! I'll add it back and try to add a test build for these messages inside rosidl_generator_rs.

rclrs/src/node/mod.rs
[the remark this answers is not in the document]
lelongg: Resolved in 5589047

rclrs_examples/src/rclrs_publisher.rs
lelongg: > Moreover, the concept of the context as defined in ROS2 differs slightly from what's presented in this pull request. It's not the context that calls rcl_init, but the opposite

It looks like rclcpp context calls rcl_init.

[https://github.com/ros2/rclcpp/blob/a54a32915313960497bda2d601cdf98e9f2e9bf0/rclcpp/src/rclcpp/context.cpp#L89](https://github.com/ros2/rclcpp/blob/a54a32915313960497bda2d601cdf98e9f2e9bf0/rclcpp/src/rclcpp/context.cpp#L89)

> in order to have mimic the other client libraries, we need to have a rclrs::init function that accepts a context and all functions are done at the rclrs level

rclpy.init does take an optional context but rclcpp::init does not.

[https://github.com/ros2/rclcpp/blob/e32d17980dcf761712cb854b3703336c380a570e/rclcpp/src/rclcpp/utilities.cpp#L28](https://github.com/ros2/rclcpp/blob/e32d17980dcf761712cb854b3703336c380a570e/rclcpp/src/rclcpp/utilities.cpp#L28)

How much of a requirement do you think it is to mimic another client interface ? I think it is an important question to settle since it can considerably restrain the design space in the future.

rcl_sys/CMakeLists.txt
esteve: I see pros and cons to both approaches, the biggest issue for bindgen from my point of view is having to install clang for parsing the header files, but it's not such a big deal. In any case, I'm ok with switching to bindgen.

rclrs_examples/src/rclrs_publisher.rs
esteve: > How much of a requirement do you think it is to mimic another client interface ? I think it is an important question to settle since it can considerably restrain the design space in the future.


I'd say it's pretty important that rclrs follows the other client libraries, if only to help users switch to Rust if they know other ROS2 libraries. There's only a handful of methods common to all the libraries (i.e. init, createX, destroyX), so apart from that, we have enough freedom to implement the API in any way we want.

esteve: @lelongg did you have time to look into making the `rclrs` API similar to the other clients'? Thanks.

lelongg: I didn't make the change yet. What API do you suggest ? We can mimic the rclcpp API by making rclrs::ok take a context object but rclcpp::init does not and so we cannot add rclrs::init until we get the global context.

esteve:
Let's better merge this now and address anything else later. Thanks @lelongg !

PR #7, opened 2019-03-05
copy crate at build time

lelongg:
Addressing [https://github.com/esteve/ros2_rust/pull/3#issuecomment-468972443](https://github.com/esteve/ros2_rust/pull/3#issuecomment-468972443).

rclrs_examples crate source files are now copied at build time with CMake.

### DIFF
--- a/rclrs_examples/CMakeLists.txt
+++ b/rclrs_examples/CMakeLists.txt
@@ -39,14 +39,12 @@ endforeach()
 
 include(ExternalProject)
 
-add_custom_target(copy_crate ALL
-    COMMENT "Copying crate from ${CMAKE_SOURCE_DIR}/src to ${CMAKE_BINARY_DIR}/src"
-)
+configure_file(${CMAKE_SOURCE_DIR}/src/rclrs_publisher.rs ${CMAKE_BINARY_DIR}/src/rclrs_publisher.rs COPYONLY)
+configure_file(${CMAKE_SOURCE_DIR}/src/rclrs_subscriber.rs ${CMAKE_BINARY_DIR}/src/rclrs_subscriber.rs COPYONLY)
 
-add_custom_command(
-    TARGET copy_crate
-    COMMAND ${CMAKE_COMMAND} -E copy_if_different "${CMAKE_SOURCE_DIR}/src/rclrs_publisher.rs" "${CMAKE_BINARY_DIR}/src/rclrs_publisher.rs"
-    COMMAND ${CMAKE_COMMAND} -E copy_if_different "${CMAKE_SOURCE_DIR}/src/rclrs_subscriber.rs" "${CMAKE_BINARY_DIR}/src/rclrs_subscriber.rs"
+add_custom_target(
+    copy_crate
+    DEPENDS ${CMAKE_BINARY_DIR}/src/rclrs_publisher.rs ${CMAKE_BINARY_DIR}/src/rclrs_subscriber.rs
 )
 
 file(MAKE_DIRECTORY "${CMAKE_BINARY_DIR}/.cargo")
@@ -77,5 +75,4 @@ install(FILES
     PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
     DESTINATION lib/${PROJECT_NAME}
 )
-
 ament_package()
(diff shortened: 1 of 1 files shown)


rclrs_examples/CMakeLists.txt
```diff
@@ -39,8 +39,15 @@ endforeach()
 
 include(ExternalProject)
 
-file(REMOVE_RECURSE "${CMAKE_BINARY_DIR}/src")
-file(COPY "${CMAKE_SOURCE_DIR}/src" DESTINATION "${CMAKE_BINARY_DIR}")
+add_custom_target(copy_crate ALL
+    COMMENT "Copying crate from ${CMAKE_SOURCE_DIR}/src to ${CMAKE_BINARY_DIR}/src"
+)
+
+add_custom_command(
```
esteve: Wouldn't this force `cargo` to rebuild the .rs file every time, even if they are not updated? Given that the number of examples won't change much over time, I'd suggest using `configure_file` instead, and copy files one by one (https://cmake.org/cmake/help/v3.8/command/configure_file.html), something like this:

```cmake
configure_file(${CMAKE_SOURCE_DIR}/src/rclrs_publisher.rs ${CMAKE_BINARY_DIR}/src/rclrs_publisher.rs COPYONLY)

configure_file(${CMAKE_SOURCE_DIR}/src/rclrs_subscriber.rs ${CMAKE_BINARY_DIR}/src/rclrs_subscriber.rs COPYONLY)
```

the advantage of `configure_file` over `file(COPY ...)` is that it creates a file dependency and thus, whenever the input file is updated, CMake is retriggered.

rclrs_examples/CMakeLists.txt
```diff
@@ -45,8 +45,8 @@ add_custom_target(copy_crate ALL
 
 add_custom_command(
     TARGET copy_crate
-    COMMAND ${CMAKE_COMMAND} -E remove_directory "${CMAKE_BINARY_DIR}/src"
-    COMMAND ${CMAKE_COMMAND} -E copy_directory "${CMAKE_SOURCE_DIR}/src" "${CMAKE_BINARY_DIR}/src"
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different "${CMAKE_SOURCE_DIR}/src/rclrs_publisher.rs" "${CMAKE_BINARY_DIR}/src/rclrs_publisher.rs"
```
esteve: This will spawn a new process, whereas `configure_file` runs within the same CMake process. Moreover, with `configure_file` other targets can depend directly on the files, so you don't need a custom command. In general, `add_custom_command` should be used as the last resort and use the functions and macros provided by CMake.

lelongg: Sorry, a comment was coming along with the update but I was delayed.
I tried to make it work with configure_file but it doesn't seem to trigger on file change so I thought that using copy_if_different would be an acceptable workaround for now.

lelongg:
It's working with configure_file now

rclrs_examples/CMakeLists.txt
```diff
@@ -56,13 +61,13 @@ ${_native_libraries_dirs}
...
-    rclrs_examples
-    DOWNLOAD_COMMAND ""
-    CONFIGURE_COMMAND ""
-    BUILD_COMMAND cargo build --release --manifest-path "${CMAKE_BINARY_DIR}/Cargo.toml"
-    INSTALL_COMMAND ""
-    LOG_BUILD ON)
+add_custom_target(cargo_build ALL
+    COMMENT "build crate")
+    
+add_custom_command(
```
esteve: To ensure that this command is only run when the output files are missing or out of date, you can add the following:

```cmake
add_custom_command(
OUTPUT
${CMAKE_BINARY_DIR}/ament_cargo/${PROJECT_NAME}/target/release/rclrs_publisher
${CMAKE_BINARY_DIR}/ament_cargo/${PROJECT_NAME}/target/release/rclrs_subscriber
TARGET cargo_build
...
)
```

lelongg: Apparently OUTPUT and TARGET cannot be mixed in add_custom_command.
> add_custom_command Wrong syntax.  A TARGET and OUTPUT can not both be specified.

Without TARGET I can't manage to make CMake trigger rebuild on source change :disappointed:

tprk77:
I just wanted to report in. I've rebased this branch onto your Crystal branch, but I'm not having much luck. I'm still having to delete the build directory to get proper builds. :speak_no_evil:

lelongg:
@esteve This should now be closer to what you expect.

@tprk77 Maybe you can try with last changes.

esteve:
@lelongg thanks for iterating on this

@tprk77 can you give it a try and see if it works for you?

tprk77:
Yay! It works! :tada:

(It might have also been working before. I originally thought this PR was making a change that would automatically apply to my package, but then I realized I needed to make similar changes to my own `CMakeLists.txt`.)

esteve:
I'll enable CI in the following days, but for now I'll just go ahead and merge this PR. Thanks @lelongg and @tprk77

PR #14, opened 2019-04-21
Set up CI with Azure Pipelines

esteve:


esteve:
@lelongg could you have a look at this? This will build and test any pull request so we can ensure that nothing breaks. Unfortunately, https://crates.io/crates/cargo-tarpaulin (the Rust code coverage crate) only supports Rust nightly, so I didn't add it, but we think it's worth adding rust-nightly to our CI, we can give it a try.

lelongg:
Ok, I will look at this while adding tests to #15

esteve:
@lelongg thanks for the review, I'll review #15 soon I hope.

lelongg:
Ok, unfortunately tests are not coming so soon :(

PR #18, opened 2019-07-12
Added closure subscribe

tprk77:
Adds the ability to subscribe with a `FnMut` closure.

Updated the example subscriber to use it.


esteve (approved):
@tprk77 thanks!

PR #20, opened 2020-03-11
fix warnings

deb0ch:
Hey 👋 

I fixed the warnings and the install steps in the README.

I am using Ubuntu 18.04 and the colcon build was failing because can't find `stdbool.h` and another thing, until I installed both `clang` _and_ `libclang`.


deb0ch:
The build is failing before compilation

esteve:
Changes in this PR are now in master as part of #24 Thanks @deb0ch !

Issue #21, opened 2020-05-12
IDL Support (Dashing+ Support)


AndrewJSchoen:
I wanted to separate out this specific issue to prevent overloading the topic in the main Dashing support thread. Specifically, this is a discussion for the planned feature of IDL support, required by ROS2 Dashing and later versions. 

Given the discussions there, there seem to be a couple of issues that need resolving, one major one being the choice of either using static compilation or introspection. I would be interested in any opinions that people have about that, as I look myself into how best to go about this.

lelongg:
For reference here is the link to the start of this discussion: #19 

Looking at the state of RTPS and IDL support on rust, here is a pair of interesting developments: [rtps-gen](https://github.com/frehberg/rtps-gen) and [rtps-rs](https://github.com/Klapeyron/rtps-rs).

It doesn't look ready but using these could prevent having to copy messages from Rust to C before serialization (and vice versa for deserialization).

nnarain:
Hi. Just wanted to say I'm also interested in this and giving it a shot. Current changes on the [dashing-plus](https://github.com/ros2-rust/ros2_rust/compare/master...nnarain:dashing-plus) branch in my fork.

Ported the IDL generator and made some fixes for the rcl API. Publish and subscribe examples are working. Need to clean it up though.

Also, I'm specifically building against Dashing. Figured I'd do Eloquent and Foxy too (Not sure to what extent things need to change there).

nnarain:
Wanted to note that Galactic has introduced some new tools for `rosidl`. Might be easier now to integrate with `cargo`.

nnmm:
I think this can be closed now – after #80, IDL support should be practically complete, at least in Foxy+.

PR #15, opened 2019-04-27
generation for all messages

lelongg:
rosidl_generator_rs now generates code for messages containing static and dynamic arrays and handles nested messages as well.

PR #24, opened 2020-06-24
Build on Dashing+

nnarain:
Hey. So I spent some time recently getting this working on Dashing, Eloquent and Foxy.

The main changes were updating message generation from IDL files in Dashing.

Pulled in @deb0ch warning fixes from this PR: https://github.com/ros2-rust/ros2_rust/pull/20 as a base.

Will probably take a shot at full IDL support #21. @lelongg has a [https://github.com/ros2-rust/ros2_rust/pull/15](PR), but it is pre-dashing.


nnarain:
I'll have to update the azure pipeline

ruffsl:
Could one of the current maintainers take a look to merge this? Seems like the number of folks writing their own rust client libraries from scratch is increasing due to the perceived inactivity of this official repo. @esteve @lelongg

lelongg:
I don't have a working dashing+ ros2 installation right now so it's difficult for me to test the changes but the changes look sensible. I suppose someone else making sure it's working should be enough to merge it (the pipeline tests are not testing much as I remember).

esteve (approved):
@nnarain thanks!

rosidl_generator_rs/rosidl_generator_rs/__init__.py
```diff
@@ -198,107 +223,88 @@ def constant_value_to_rs(type_, value):
...
+        elif type_.typename == 'uint32':
+            return 'u32'
+        elif type_.typename == 'int64':
+            return 'i64'
+        elif type_.typename == 'uint64':
+            return 'u64'
+    elif isinstance(type_, AbstractGenericString):
         return 'std::string::String'
+    elif isinstance(type_, Array):
+        return '[{}, {}]'.format(get_rs_type(type_.value_type), 32 if type_.size <= 32 else 32)
```
LoyVanBeek, line 257: I think this syntax for [defining Rust arrays of a fixed size ](https://doc.rust-lang.org/std/primitive.array.html)is slightly wrong. Instead of `[type, length]` shouldn't it be `[type; length]`? 
I'm a Rust n00b, so could be wrong, but I've added a dependency on `sensor_msgs` (which depends on `geometry_msgs`) in my project and got 3 errors like:

```
   Compiling geometry_msgs v0.1.0 (/home/loy/ros2_rust_ws/install/geometry_msgs/share/geometry_msgs/rust)
error: expected one of `!`, `(`, `+`, `::`, `;`, `<`, or `]`, found `,`
   --> /home/loy/ros2_rust_ws/install/geometry_msgs/share/geometry_msgs/rust/src/msg.rs:150:25
    |
150 |     pub covariance: [f64, 32],
    |                         ^ expected one of 7 possible tokens
```

With the change above, no such errors. One other remained, but that's a [separate issue](https://github.com/ros2-rust/ros2_rust/issues/26).

nnarain: Ah crap. Looks like you're right. I guess these packages don't get compiled unless you are importing `geometry_msgs`, since when I run `colcon build` I get:

```
Starting >>> geometry_msgs
Finished <<< geometry_msgs [2.91s]
```

We'll probably need ROS tests to cover all the standard ROS messages.

LoyVanBeek: Sometimes I hate being right :-). I'll make a PR

Issue #12, opened 2019-03-05
Idiomatic API for Rust


ruffsl:
Rather than merely a wrapper around the rcl c library and types, what other ways can we make the rclrs API more Idiomatic to rust design patterns and style? How do we expect to expose error handling, logging, composition of nodes, etc. This is intentionally open ended, but please feel free to narrow in scope.

tprk77:
Can you give some specific examples of what you mean? It's not like the current API is using raw pointers and using `int` for error codes. It already seems pretty Rusty to me. For example, it's using structs, traits, results, etc.

esteve:
One improvement I didn't have time to make was to tie the lifetime of `Publisher`s and `Subscription`s to `Node`s, but beside that, I'd say the API is fairly idiomatic. Like @tprk77 asked, @ruffsl do you have a concrete example of where the API of `rclrs` is not idiomatic? Thanks.

PR #25, opened 2020-08-30
Rewrite parts to make use of more idiomatic Rust

ThijsRay:
This work-in-progress pull request will aim to rewrite certain parts of the rclrs wrapper to use more idiomatic Rust. This will hopefully address the issue raised in #12. In this scenario, idiomatic Rust would mean:
* Preferring Rust's ownership, borrowing and lifetime rules over reference counting.
* Minimizing the amount of `unsafe` blocks and validating that all code in `unsafe` blocks is used properly.
* Using errors that are compatible with `std::error::Error`
* Preferring `Result`s over panicking. The user of the API should have full control over how errors are handled. The API should not panic when it is also possible to return a `Result`.


ZhenshengLee:
@ThijsRay I think there is a repo that can be used to help to design a more rustful api.

https://github.com/sequenceplanner/r2r

esteve:
@ZhenshengLee thanks for the link. The project you linked looks promising, however it lacks integration with the ROS2 build system, so it'll be very difficult for it to be part of the standard ROS2 distribution. It's unfortunate that given that despite `r2r` starting two years after `ros2-rust`, they didn't reach out to contribute, but I guess they had their reasons for not doing that

esteve:
@ThijsRay thanks for your contribution. This PR still marked as a draft after a while, are you still interested in getting in merged? Thanks!

ZhenshengLee:
> @ZhenshengLee thanks for the link. The project you linked looks promising, however it lacks integration with the ROS2 build system, so it'll be very difficult for it to be part of the standard ROS2 distribution. It's unfortunate that given that despite `r2r` starting two years after `ros2-rust`, they didn't reach out to contribute, but I guess they had their reasons for not doing that

It's true that the [r2r](https://github.com/sequenceplanner/r2r) repo lacks consideration of ros2 build system like `colcon` and `ament_cmake_cargo`, but I believe the api in https://github.com/sequenceplanner/r2r/blob/master/src is a more friendly api.

This friendly api, can be refactored and merged into https://github.com/ros2-rust/ros2_rust/tree/master/rclrs/src as `a friendly rust wrapper of rcl based on bindgen`, or as a layer on top of https://github.com/ros2-rust/ros2_rust/tree/master/rclrs , as `a rust client of the rclrs`, even with a name `rclrs_client`.

IamTheCarl:
Looking at That r2r repository, it's very well done. The only thing it lacks at all is The integration with the ros2 build system.

It is indeed unfortunate that they ran off ti make their own library rather than support the community directly.

But that does leave me wondering... They use an MIT license and you use an Apache license. Those are quite similar.

Is there a reason we shouldn't just fork their repo and add the build system support?

midnightexigent:
If you're looking to take inspiration from other repos, I came accross [rclrust](https://github.com/rclrust/rclrust) which seems to provide the easiest integration with colcon (https://github.com/rclrust/rclrust-examples)

I also found [RustDDS](https://github.com/jhelovuo/RustDDS/) but that isn't ros2 specific

esteve:
@ThijsRay I'm closing this since I've haven't heard from you in a while. Feel fee to submit these changes in another PR when you have more time. Thanks.

Issue #26, opened 2020-09-26
Error when generating code for sensor_msgs/JoyFeedback


LoyVanBeek:
The `sensor_msgs/JoyFeedback` message has a field called `type`, which is a keyword in Rust. 
A suggested fix is to prefix with `r#` but of course this should be done automagically.

Being a Rust n00b, I don't know the effect this has on how the code can be used and other effects.

nnarain:
Ya this needs to be fixed. `rosidl_python` prints a warning when keywords are used. Debated on adding this to the Dashing+ support MR but wanted keep it as simple as possible.

nnmm:
Closed by #30.

PR #30, opened 2020-11-02
Fix array type generation and field names that are keywords

nnarain:
Fixes a bug in array type generation. Also append an underscore to field names that are rust keywords.

https://github.com/ros2-rust/ros2_rust/issues/26


esteve:
@nnarain thanks!

esteve (approved):

Issue #43, opened 2021-09-03
Make compatible with Rolling


jhdcs:
Since this client library is under such active development, perhaps we should make it so that it is compatible with / buildable on Rolling?

I will look into what needs to change for this to happen.

esteve:
> I will look into what needs to change for this to happen.

Thanks! That'd be great

jhdcs:
So, currently I'm running into an issue where I can't build `rclrs_examples`:
<details>
    <summary>Complete Colcon Build output</summary>
    
```
jhassold@952e670e3ad9:~/ros2_rust_ws$ colcon build
Starting >>> ament_cmake_export_crates
Finished <<< ament_cmake_export_crates [0.14s]                  
Starting >>> rclrs_common
Finished <<< rclrs_common [0.12s]                
Starting >>> rclrs
Starting >>> rosidl_generator_rs
Finished <<< rclrs [0.13s]                                                               
Finished <<< rosidl_generator_rs [0.52s]                  
Starting >>> rclrs_examples
--- stderr: rclrs_examples                              
 Downloading crates ...
  Downloaded syn v1.0.76
  Downloaded cc v1.0.70
warning: `extern` block uses type `u128`, which is not FFI-safe
    --> /home/jhassold/ros2_rust_ws/build/rclrs_examples/ament_cargo/rclrs_examples/target/release/build/rclrs-d3b569f0be5f93c4/out/rcl_bindings.rs:2392:10
     |
2392 |     ) -> u128;
     |          ^^^^ not FFI-safe
     |
     = note: `#[warn(improper_ctypes)]` on by default
     = note: 128-bit integers don't currently have a known stable ABI

warning: `extern` block uses type `u128`, which is not FFI-safe
    --> /home/jhassold/ros2_rust_ws/build/rclrs_examples/ament_cargo/rclrs_examples/target/release/build/rclrs-d3b569f0be5f93c4/out/rcl_bindings.rs:4101:18
     |
4101 |         __value: u128,
     |                  ^^^^ not FFI-safe
     |
     = note: 128-bit integers don't currently have a known stable ABI

warning: `extern` block uses type `u128`, which is not FFI-safe
    --> /home/jhassold/ros2_rust_ws/build/rclrs_examples/ament_cargo/rclrs_examples/target/release/build/rclrs-d3b569f0be5f93c4/out/rcl_bindings.rs:4109:18
     |
4109 |         __value: u128,
     |                  ^^^^ not FFI-safe
[...]

PR #44, opened 2021-09-15
Compile `ros2_rust` in Rolling

jhdcs:
This branch is intended to fix issue #43

This PR allows `ros2_rust` to compile on ROS 2's Rolling branch, which will allow us to develop `ros2_rust` at near-parity with the cutting-edge of ROS 2.


jhdcs:
Figured out the non-compiling problem - I had forgotten to use `vcs import ./ < ros2_rust.repos`

The code compiles, and the example pubsub is able to execute successfully.

esteve:
@jhdcs thanks for your work, can you update the description of the PR with a brief summary of what the PR does? Thanks!

ros2_rust.repos
```diff
@@ -2,15 +2,15 @@ repositories:
   ros2/common_interfaces:
     type: git
     url: https://github.com/ros2/common_interfaces.git
-    version: foxy
```
esteve, line 5: Perhaps it'd make sense to have a separate `.repos` file only for Rolling, since interfaces packages may introduce incompatibilities between ROS 2 distros, @jhdcs what do you think?

esteve (changes requested):

esteve:
@jhdcs I only have one comment, but the changes look good to me, thanks.

ros2_rust.repos
jhdcs: This sounds reasonable to me. I've updated the Dockerfile as well to reflect the change, and allow the user to select different `.repos` files for building.

esteve (changes requested):
Changes look good. Could you update the README file and the CI to use the new `.repos` files? Thanks

jhdcs:
Sure, I'm assuming the CI is the `rust.yml` file, correct? Should it stay on Foxy, or go to Rolling?

esteve:
@jhdcs it'd be cool to have both, but at least Foxy. BTW, given that `ros2-rust` works with Rolling, it should be trivial to migrate to Galactic, what do you think?

esteve:
@jhdcs adding the following should run these jobs for Rolling and Galactic:

```
  build:

    strategy:
    matrix:
      ros_distro: [rolling, galactic]
    runs-on: ubuntu-latest
    container: ros:${{ matrix.ros_distro }}

...
          cp -av ${GITHUB_WORKSPACE}/ros2_rust_${{ matrix.ros_distro }}.repos .
          vcs import src < ros2_rust_${{ matrix.ros_distro }}.repos
```

jhdcs:
Will I need to repeat all of the `steps` portion of the `rust.yml` to run all the tests? Or is this instead of the stuff on lines 14-16?

esteve:
@jhdcs the CI config stays largely the same, adding those lines would make Github spawn two jobs, one for Rolling and another for Galactic, both with the same steps.

jhdcs:
Ok, I _think_ I've figured out the proper way to implement this. I misinterpreted your comment, earlier, but I think I've got it now...

jhdcs:
Hang on a second, do not approve this: There is a `wait_set` error that's showing up when you run the `rclrs_examples` subscriber and publisher. Trying to diagnose.

If you get a chance, can you confirm if you get the error as well?

.github/workflows/rust.yml
```diff
@@ -49,7 +52,7 @@ jobs:
 
     - name: Build ros2-rust (Linux)
       run: |
-          . /opt/ros/foxy/setup.sh
+          . /opt/ros/rolling/setup.sh
```
esteve, line 55: ``{{ matrix.ros_distro }}`` here instead of rolling

esteve (changes requested):

esteve:
@jhdcs I'm going to approve this PR just so that CI runs, but there's still one more change to make

esteve:
@jhdcs Github does not seem to like the syntax in the CI config:

```
The workflow is not valid. .github/workflows/rust.yml (Line: 15, Col: 14): Unexpected value '' .github/workflows/rust.yml (Line: 16, Col: 5): Unexpected value 'matrix'
```

esteve:
@jhdcs I had to approve this PR so that Github would allow the CI to run because this is your first contribution, but next PRs won't have that requirement

jhdcs:
Ah, ok, I understand!

Still trying to track down where the error is. It seems to be a wait set issue. Sanitized output:

```
>>> [rcutils|error_handling.c:108] rcutils_set_error_state()
This error state is being overwritten:

  'failed to create waitset, at (ros2_rolling_source_dir)/src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp:3521'

with this new error message:

  'wait_set is null, at (ros2_rolling_source_dir)src/ros2/rmw_cyclonedds/rmw_cyclonedds_cpp/src/rmw_node.cpp:3561'

rcutils_reset_error() should be called after error handling to avoid this.
<<<
Error: WaitSetInvalid
[ros2run]: Process exited with failure 1
```

Is this a new requirement that popped up for Rolling? Or is there something strange with my environment?

esteve:
@jhdcs I've pushed a change to fix the syntax of the CI config file

jhdcs:
So the issue seems to be occurring in `rclrs/src/lib.rs` within the `spin_once` command. Specifically, the call to `rcl_wait_set_init` will return a 900 (WaitSetInvalid) error after 8 to 10 seconds. However, I have been adding in debug statements that print out the input to the function, but I'm not seeing anything actually changing between calls.

@esteve, would you happen to know how waitsets are supposed to work? And what things could cause an `rcl_wait_set_init` call to fail? The only thing I can think of is that every time we have a TIMEOUT error, we just abandon the waitset within the function. Perhaps we should use `rcl_wait_set_fini()` on it?

If that's the case, I'm sure there's a way to implement a `drop` for the wait set type to do it for us...

esteve:
@jhdcs here's the documentation for `rcl_wait_set_init`: https://github.com/ros2/rcl/blob/master/rcl/include/rcl/wait.h#L77-L143

Looks like if the waitset is not destroyed properly, `rcl_wait_set_init` would return `RCL_RET_WAIT_SET_INVALID` (https://github.com/ros2/rcl/blob/master/rcl/include/rcl/wait.h#L128). Is that the error code you get?

If so, yeah, implementing `drop` would solve that. An even better solution would be to have a `Waitset` struct and implement the `Drop` trait for it, but for now I'd go for the simpler solution and add `Waitset` in a followup PR.

jhdcs:
Indeed, that's the error! The only issue I realized about implementing `Drop`, though, is that I the function for destroying a WaitSet (`rcl_wait_set_fini`) can fail. This would return an error code, but I don't think `Drop` allows returning a result - meaning our only resort if an error occurs is to `panic!`...

Is this acceptable? Should we allow the code to `panic!` in such an instance?

For reference, the documentation I found for `rcl_wait_set_fini` was [here](https://docs.ros2.org/beta1/api/rcl/wait_8h.html#a16b71f790a6cc23c1a7174b35e70fc8e). The `InvalidArgument` error should be taken care of if we have it wrapped up in a struct, and don't allow people to destroy the wrapped WaitSet (though not sure how to guard for that). The unspecified error I'd be okay with `panic!`ing on.

jhdcs:
There we go! Not the prettiest solution, but it will work until we get the WaitSet struct created!

esteve (approved):
@jhdcs awesome, thanks!

Issue #45, opened 2021-09-17
Replace `failure` with `thiserror`


jhdcs:
The [`failure` crate](https://github.com/rust-lang-deprecated/failure) currently used by `ros2_rust` is deprecated. As such, we should probably move towards a different error-handling crate.

The `failure` crate maintainer notes in the readme that the [`thiserror` crate](https://github.com/dtolnay/thiserror) is a near drop-in replacement, so that may be the direction we want to take.

jhdcs:
With the merging of PR #46, this issue is resolved.

esteve:
@jhdcs just a tip, when you open a PR, if it says "Closes #XXX", when the latter is merged, the issues will be closed automatically:

https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

PR #46, opened 2021-09-17
Conversion to `thiserror`

jhdcs:
This pull request addresses issue #45 

While the conversion is taking place, if anyone can see any improvements to the error-handling system of `ros2_rust`, please let me know!

### DIFF
--- a/rclrs_examples/src/rclrs_subscriber.rs
+++ b/rclrs_examples/src/rclrs_subscriber.rs
@@ -1,5 +1,5 @@
-use anyhow::{Error, Result, Context};
-use rclrs::{self, RclError};
+use anyhow::{Error, Result};
+use rclrs;
 use std_msgs;
 
 fn main() -> Result<(), Error> {
@@ -19,7 +19,6 @@ fn main() -> Result<(), Error> {
         },
     )?;
 
-    // rclrs::spin(&node).with_context(|| "RCL Error Occurred")
     match rclrs::spin(&node) {
         Ok(()) => Ok(()),
         Err(RclError) => Err(anyhow::Error::from(RclError)),
--- a/rclrs_examples/src/rclrs_subscriber.rs
+++ b/rclrs_examples/src/rclrs_subscriber.rs
@@ -1,7 +1,8 @@
-use rclrs;
+use anyhow::{Error, Result, Context};
+use rclrs::{self, RclError};
 use std_msgs;
 
-fn main() -> rclrs::RclResult {
+fn main() -> Result<(), Error> {
     let context = rclrs::Context::default();
 
     let mut node = context.create_node("minimal_subscriber")?;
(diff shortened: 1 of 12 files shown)


jhdcs:
Is there more that should be done for this? Do we want to re-organize how the error-handling code is laid out in general?

esteve:
@jhdcs I think it's fine to work on this now, but plan for streamlining error handling after this PR is merged

rclrs_examples/src/rclrs_subscriber.rs
```diff
@@ -18,5 +19,9 @@ fn main() -> rclrs::RclResult {
         },
     )?;
 
-    rclrs::spin(&node)
+    // rclrs::spin(&node).with_context(|| "RCL Error Occurred")
+    match rclrs::spin(&node) {
+        Ok(()) => Ok(()),
+        Err(RclError) => Err(anyhow::Error::from(RclError)),
```
esteve, line 25: Is this correct? I don't see RclError declared anywhere.

esteve: Ah, I see, thanks.

In that case, I'd get rid of the RCLError struct and rename RCLStatusCode to RclError, there's a naming inconsistency here RCL vs Rcl, I don't mind either way, as long as it's consistent. I wonder if there's a way to automatically generate RclError from the status codes in rcl.h, might be worth exploring that.

jhdcs: Oh, good catch on the inconsistency...

I was thinking of renaming RCLStatusCode to RclError, but one of the status codes actually pertains to a non-error: "Ok". I haven't figured out a nice way to deal with that, since it feels very confusing to have a non-error "Error". If you know what I mean?

And the current way to generate `RclStatusCode`s is that the enum implements the `From<i32>` trait. Though you have to be careful with that - if you pass an unsupported enum, you get an `unimplemented!` panic. See `rclrs_common/src/lib.rs`, line 76 for the current implementation.

esteve: > I haven't figured out a nice way to deal with that, since it feels very confusing to have a non-error "Error". If you know what I mean?

Yeah, I get you, I think that's why I added an `RclResult` struct, but I can't remember for sure. In any case I like how succinct your solution is. Maybe `RclStatus`? Though it does not convey when it's an actual error. Anyway, let's keep it as `RclError`.

I thought of generating `RclError` using `libclang` and `bindgen`, but on a second thought, doesn't make much sense because the error values won't change often.

jhdcs: I think what I'll do is leave this for a general error streamlining process. I'm thinking that perhaps the best way to handle this is to wrap all the `unsafe` `rcl` calls and convert their outputs to things that Rust is more familiar with handling (i.e: Result types). This could be doable with a quick tweak in the `rclrs_common/src/library.rs` code, where instead of a `From<i32> for RclError` we could figure out a function to convert said `i32` into a full-fledged Result type.

But that's for another PR, if it's possible.

esteve (approved):
@jhdcs LGTM, thanks for addressing my feedback

PR #48, opened 2021-09-28
Implement WaitSet

### LINKED ISSUE #47: Implement `WaitSet` wrapper struct
While updating the code to Rolling, we noticed that it's very important that the `WaitSet`s are destroyed properly. However, as they are an FFI type, Rust can't `Drop` them as-is.

We need to construct a wrapper `struct` around the `WaitSet` that implements `Drop` to call the proper code to dispose of the `WaitSet`. A thing to consider, though: `WaitSet` disposal is not guaranteed. If a `WaitSet` is improperly `Drop`ped, should we `panic!`?

`WaitSet` source documentation: https://github.com/ros2/rcl/blob/master/rcl/include/rcl/wait.h#L77-L143

jhdcs:
Closes #47 

Also includes a rework of the error system around the `spin` functionality, to accept the new ways that the `WaitSet` can fail.

### DIFF
--- a/rclrs/src/error.rs
+++ b/rclrs/src/error.rs
@@ -14,10 +14,5 @@ pub(crate) trait ToResult {
 impl ToResult for rcl_ret_t {
     fn ok(&self) -> Result<(), RclError> {
         to_rcl_result(*self as i32)
-    //     if *self as u32 == RCL_RET_OK {
-    //         Ok(())
-    //     } else {
-    //         Err(RclError::from(*self))
-    //     }
     }
 }
--- a/rclrs/src/error.rs
+++ b/rclrs/src/error.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 
 use crate::rcl_bindings::*;
-pub use rclrs_common::error::RclError;
+pub use rclrs_common::error::{RclError, to_rcl_result};
 
 pub(crate) trait ToResult {
     fn ok(&self) -> Result<(), RclError>;
@@ -13,10 +13,11 @@ pub(crate) trait ToResult {
 
 impl ToResult for rcl_ret_t {
     fn ok(&self) -> Result<(), RclError> {
-        if *self as u32 == RCL_RET_OK {
-            Ok(())
-        } else {
-            Err(RclError::from(*self))
-        }
+        to_rcl_result(*self as i32)
+    //     if *self as u32 == RCL_RET_OK {
+    //         Ok(())
+    //     } else {
+    //         Err(RclError::from(*self))
+    //     }
     }
 }
--- a/rclrs/src/lib.rs
+++ b/rclrs/src/lib.rs
@@ -97,8 +97,6 @@ pub fn spin_once(node: &Node, timeout: i64) -> Result<(), WaitSetError> {
         number_of_events,
         context)?;
 
-    wait_set.clear()?;
-
     for subscription in &node.subscriptions {
         match wait_set.add_subscription(subscription) {
             Ok(()) => (),
--- a/rclrs/src/lib.rs
+++ b/rclrs/src/lib.rs
@@ -92,19 +92,6 @@ pub fn spin_once(node: &Node, timeout: i64) -> Result<(), WaitSetError> {
 
     let context = &mut *node.context.get_mut();
 
-    // unsafe {
-    //     rcl_wait_set_init(
-    //         &mut wait_set_handle as *mut _,
-    //         number_of_subscriptions,
-    //         number_of_guard_conditions,
-    //         number_of_timers,
-    //         number_of_clients,
-    //         number_of_services,
-    //         number_of_events,
-    //         context,
-    //         rcutils_get_default_allocator(),
-    //     ).ok()?;
-    // }
     wait_set.init(
         number_of_subscriptions,
         number_of_guard_conditions,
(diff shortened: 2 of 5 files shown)


esteve:
@jhdcs is this ready for review?

jhdcs:
Yes it is! Sorry, was just about to assign you...

rclrs/src/error.rs
```diff
@@ -13,10 +13,11 @@ pub(crate) trait ToResult {
 
 impl ToResult for rcl_ret_t {
     fn ok(&self) -> Result<(), RclError> {
-        if *self as u32 == RCL_RET_OK {
-            Ok(())
-        } else {
-            Err(RclError::from(*self))
-        }
+        to_rcl_result(*self as i32)
+    //     if *self as u32 == RCL_RET_OK {
```
esteve, line 17: If this code is no longer used, it should be removed.

rclrs/src/lib.rs
```diff
@@ -87,80 +92,97 @@ pub fn spin_once(node: &Node, timeout: i64) -> Result<(), RclError> {
...
-                    Ok(()) => return Err(rcl_err),
-                    Err(rcl_err2) => {
-                        eprintln!("Trying to clear the WaitSet caused a second error!! Error code: {:?}", rcl_err2);
-                        return Err(rcl_err2);
-                    }
-                }
-            }
-        };
-    }
+    // unsafe {
```
esteve, line 95: Same here about commented code

jhdcs:
Ah, hang on a second, missed some comments!

rclrs/src/lib.rs
```diff
@@ -87,80 +92,97 @@ pub fn spin_once(node: &Node, timeout: i64) -> Result<(), RclError> {
...
+        number_of_guard_conditions,
+        number_of_timers,
+        number_of_clients,
+        number_of_services,
+        number_of_events,
+        context)?;
+
+    
+        
+    // unsafe {
```
esteve, line 119: More commented out code.

rclrs/src/lib.rs
```diff
@@ -87,80 +92,97 @@ pub fn spin_once(node: &Node, timeout: i64) -> Result<(), RclError> {
...
-                        }
-                    }
-                };
-            }
-        }
-    }
-
-    unsafe {
-        match rcl_wait(&mut wait_set_handle as *mut _, timeout).ok() {
+        // if let Some(subscription) = subscription.upgrade() {
```
esteve, line 137: Commented code.

rclrs/src/lib.rs
```diff
@@ -87,80 +92,97 @@ pub fn spin_once(node: &Node, timeout: i64) -> Result<(), RclError> {
...
-                }
-            }
-
+            Err(WaitSetError::DroppedSubscription) => (),
+            err => return err,
         };
     }
 
+    wait_set.wait(timeout)?;
+    // unsafe {
```
esteve, line 168: Commented code.

rclrs/src/wait.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub wait_set: rcl_wait_set_t,
+    initialized: bool,
+}
+
+impl WaitSet {
+    /// Creates a new WaitSet object.
+    ///
+    /// Under the hood, this calls `rcl_get_zero_initialized_wait_set()`, and stores it
+    /// within the WaitSet struct, while also noting that the returned value is uninitialized.
+    pub fn new() -> Self {
```
esteve, line 40: I'd fuse `new` and `init`, it makes sense to have the both in the same function because one wouldn't reinitialize a previously initialized waitset.

rclrs_common/src/lib.rs
```diff
@@ -65,46 +65,93 @@ pub mod error {
...
+
+    #[derive(Debug, Error)]
+    pub enum WaitSetError {
+        #[error("Passed subscription was dropped")]
+        DroppedSubscription,
+        #[error("Rcl error occurred")]
+        RclError (#[from] RclError)
+    }
+
+    // impl From<i32> for RclError {
```
esteve, line 117: More commented code.

esteve (changes requested):

esteve:
@jhdcs

> Ah, hang on a second, missed some comments!

I didn't see this comment before my review :sweat_smile:

jhdcs:
Odd, it doesn't seem to be building on GitHub... Did I make a mistake and use an unclean build?

rclrs/src/lib.rs
```diff
@@ -87,80 +88,26 @@ pub fn spin_once(node: &Node, timeout: i64) -> Result<(), RclError> {
...
-                    Ok(()) => return Err(rcl_err),
-                    Err(rcl_err2) => {
-                        eprintln!("Trying to clear the WaitSet caused a second error!! Error code: {:?}", rcl_err2);
-                        return Err(rcl_err2);
-                    }
-                }
-            }
-        };
-    }
+    wait_set.clear()?;
```
esteve, line 100: The entities in the `WaitSet` should already be initialized to `NULL` because the `new` function calls `rcl_get_zero_initialized_wait_set()`, so this call shouldn't be necessary.

esteve (changes requested):

esteve:
@jhdcs CI is failing now, but the changes overall look good. I'll have another look at the PR later, but thanks for implementing this.

rclrs/src/lib.rs
jhdcs: This was how it was set up originally, but I agree - it looks odd.

jhdcs:
You're quite welcome!

I'm not sure why my local environment was able to build this while GitHub couldn't (It seems that the GitHub version doesn't have an `rcl_context_s`, so I had to use `rcl_context_t`). I'm going to try rebuilding the container, hopefully that irons out further issues.

esteve (approved):
@jhdcs excellent! Thank you so much :slightly_smiling_face:

PR #51, opened 2021-10-05
Add Arc and Mutex

### LINKED ISSUE #50: Move to use `Arc` instead of `Rc`
In order to be able to use `rclrs` in multi-threaded applications, we will need to move away from single-threaded constructs such as `Rc`. As such, we should go through and convert them to `Arc` (or other multi-threaded equivalents) soon, while the library is still small.

jhdcs:
Closes #50

This PR replaces the `Rc` and `RefCell` usages with `Arc` and `parking_lot::Mutex`, respectively.

This PR does **NOT** implement a multithreaded spinner - this is just work to make such a spinner easier to implement in future.

### DIFF
--- a/rclrs/Cargo.toml
+++ b/rclrs/Cargo.toml
@@ -5,9 +5,10 @@ authors = ["Esteve Fernandez <esteve@apache.org>"]
 edition = "2018"
 
 [dependencies]
-libc = "0.2.43"
-thiserror = "1"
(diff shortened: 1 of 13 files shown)


esteve:
@jhdcs looks good, we'll need to rework some of this once we have an API for executors, but that's ok. Thanks!

PR #52, opened 2021-10-11
Add unit testing

### LINKED ISSUE #35: Missing Unit Tests
Similar to #33, the code on a whole lacks unit testing. We should implement these tests sooner rather than later, since unit tests should ideally be created at around the same time the code that they're testing is made.

For the Rust portions of the project, I think the files are small enough that we can go with the standard Rustic way of adding unit testing in a module at the bottom of a file. Once the files get bigger, we can consider moving them off into their own folder. We will need to teach `colcon-cargo` how to run these tests, but that shouldn't be too difficult.

The other languages should follow the standard ROS 2 rules for unit testing.

jhdcs:
Closes #35 

This pull request will implement at least a basic level of unit tests across the client library.

Currently running into difficulty wiring up the unit tests with the greater ROS environment - in other words, the Rust unit tests can't see the other crates within the client library at test-time, even when run via `colcon test`.

Any advice or suggestions on how to deal with this problem would be welcome!

### DIFF
--- a/rclrs/CMakeLists.txt
+++ b/rclrs/CMakeLists.txt
@@ -21,11 +21,45 @@ install(FILES
 )
 
 if (BUILD_TESTING)
-  add_test(NAME rclrs_test
-    COMMAND cargo test
-    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
-endif()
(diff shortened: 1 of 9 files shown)


jhdcs:
As this PR is being superseded by #84, this can be safely closed.

PR #59, opened 2021-10-14
Implement no_std version of error

### LINKED ISSUE #53: Port RCL Errors to RCLRS Core
The RCL errors need to be ported to the new RCLRS Core crate.

Due to `thiserror` apparently not being `no_std` compatible (as far as I can tell), we will have to rely on more "old-fashioned" error handling - namely, really descriptive enums...

jhdcs:
Closes #53 

Creates the `rclrs_core` crate, and adds the error types + unit tests.

From a design standpoint - should I attempt to re-work the `RclStatusCode` to be more like a `Result`? Or stay as close to what `rcl` provides as possible?

### DIFF
--- a/rclrs_core/Cargo.toml
+++ b/rclrs_core/Cargo.toml
@@ -9,3 +9,6 @@ libc = {version = "0.2", features = []}
 
 [build-dependencies]
 bindgen = "0.59.1"
+
+[profile.release]
+panic = 'abort'


jhdcs:
I can do that, sure, but that means this one PR will be closing a lot of issues at the same time (since I'm not sure that this will compile until all the changes are implemented).

Also, from what I understand, `Arc` and `Mutex`, as well as most heap allocations will not work with `no_std`. Which means we may have difficulty implementing things like the "handles" for the various components. Part of the reason I was splitting this to a separate crate was so that we could use `rclrs` to import from that crate and add in the heap features, making the API look like it does now, and leaving `rclrs_core` as just a wrapper over the various `rcl` pieces.

Did you have a different idea for how to approach this problem?

PR #60, opened 2021-10-19
Make `rclrs` `no_std` compatible

### LINKED ISSUE #53: Port RCL Errors to RCLRS Core

### LINKED ISSUE #54: Port Node to RCLRS Core
Nodes needs to be ported to the new `rclrs_core` crate.

I believe we cannot use `Arc` or `Mutex` with the `!#[no_std]`  attribute, so that logic might need to be left up to a higher-level API.

### LINKED ISSUE #55: Port Context to RCLRS Core
Context needs to be ported to the new `rclrs_core` crate.

I believe we cannot use `Arc` or `Mutex` with the `!#[no_std]` attribute, so that logic might need to be left up to a higher-level API.

### LINKED ISSUE #56: Port QOS to RCLRS Core
The QOS profiles need to be ported to the new `rclrs_core` crate.

This will probably be as simple as copying over the current QOS implementation.

### LINKED ISSUE #58: Port WaitSet to RCLRS Core
WaitSets need to be ported to the new `rclrs_core` crate.

I believe we cannot use `Arc` or `Mutex` with the `!#[no_std]` attribute, so that logic might need to be left up to a higher-level API.

jhdcs:
Closes #53, Closes #54, Closes #55, Closes #56, Closes #58

This is a **MASSIVE** refactoring of `rclrs`. Due to the limitations of `no_std`, I had to spin my own simple mutex (which will need to be looked at and fine-tuned later, it can be found in `spinlock.rs`). We also lose `thiserror` support, since `Error` does not exist in `core`.

I am also unable to use environment variables in `no_std`. As such, the arguments required for creating an `rcl` context have to be passed to the pseudo-`Default` function.

While working on this crate, I found myself wishing to merge `rclrs_common` into `rclrs` (both for convenience, and because I think I might be able to tweak the Errors to directly reference the `rcl_return_type_t` codes - meaning if `rcl` changes the code numbers, we'd automatically reflect that), however, I will make another issue to discuss whether or not that's a good idea.


esteve (approved):
Awesome!

PR #64, opened 2021-11-04
Build system refactor

### LINKED ISSUE #13: Improve build integration with cargo
As mentioned in the README, the current status with CMake is less than settler:

> The current examples use CMake to install and build the binaries... and it's really ugly.

The Rust ecosystem uses cargo as it's native build system; thus advancing build framework to better integrate with cargo could go a long way in helping the usability of the client library. Given [colcon-cargo](https://github.com/colcon/colcon-cargo) is still in early development, how can we improve upon this for rclrs?

### LINKED ISSUE #31: Organize Rust Code As Cargo Workspace?
Would it be valuable for us to organize the Rust portions of `rclrs` as a [cargo workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html)? I could see it theoretically help with organizing the code, but I'm unsure of how valuable it would be.

Anyone have any thoughts?

### LINKED ISSUE #34: Improve support for rust language analyzers
During the last Rust WG meeting, members reported difficulties in trying to get rclrs to play nicely with VSCode + Rust-Analyzer. In particular, one of the hangups is that rust-analyzer gets confused by auto generated messages, and can’t recognize them as types. This could tie into some of the other tickets for improving message generation and build integration with cargo, although this ticket specifically tracks issues for improving support for rust language analyzers such as rls or rust-analyzer:

* https://github.com/rust-analyzer/rust-analyzer
* https://github.com/rust-lang/rls

This'll not only aid in rclrs development, but also help downstream users creating their own ROS Rust packages.

Related:
* https://github.com/ros2-rust/ros2_rust/issues/11
* https://github.com/ros2-rust/ros2_rust/issues/13

### LINKED ISSUE #38: Use of ament_cargo directory in ros2_rust examples
In the rclrs_examples CMakeLists.txt files, one can find the following lines:

```
file(WRITE "${CMAKE_BINARY_DIR}/.cargo/config"
"\
[build]\n\
target-dir = 'ament_cargo/${PROJECT_NAME}/target'\n\
rustflags = [\
${_native_libraries_dirs}
]\n\
"
)

file(COPY "${CMAKE_SOURCE_DIR}/Cargo.toml" DESTINATION "${CMAKE_BINARY_DIR}/")
file(APPEND "${CMAKE_BINARY_DIR}/Cargo.toml" "${_crates_dependencies}")

add_custom_command(
    OUTPUT
    ¦   ${CMAKE_BINARY_DIR}/ament_cargo/${PROJECT_NAME}/target/release/rclrs_publisher
    ¦   ${CMAKE_BINARY_DIR}/ament_cargo/${PROJECT_NAME}/target/release/rclrs_subscriber
    COMMAND cargo build --release --manifest-path "${CMAKE_BINARY_DIR}/Cargo.toml"
    DEPENDS
    ¦   ${CMAKE_BINARY_DIR}/src/rclrs_publisher.rs
    ¦   ${CMAKE_BINARY_DIR}/src/rclrs_subscriber.rs
)
```

Is there a particular reason we are putting the output executables under the `ament_cargo` directory?

### LINKED ISSUE #61: Reduce amount of bindings
While working on `rclrs`, I've noticed a lot of warnings being generated about `u128` not being FFI safe. While this is true, fixing that is beyond our power.

What _ISN'T_ beyond our power is removing the bindings completely - as it seems that most of these bindings are for libraries like `math.h`, which aren't going to be very useful to users of `rclrs`. As such, we should look through the generated code, and see what can be done to block unneeded functions from showing up in the bindings.

jhdcs:
Closes #61, Closes #34, Closes #38, Closes #31, Closes #13

I have refactored the build system so that things can be build in a more Rust-like way: The biggest surprise people may have is that the `rclrs_examples` crate will generate a `.cargo/config.toml` file in order to patch in the file dependency for `std_msgs`.

Do the changes look acceptable? While I'm still not super-fond of the dependency search we have to do for `rclrs_examples`, I'm sure that we can find something else to make that easier in another PR.


jhdcs:
It appears the changes to the generator that are required for H-Turtle/Rolling aren't backwards-compatible. Is there a way we can deal with this?

jhdcs:
Managed to fix it. This should be good for review!

rclrs/Cargo.toml
```diff
@@ -13,6 +16,10 @@ core-error = "0.0.0"
 parking_lot = {version = "0.11.2", optional = true}
 spin = "0.9.2"
 
+[dependencies.rclrs_common]
+version = "*"
+path = "../rclrs_common"
```
esteve, line 21: Is there a way to avoid this? It'd be unfortunate because it's bypassing the ROS dependency resolution mechanism, but it makes `rust-analyzer`, it's fine as a compromise. Or maybe, it'd just make sense to integrate `rclrs_common` into `rclrs`. What do you think?

jhdcs: I had been debating merging the two crates for some time. I will give that a shot and see how it goes.

I am doing something similar in `rclrs_examples` as well. I'm assuming it's a problem there as well? What I'd like to do is figure out some way to make ROS 2's dependency resolution work, while also allowing `cargo` the freedom to build crates as it likes, since that's what most IDEs will be expecting...

jhdcs: Okay, that took quite a bit more effort than I was expecting, but I have managed to combine most of `rclrs_common` with `rclrs`. The message traits had to be spun off into their own crate, otherwise we ended up with circular build dependencies.

I can do a bit more thinking on how to avoid the explicit paths in the manifest. Perhaps we should use an auto-generated `.cargo/config.toml` file, such as the one that I use for the `std_msgs` dependencies?

rclrs_msg_utilities/package.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0"?>
```
esteve, line 1: Correct me if I'm wrong, but it seems that `rclrs_msg_utilities` serves a similar purpose as `rclrs_common`, that is, code shared between `rclrs` and downstream packages / generators. Would it make sense to rework `rclrs_common` to include the additions from `rclrs_msg_utilities` ?

jhdcs: It cannot, as best as I am able to tell. Every time I try, it introduces a circular build dependency.

jhdcs:
Please stand by, working on a solution to having explicit paths in the manifests for the packages...

rclrs_msg_utilities/package.xml
jhdcs: Oh, sorry, misunderstood your question. So basically what `rclrs_msg_utilities` is is just `rclrs_common` stripped of everything that doesn't have to do with messages. My goal with it is to eventually use it to help with message generation via CMake.

Since I'm not fully sure what the plan is anymore - We were going to switch over to `r2r`, but then I got comments saying that some other people preferred the CMake implementation...

jhdcs:
Okay, should be refactored now. Basically, you feed in the packages you want (and listed in the `package.xml` into a new CMake macro called `rclrs_gen_crate_config` and it will take care of generating the necessary `.cargo/config.toml` file for you!

esteve:
@jhdcs thank you so much! Looks great :+1:

esteve (approved):

PR #73, opened 2022-02-14
Use the ament_cargo build type

nnmm:
The rclrs_crate_config_generator is superseded by colcon-ros-cargo.
The ament_cmake_export_crates mechanism is subsumed by creating entries in the ament index directly in the rosidl_generator_rs and cargo-ros-install.


nnmm:
@esteve @jhdcs @mxgrey This is waiting on https://github.com/colcon/colcon-ros-cargo/pull/1, but I expect no major changes to this part, so I'm inviting reviews.

jhdcs (dismissed):
Looks good to me!

I've tested this on my Rolling-based system - that needed some changes to the build script in `rclrs` in order to run, but I'll open up a separate PR to bring that into `ros2_rust`

esteve:
@jhdcs for first time contributors, we need to run the CI manually, I've just triggered it.

@nnmm thanks for this amazing work, I'll try to run this today or tomorrow at the latest

rosidl_generator_rs/cmake/rosidl_generator_rs_generate_interfaces.cmake
```diff
@@ -242,25 +231,21 @@ foreach(_typesupport_impl ${_typesupport_impls})
 
   if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
     install(
-      DIRECTORY "${_output_path}/rust"
+      DIRECTORY "${_output_path}/Rust"
```
esteve, line 234: Why the rename?

ros2_rust_galactic.repos
```diff
@@ -11,8 +11,11 @@ repositories:
     type: git
     url: https://github.com/ros2/rosidl_defaults.git
     version: galactic
-  # ros2-rust/ros2_rust:
-  #   type: git
-  #   url: https://github.com/ros2-rust/ros2_rust.git
-  #   version: master
-
+  colcon/colcon-cargo:
```
esteve, line 14: I wouldn't put this in a `.repos` file, but just use `pip` to install the Python package

esteve (changes requested):

esteve:
@nnmm CI is failing, could you update the CI config to skip the packages that no longer exist? Thanks

ros2_rust_galactic.repos
nnmm: I'm not sure how to make that work. Can you give more detail? If we don't have colcon-cargo checked out, it would need to be published on PyPI, right?

rosidl_generator_rs/cmake/rosidl_generator_rs_generate_interfaces.cmake
nnmm: Well, I'd always capitalize Rust, so it just seem tidier to me. But we could also use "rs" as an abbreviation, or revert that.

esteve: Rust in documents and comments I'd say yes always, but for paths I don't see why. For example, Cargo uses `.cargo` in lower case for its path. Having paths in lowercase seems more convenient to me.

ros2_rust_galactic.repos
esteve: No problem! You can install Python packages using `pip` directly from Git repositories `pip install git+https://github.com/colcon/colcon-cargo.git` should be sufficient. You can even pass a tag or a commit id like this `pip install git+https://github.com/colcon/colcon-ros-cargo.git@v1`

We should tag `colcon-cargo` so you can more easily depend on it.

esteve:
> @nnmm t's not about missing packages, but the two-stage build process (colcon-ros-cargo needs to be built & sourced first).

You're right. We can install the `colcon` plugins via `pip` and then use the GitHub action with the repos file as it was, if you're ok with that, I'll make a couple of changes to the CI and see if it works.

ros2_rust_galactic.repos
esteve: Yes

esteve:
> @esteve You're right. We can install the colcon plugins via pip and then use the GitHub action with the repos file as it was, if you're ok with that, I'll make a couple of changes to the CI and see if it works.

@nnmm turns out I can't install `colcon-ros-cargo` directly with `pip` because the `setup.py` file in the repo is not in the root directory. Would you be ok with moving `colcon-ros-install` to its own repo in the `ros2-rust` organization and have `colcon-ros-cargo` only for the Python package? I've noticed that all the other `colcon` plugins have that directory structure (e.g. https://github.com/colcon/colcon-ros, https://github.com/colcon/colcon-gradle, etc.)

ros2_rust_galactic.repos
nnmm: Thanks for clarifying. This works with transitive dependencies? And the resulting packages still need to be built by colcon, right?

From my current point of view, I'd favor the `.repos` approach, since Python package management has a deserved reputation for being really finicky. We're using `vcs` anyway, so there's no extra cognitive load. Everything is visible and in one place (the ros2_rust dir).

Differently stated, what does using `pip` here buy us?

rosidl_generator_rs/cmake/rosidl_generator_rs_generate_interfaces.cmake
nnmm: I guess when it's lowercase I read it as a verb, "to rust". But I'll revert it.

ros2_rust_galactic.repos
esteve: > This works with transitive dependencies?

Yes.

> And the resulting packages still need to be built by colcon, right?

Not sure I'm following. `colcon` plugins are just Python packages, so once they are in the PYTHONPATH, `colcon` will discover them and use them, but they are not built by `colcon`.

> From my current point of view, I'd favor the .repos approach, since Python package management has a deserved reputation for being really finicky. We're using vcs anyway, so there's no extra cognitive load. Everything is visible and in one place (the ros2_rust dir).

Eventually that won't be the case, `colcon` plugins are released independently (e.g. `colcon-ros` is just a Python package that's been published to PyPI https://pypi.org/project/colcon-ros/)

> Differently stated, what does using pip here buy us?

`pip` is the standard way of installing Python packages, it can also use PyPI (the Python equivalent to crates.io), and it's the way all the other `colcon` plugins are installed.

nnmm: Ok, so if I understand right, we wouldn't need the two-stage build process anymore (if we forget about cargo-ros-install for a second). That's a nice benefit.

Now about cargo-ros-install, I could upload it to crates.io and we install it from there. Otherwise, if there's a way to execute the `cargo build` command as part of the setup process of colcon-ros-cargo (which is invoked by pip, right?), then that would also solve the issue. Can you run custom commands during setup, à la build.rs?

esteve: I wouldn't make `cargo build` a part of the install process of `colcon-ros-cargo`. `pip` may be invoked as root and not the as the user that has `cargo` installed, it may be installed into a virtualenv, etc. If you want to be very strict with ROS dependencies, I'd create a `colcon-ros-install_vendor` package that checks if the `colcon-ros-install` crate is installed, and if not, run `cargo build/install`. But I'd just push `colcon-ros-install` to crates.io and add a line in the README telling users to run `cargo install colcon-ros-install`

nnmm: Ok. I'll probably rename it to `cargo-ros-build` or `cargo-ament-build`.

nnmm: Renamed to `cargo-ament-build`. Can you rename the new repo?

rosidl_generator_rs/cmake/rosidl_generator_rs_generate_interfaces.cmake
nnmm: Done.

esteve:
@nnmm I've added a step in the CI to install `cargo-ament-build` now that it's available in crates.io, this should fix any issues in the CI

esteve:
@nnmm one more commit to remove `ament_cmake_export_crates` and `colcon-ros-cargo` as dependencies from the `package.xml` files to make `rosdep` happy

nnmm:
@esteve Can you mark the rolling job as allowed to fail? Jacob wanted to land the fix for that after this PR.

esteve:
@nnmm I've disabled the Rolling job because even if a job is marked as `continue-on-error`, the workflow will still be marked as failing. We can reenable Rolling once the `rcl` issue is fixed.

nnmm:
I'll take a look at that error. I also think a failed cargo build shouldn't throw an uncaught exception, so I'll fix that.

nnmm:
@esteve Could I get permission to kick off workflows?

README.md
```diff
@@ -26,69 +26,61 @@ What's missing?
...
 
 ### Publisher and subscriber
 
 Publisher:
 
 ```
-. ./install/setup.sh
-
+# Do this in a new terminal
+source ./install/setup.sh
```
esteve, line 72: `source` is a bash-ism. So we either use `.` and the shell scripts (ending in `.sh`), or `source` and the scripts that end in `.bash`, hence why the current version of the README.md uses `.`

azure-pipelines.yml
```diff
@@ -54,6 +54,6 @@ steps:
   - script: |
       source /opt/ros/foxy/setup.bash
       cd ~/ros2_rust_ws
-      colcon test --packages-select rclrs rclrs_common rclrs_examples rosidl_generator_rs
+      colcon test --packages-select rclrs rclrs_examples rosidl_generator_rs
```
esteve, line 57: I've submitted https://github.com/ros2-rust/ros2_rust/pull/74 for review to remove the Azure pipelines config, now that we're using GitHub actions for CI. Can you approve that PR and rebase your branch? Thanks.

README.md
```diff
@@ -26,69 +26,61 @@ What's missing?
...
-git clone https://github.com/ros2-rust/ros2_rust.git
-cd ../../
-vcs import src < src/ros2-rust/ros2_rust/ros2_rust_foxy.repos
+# In your workspace directory (ideally an empty one), run 
+mkdir src
+git clone https://github.com/ros2-rust/ros2_rust.git src/ros2_rust
+vcs import src < src/ros2_rust/ros2_rust_foxy.repos
 source /opt/ros/foxy/setup.sh
-colcon build
+colcon build --install-base install_colcon --packages-up-to colcon-ros-cargo
```
esteve, line 56: A separate workspace for `colcon` is no longer needed because we can install `colcon-ros-cargo` via `pip`

README.md
```diff
@@ -26,69 +26,61 @@ What's missing?
...
-Or do so manually as summarized in the steps below:
-
-> The following steps were last tested on Ubuntu 20.04.
-
-```
-# first, install vcstool from PyPI or apt:
+# First, make sure to have ROS 2 and vcstool installed:
 # sudo apt install ros-foxy-desktop ros-foxy-test-interface-files python3-vcstool libclang-dev clang
+# vcstool can alternatively be installed from PyPI:
 # pip install vcstool
```
esteve, line 49: A line for installing `cargo-ament-build` needs to be added to README.md

esteve (changes requested):

README.md
nnmm: Done.

azure-pipelines.yml
nnmm: Done.

README.md
nnmm: Ok, replaced with `.`.

esteve (approved):

esteve:
@nnmm bravo! :clap:  :clap:  :clap:

PR #74, opened 2022-02-17
Removed outdated Azure pipelines config

esteve:
We're using GitHub actions now, there's no need for the Azure pipelines config file


nnmm (approved):
LGTM

PR #76, opened 2022-02-20
Add support for sized arrays of primitive type in rosidl_generator_rs

nnmm:
Annotations are not yet supported. I tested it by converting rclrs_example to send a `TwistWithCovariance`. The covariance part works now, but not the twist part (which needs nested type support).

I also deleted a bunch of files that were unused, as well as some empty variables in the CMake script.


rosidl_generator_rs/rosidl_generator_rs/__init__.py
```diff
@@ -269,7 +269,7 @@ def get_builtin_rs_type(type_, package_name=None):
     elif isinstance(type_, AbstractGenericString):
         return 'std::string::String'
     elif isinstance(type_, Array):
-        return '[{}; {}]'.format(get_rs_type(type_.value_type), 32 if type_.size <= 32 else 32)
```
nnmm, line 272: Not sure what that was for

nnmm:
Also: Why are the `*_rs.ep**` files not marked as script outputs in CMake?

PR #77, opened 2022-02-21
Bug/fix missing includes

### LINKED ISSUE #72: [Rolling] `rclrs` unable to build - cannot find headers for wrapper
`rclrs` is failing to build on the latest version of Rolling - `bindgen` is unable to find the includes for `rcl` during compilation.

It is possible that this issue is related to https://github.com/ros2/rcl/pull/959. If so, the build script for `rclrs` will need to be tweaked to point to the new location of the headers.

jhdcs:
Fixes #72 

Re-configured the build script to follow the new folder structure set in `rcl`. Also fixed a longstanding bug in the `allowlist_var` calls.

### DIFF
--- a/rclrs/build.rs
+++ b/rclrs/build.rs
@@ -2,7 +2,7 @@ extern crate bindgen;
 
 use std::env;
 use std::fs::read_dir;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 fn main() {
     let mut builder = bindgen::Builder::default()
@@ -36,7 +36,7 @@ fn main() {
     // Search through ament index at ${prefix}/share/ament_index/resource_index/packages/ to find packages to include
     // The include root will be located at either:
     // - ${prefix}/include/ (old style)
-    // - ${prefix}/include/package_name (new style)
+    // - ${prefix}/include/${package_name} (new style)
     // - ${prefix}/include/CycloneDDS (special case, match for this)
     // End of loop
     // Compiled libraries are always at ${prefix}/lib
@@ -44,49 +44,40 @@ fn main() {
     // See REP 122 for more details: https://www.ros.org/reps/rep-0122.html#filesystem-layout
 
     if let Ok(ament_prefix_paths) = ament_prefix_var {
-        for ament_prefix_path in ament_prefix_paths.split(':').map(PathBuf::from) {
+        for ament_prefix_path in ament_prefix_paths.split(':').map(Path::new) {
             // Locate the ament index
             let ament_index = ament_prefix_path.join("share/ament_index/resource_index/packages");
             if !ament_index.is_dir() {
                 continue;
             }
 
-            // Base directory all headers will be
+            // Old-style include directory
             let include_dir = ament_prefix_path.join("include");
 
             // Including the old-style packages
-            // todo(jhdcs): Remove this when old-style installation paths are fully removed
-            builder = builder.clang_arg(format!("-isystem{}", include_dir.to_string_lossy()));
+            builder = builder.clang_arg(format!("-isystem{}", include_dir.display()));
 
             // Search for and include new-style-converted package paths
             for dir_entry in read_dir(&ament_index).unwrap().filter_map(|p| p.ok()) {
-                let package = dir_entry.file_name().to_string_lossy().to_string();
-                let project_include_dir = include_dir.join(&package);
+                let package = dir_entry.file_name();
+                let package_include_dir = include_dir.join(&package);
 
-                if project_include_dir.is_dir() {
-                    // If the project_include_dir has another directory within it with the same name as the project
-                    // then this is a new-style installation path
-                    // todo(jhdcs): remove this check when no more packages use the old style
-                    let new_style_include_dir = project_include_dir.join(&package);
+                if package_include_dir.is_dir() {
+                    let new_style_include_dir = package_include_dir.join(&package);
 
                     // CycloneDDS is a special case - it needs to be included as if it were a new-style path, but
                     // doesn't actually have a secondary folder within it called "CycloneDDS"
-                    // todo(jhdcs): if this changes in future, remove this check
+                    // TODO(jhdcs): if this changes in future, remove this check
                     if package == "CycloneDDS" || new_style_include_dir.is_dir() {
-                        builder = builder.clang_arg(format!(
-                            "-isystem{}",
-                            project_include_dir.to_string_lossy()
-                        ));
+                        builder =
+                            builder.clang_arg(format!("-isystem{}", package_include_dir.display()));
                     }
                 }
             }
 
             // Link the native libraries
             let library_path = ament_prefix_path.join("lib");
-            println!(
-                "cargo:rustc-link-search=native={}",
-                library_path.to_string_lossy()
-            );
+            println!("cargo:rustc-link-search=native={}", library_path.display());
         }
     }
 
--- a/rclrs/build.rs
+++ b/rclrs/build.rs
@@ -1,30 +1,40 @@
 extern crate bindgen;
 
 use std::env;
-use std::path::{PathBuf, Path};
+use std::path::{Path, PathBuf};
 
-fn generate_clang_include_arg(ament_prefix_path: &str) -> Vec<String> {
+fn generate_clang_include_arg(ament_prefix_path: &str) -> Option<Vec<String>> {
     let ament_prefix_path = Path::new(ament_prefix_path);
     if ament_prefix_path.exists() && ament_prefix_path.is_dir() {
         let include_dir = ament_prefix_path.join(Path::new("include"));
         if include_dir.exists() {
-            let mut flags = Vec::<String>::new();
-            flags.push(format!("-I{}", include_dir.to_str().unwrap()));
             // If the path exists, there should be a directory name. Otherwise, abort
             let project_name = ament_prefix_path.file_name().unwrap();
+
+            // Always include top-level include dir
+            let mut clang_flags = Vec::<String>::new();
+            clang_flags.push(format!("-I{}", include_dir.to_str().unwrap()));
+            
+            // Generate potential include path
             let nested_project_include_path = include_dir.join(Path::new(project_name));
-            println!("\nNested project include path: {}", nested_project_include_path.to_str().unwrap());
+            println!(
+                "\nNested project include path: {}",
+                nested_project_include_path.to_str().unwrap()
+            );
+
             if nested_project_include_path.exists() {
                 println!("Nested path exists!");
                 // The path was created from only `str` objects, this should be safe to do
-                flags.push(format!("-I{}", nested_project_include_path.to_str().unwrap()));
-            }
+                let generated_arg = format!("-I{}", nested_project_include_path.to_str().unwrap());
+                println!("Generated clang_arg: {}\n", generated_arg);
+                clang_flags.push(generated_arg)
+            } 
 
-            return flags;
+            return Some(clang_flags);
         }
     }
 
-    return Vec::<String>::new();
+    None
 }
 
 fn main() {
@@ -53,9 +63,8 @@ fn main() {
 
     if let Ok(ament_prefix_paths) = ament_prefix_var {
         for ament_prefix_path in ament_prefix_paths.split(':') {
-            for clang_arg in generate_clang_include_arg(ament_prefix_path) {
-                println!("Generated clang_arg: {}\n", clang_arg);
-                builder = builder.clang_arg(clang_arg);
+            if let Some(clang_args) = generate_clang_include_arg(ament_prefix_path) {
+                builder = builder.clang_args(clang_args);
             }
             println!("cargo:rustc-link-search=native={}/lib", ament_prefix_path);
         }
--- a/rclrs/build.rs
+++ b/rclrs/build.rs
@@ -1,7 +1,8 @@
 extern crate bindgen;
 
 use std::env;
-use std::path::PathBuf;
+use std::fs::read_dir;
+use std::path::{Path, PathBuf};
 
 fn main() {
     let mut builder = bindgen::Builder::default()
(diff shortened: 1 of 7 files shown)


nnmm:
Could you also restore the Rolling CI job that was disabled in https://github.com/ros2-rust/ros2_rust/pull/73? (First file in the diff)

rclrs/build.rs
```diff
@@ -29,7 +63,9 @@ fn main() {
 
     if let Ok(ament_prefix_paths) = ament_prefix_var {
         for ament_prefix_path in ament_prefix_paths.split(':') {
-            builder = builder.clang_arg(format!("-I{}/include", ament_prefix_path));
+            if let Some(clang_args) = generate_clang_include_arg(ament_prefix_path) {
```
nnmm, line 66: Could the `generate_clang_include_arg()` function be simply replaced by
```
builder = builder.clang_arg(format!("-I{}/include", ament_prefix_path));
builder = builder.clang_arg(format!("-I{}/include/rcl", ament_prefix_path));
```
? After all it's only rcl that has this problem.

esteve (changes requested):

rclrs/build.rs
```diff
@@ -17,7 +51,7 @@ fn main() {
         .allowlist_function("rmw_.*")
         .allowlist_function("rcutils_.*")
         .allowlist_var("rcl_.*")
-        .allowlist_var("rcl_.*")
+        .allowlist_var("rmw_.*")
```
esteve, line 54: Good catch! :+1:

rclrs/build.rs
jhdcs: I thought that since `rcl` is adopting this change due to concerns about overlays, other packages would soon follow the new format. And I might be misremembering, but I thought a few already had... I could be wrong though.

nnmm: Please revert to the simpler format. Besides the extra lines of code, I can see the following issues with the "general" solution:
- You take the directory name of the prefix path, but for --merge-install, that would be e.g. "rolling", not the package name, right?
- It makes the whole include system looser, i.e. if a package ever decides to both provide `<package_name>/foo.h` and `foo.h`, you may get `<package_name>/foo.h`  even though you asked for `foo.h`.

jhdcs: So the trouble I'm running into, is that I have to keep extending the number of "special" include statements to get things to compile: In addition to `rcl` needing that special include pattern, I've also had to do `rcutils`, `rmw`, `rcl_yaml_param_parser`, and `rosidl_runtime_c` (so far, I'm still going through them), which was why I did that generic function.

If the generic function as it is written currently is not correct, is there a better way to write it?


> * You take the directory name of the prefix path, but for --merge-install, that would be e.g. "rolling", not the package name, right?

I don't _think_ so? But I use `--symlink-install`, so I'm not sure...

nnmm: > I don't think so? But I use --symlink-install, so I'm not sure...

`AMENT_PREFIX_PATH` is `/opt/ros/rolling` on a merged installation (like the one in my environment), so `project_name` will be `rolling`. So I don't think this approach will work, but you can try it out if you're not convinced (I don't have a Rolling installation ready at the moment).

Annoying that there are more packages than `rcl` with this issue. I guess we have to add all of them, at least I can't think of a cleaner solution right now.

rclrs/build.rs
```diff
@@ -63,9 +29,26 @@ fn main() {
 
     if let Ok(ament_prefix_paths) = ament_prefix_var {
         for ament_prefix_path in ament_prefix_paths.split(':') {
-            if let Some(clang_args) = generate_clang_include_arg(ament_prefix_path) {
-                builder = builder.clang_args(clang_args);
+            // Should not panic, since this was originally a valid UTF-8 String
+            builder = builder.clang_arg(format!("-I{}/include", ament_prefix_path));
+
+            // If this is RCL or rcutils, we need to add a special include
```
nnmm, line 35: Can we simply do
```
for pkg in &["rcl", "rcutils", etc] {
    builder = builder.clang_arg(format!("-I{}/include/{}", ament_prefix_path, pkg))
}
```
? I'm still not convinced that this version, where the package name is assumed to be the last element of `AMENT_PREFIX_PATH`, will work.

jhdcs: If I do that, we'll have two for loops, one after the other. May not be a huge issue, but still.

When I built and ran this using `--merge-install` everything seemed to work fine. `AMENT_PREFIX_PATHS` seems to hold to the rule that the package name is on the end of each of the paths, though the paths are now in the `ros2_rolling/install` folder.

nnmm: Can you tell me what's in your `AMENT_PREFIX_PATH` when you build `rclrs`, and the output of `ls` for each of these directories?

nnmm: Not sure if you saw the comment in the chat – I verified that the nested loop is necessary and that it doesn't build when the `rcl` etc. packages are part of a `--merge-install` install. So I propose to add the nested loop.

A possible performance optimization would be to add only those paths that exist.

jhdcs:
Major thanks to @sloretz for the assistance/algorithm for this latest version!

rclrs/build.rs
```diff
@@ -27,10 +28,68 @@ fn main() {
...
+        }) {
+            // Locate the ament index
+            let mut ament_index = ament_prefix_path.clone();
+            ament_index.push("share/ament_index/resource_index/packages");
+            if !ament_index.is_dir() {
+                continue;
+            }
+
+            // Including the old-style packages
+            // todo<jhassold>: Remove this when old-style installation paths are fully removed
```
esteve, line 61: Is `todo<username>` the accepted syntax for TODO comments in Rust? I've always used `TODO(username)` in all languages

rclrs/build.rs
```diff
@@ -27,10 +28,68 @@ fn main() {
...
+            {
+                let package = package_index_path.file_name().unwrap().to_string_lossy();
+
+                let mut include_dir = ament_prefix_path.clone();
+                include_dir.push(format!("include/{}", package.to_string()));
+
+                if include_dir.is_dir() {
+                    let mut new_style_include_dir = include_dir.clone();
+                    new_style_include_dir.push(package.to_string());
+                    if package == "CycloneDDS" || new_style_include_dir.is_dir() {
```
esteve, line 81: This looks a bit fragile, is there any other way than to address by package name? If there isn't, could you add a note here? Thanks.

rclrs/build.rs
```diff
@@ -27,10 +28,68 @@ fn main() {
...
+            let path = Path::new(p).to_path_buf();
+            if path.is_dir() {
+                Some(path)
+            } else {
+                None
+            }
+        }) {
+            // Locate the ament index
+            let mut ament_index = ament_prefix_path.clone();
+            ament_index.push("share/ament_index/resource_index/packages");
```
esteve, line 55: Will this work on Windows? Can you change this to a `Path` instead to make sure it's cross platform?

esteve (changes requested):

rclrs/build.rs
```diff
@@ -27,10 +28,68 @@ fn main() {
...
+    // > AMENT_PREFIX_PATH/lib  (For the link libraries)
+    //
+    // See REP 122 for more details: https://www.ros.org/reps/rep-0122.html#filesystem-layout
+
     if let Ok(ament_prefix_paths) = ament_prefix_var {
-        for ament_prefix_path in ament_prefix_paths.split(':') {
-            builder = builder.clang_arg(format!("-I{}/include", ament_prefix_path));
-            println!("cargo:rustc-link-search=native={}/lib", ament_prefix_path);
+        for ament_prefix_path in ament_prefix_paths.split(':').filter_map(|p| {
+            let path = Path::new(p).to_path_buf();
```
nwn, line 46:
```suggestion
            let path = PathBuf::from(p);
```

rclrs/build.rs
nwn, line 55:
```suggestion
            let ament_index = ament_prefix_path.join("share/ament_index/resource_index/packages");
```

rclrs/build.rs
```diff
@@ -27,10 +28,68 @@ fn main() {
...
+    // # ALGORITHM #
+    // Search through ament index at AMENT_PREFIX_PATH../share/ament_index/resource_index/packages/ to find packages to include
+    // Packages will be located at (for each path in AMENT_PREFIX_PATHS):
+    // > AMENT_PREFIX_PATH/include/ (For the header files) (OLD STYLE)
+    //      -isystemAMENT_PREFIX_PATH/include
+    // > AMENT_PREFIX_PATH/include/project_name (NEW STYLE)
+    //      -isystemAMENT_PREFIX_PATH/include/project_name (if and only if it has a folder inside of it also named project_name)
+    //      -isystemAMENT_PREFIX_PATH/include/CycloneDDS (special case, match for this)
+    // -OR-
+    // > AMENT_PREFIX_PATH/lib  (For the link libraries)
```
nnmm, line 40:
```suggestion
    // For each prefix in ${AMENT_PREFIX_PATH}:
    // Search through ament index at ${prefix}/share/ament_index/resource_index/packages/ to find packages to include
    // The include root will be located at either:
    // - ${prefix}/include/ (old style)
    // - ${prefix}include/package_name (new style)
    // - ${prefix}/include/CycloneDDS (special case, match for this)
    // End of loop
    // Compiled libraries are always at ${prefix}/lib
```

rclrs/build.rs
```diff
@@ -27,10 +28,68 @@ fn main() {
...
+            // Search for and include new-style-converted packages
+            for package_index_path in read_dir(&ament_index)
+                .unwrap()
+                .filter_map(|p| p.ok().map(|p| p.path()))
+                .filter(|p| p.file_name().is_some())
+            {
+                let package = package_index_path.file_name().unwrap().to_string_lossy();
+
+                let mut include_dir = ament_prefix_path.clone();
+                include_dir.push(format!("include/{}", package.to_string()));
```
nwn, line 76:
```suggestion
                let include_dir = ament_prefix_path.join(format!("include/{}", package.to_string()));
```

rclrs/build.rs
```diff
@@ -27,10 +28,68 @@ fn main() {
...
+
+                if include_dir.is_dir() {
+                    let mut new_style_include_dir = include_dir.clone();
+                    new_style_include_dir.push(package.to_string());
```
nwn, line 80:
```suggestion
                    let new_style_include_dir = include_dir.join(package.as_ref());
```

nwn (commented):
Added some (untested) suggestions that might be more idiomatic.

rclrs/build.rs
nnmm: I'm pretty sure it does. Also, no need to clone – just `let mut ament_index = ament_prefix_path.join("share/ament_index/...")` works too.

rclrs/build.rs
```diff
@@ -27,10 +28,68 @@ fn main() {
...
+    //
+    // See REP 122 for more details: https://www.ros.org/reps/rep-0122.html#filesystem-layout
+
     if let Ok(ament_prefix_paths) = ament_prefix_var {
-        for ament_prefix_path in ament_prefix_paths.split(':') {
-            builder = builder.clang_arg(format!("-I{}/include", ament_prefix_path));
-            println!("cargo:rustc-link-search=native={}/lib", ament_prefix_path);
+        for ament_prefix_path in ament_prefix_paths.split(':').filter_map(|p| {
+            let path = Path::new(p).to_path_buf();
+            if path.is_dir() {
```
nnmm, line 47: This check is redundant with the check further down – If the prefix doesn't exist, the ament index won't exist either. I'd remove it.

rclrs/build.rs
```diff
@@ -27,10 +28,68 @@ fn main() {
...
+                ament_prefix_path.to_string_lossy()
+            ));
+
+            // Search for and include new-style-converted packages
+            for package_index_path in read_dir(&ament_index)
+                .unwrap()
+                .filter_map(|p| p.ok().map(|p| p.path()))
+                .filter(|p| p.file_name().is_some())
+            {
+                let package = package_index_path.file_name().unwrap().to_string_lossy();
```
nnmm, line 73:
```suggestion
            for dir_entry in read_dir(&ament_index)
                .unwrap()
                .filter_map(|p| p.ok())
            {
                let package = dir_entry.file_name().to_string_lossy();
```

rclrs/build.rs
nnmm, line 80:
```suggestion
                    let new_style_include_dir = include_dir.join(&package.);
```

rclrs/build.rs
nnmm, line 76:
```suggestion
                let include_dir = ament_prefix_path.join("include").join(&package);
```

rclrs/build.rs
```diff
@@ -27,10 +28,68 @@ fn main() {
...
+
+                let mut include_dir = ament_prefix_path.clone();
+                include_dir.push(format!("include/{}", package.to_string()));
+
+                if include_dir.is_dir() {
+                    let mut new_style_include_dir = include_dir.clone();
+                    new_style_include_dir.push(package.to_string());
+                    if package == "CycloneDDS" || new_style_include_dir.is_dir() {
+                        builder =
+                            builder.clang_arg(format!("-isystem{}", include_dir.to_string_lossy()));
```
nnmm, line 83:
```suggestion
                            builder.clang_arg(format!("-isystem{}", include_dir.display()));
```

nnmm:
I left a couple of suggestions as well. I didn't try if they compile, so I'm not 100% sure they all work.

jhdcs:
Thank you @esteve, @nnmm , and @nwn **so** much for your reviews and suggestions! I _THINK_ I've managed to implement them all... Please let me know if I missed one, or if you see any other things that can be improved!

rclrs/build.rs
```diff
@@ -27,10 +28,65 @@ fn main() {
...
+            if !ament_index.is_dir() {
+                continue;
+            }
+
+            // Base directory all headers will be
+            let include_dir = ament_prefix_path.join("include");
+
+            // Including the old-style packages
+            // todo(jhdcs): Remove this when old-style installation paths are fully removed
+            builder = builder.clang_arg(format!("-isystem{}", include_dir.to_string_lossy()));
```
nnmm, line 59:
```suggestion
            builder = builder.clang_arg(format!("-isystem{}", include_dir.display()));
```

rclrs/build.rs
```diff
@@ -27,10 +28,65 @@ fn main() {
...
+    // End of loop
+    // Compiled libraries are always at ${prefix}/lib
+    //
+    // See REP 122 for more details: https://www.ros.org/reps/rep-0122.html#filesystem-layout
+
     if let Ok(ament_prefix_paths) = ament_prefix_var {
-        for ament_prefix_path in ament_prefix_paths.split(':') {
-            builder = builder.clang_arg(format!("-I{}/include", ament_prefix_path));
-            println!("cargo:rustc-link-search=native={}/lib", ament_prefix_path);
+        for ament_prefix_path in ament_prefix_paths.split(':').map(PathBuf::from) {
```
nnmm, line 47:
```suggestion
        for ament_prefix_path in ament_prefix_paths.split(':').map(Path::new) {
```
to avoid a clone.

rclrs/build.rs
```diff
@@ -27,10 +28,65 @@ fn main() {
...
+
+            // Search for and include new-style-converted package paths
+            for dir_entry in read_dir(&ament_index).unwrap().filter_map(|p| p.ok()) {
+                let package = dir_entry.file_name().to_string_lossy().to_string();
+                let project_include_dir = include_dir.join(&package);
+
+                if project_include_dir.is_dir() {
+                    // If the project_include_dir has another directory within it with the same name as the project
+                    // then this is a new-style installation path
+                    // todo(jhdcs): remove this check when no more packages use the old style
```
nnmm, line 69: Nit: I think this doesn't need to be a todo, since we still want to support Foxy and Galactic, right?

rclrs/build.rs
```diff
@@ -27,10 +28,65 @@ fn main() {
...
+                        ));
+                    }
+                }
+            }
+
+            // Link the native libraries
+            let library_path = ament_prefix_path.join("lib");
+            println!(
+                "cargo:rustc-link-search=native={}",
+                library_path.to_string_lossy()
```
nnmm, line 88:
```suggestion
                library_path.display()
```

rclrs/build.rs
```diff
@@ -27,10 +28,65 @@ fn main() {
...
+            let include_dir = ament_prefix_path.join("include");
+
+            // Including the old-style packages
+            // todo(jhdcs): Remove this when old-style installation paths are fully removed
+            builder = builder.clang_arg(format!("-isystem{}", include_dir.to_string_lossy()));
+
+            // Search for and include new-style-converted package paths
+            for dir_entry in read_dir(&ament_index).unwrap().filter_map(|p| p.ok()) {
+                let package = dir_entry.file_name().to_string_lossy().to_string();
+                let project_include_dir = include_dir.join(&package);
```
nnmm, line 64:
```suggestion
                let package_include_dir = include_dir.join(&package);
```

rclrs/build.rs
```diff
@@ -27,10 +28,65 @@ fn main() {
...
+                    // todo(jhdcs): remove this check when no more packages use the old style
+                    let new_style_include_dir = project_include_dir.join(&package);
+
+                    // CycloneDDS is a special case - it needs to be included as if it were a new-style path, but
+                    // doesn't actually have a secondary folder within it called "CycloneDDS"
+                    // todo(jhdcs): if this changes in future, remove this check
+                    if package == "CycloneDDS" || new_style_include_dir.is_dir() {
+                        builder = builder.clang_arg(format!(
+                            "-isystem{}",
+                            project_include_dir.to_string_lossy()
```
nnmm, line 78:
```suggestion
                            package_include_dir.display()
```

rclrs/build.rs
```diff
@@ -27,10 +28,65 @@ fn main() {
...
+            let ament_index = ament_prefix_path.join("share/ament_index/resource_index/packages");
+            if !ament_index.is_dir() {
+                continue;
+            }
+
+            // Base directory all headers will be
+            let include_dir = ament_prefix_path.join("include");
+
+            // Including the old-style packages
+            // todo(jhdcs): Remove this when old-style installation paths are fully removed
```
esteve, line 58: Please use uppercase for `TODO`, `NOTE`, etc. they pop out when reading the code and keep the same style as the ROS2 codebase. Thanks :slightly_smiling_face:

rclrs/build.rs
```diff
@@ -27,10 +28,65 @@ fn main() {
...
-            builder = builder.clang_arg(format!("-I{}/include", ament_prefix_path));
-            println!("cargo:rustc-link-search=native={}/lib", ament_prefix_path);
+        for ament_prefix_path in ament_prefix_paths.split(':').map(PathBuf::from) {
+            // Locate the ament index
+            let ament_index = ament_prefix_path.join("share/ament_index/resource_index/packages");
+            if !ament_index.is_dir() {
+                continue;
+            }
+
+            // Base directory all headers will be
```
nnmm, line 54:
```suggestion
            // Old-style include directory
```

rclrs/build.rs
```diff
@@ -27,10 +28,65 @@ fn main() {
...
 
+    // #############
+    // # ALGORITHM #
+    // #############
+    //
+    // For each prefix in ${AMENT_PREFIX_PATH}:
+    // Search through ament index at ${prefix}/share/ament_index/resource_index/packages/ to find packages to include
+    // The include root will be located at either:
+    // - ${prefix}/include/ (old style)
+    // - ${prefix}/include/package_name (new style)
```
nnmm, line 39:
```suggestion
    // - ${prefix}/include/${package_name} (new style)
```

esteve (dismissed):

nnmm (commented):
I did a second iteration – sorry, I didn't quite catch everything yesterday. Only nitpicks. After this iteration, should be good to go!

esteve:
@jhdcs I only have one more comment (`todo` vs `TODO` comments), but approved this PR anyway so you can merge it after the changes are made. Thank you so much for all this work and for the patience :slightly_smiling_face:

jhdcs:
Ok, I _think_ I got everything now. Thank you for all the suggestions! It's always great to learn better ways of doing things!

> Thank you so much for all this work and for the patience

Of course! It's no trouble at all! Sorry it took so long, but I want this to be the best it can be before going into the library!

nnmm (approved):
Thanks for being so patient & responsive! LGTM.

PR #78, opened 2022-02-25
Explain how to use cargo in README

nnmm:


README.md
```diff
@@ -61,9 +61,27 @@ colcon build --packages-up-to rclrs_examples
 
 It's normal to see a `Some selected packages are already built in one or more underlay workspace` warning. This is because the standard message definitions that are part of ROS 2 need to be regenerated in order to create Rust bindings.
 
-Now you can just run a bunch of examples.
+### Building with `cargo`
+As an alternative to `colcon`, Rust packages can be built with pure `cargo`.
 
-### Publisher and subscriber
+However, this will not work out of the box, since the `Cargo.toml` files contain dependencies like `rclrs = "*"`, even though `rclrs` is not published on crates.io. This is intentional and follows ROS 2's principle for packages to reference their dependencies only with their name, and not with their path. At build-time, these dependencies are resolved to a path to the local package by `colcon`, and written into `.cargo/config.toml`. Therefore, the package in question should once be built with `colcon` initially, and after that `cargo` will be able to use the `.cargo/config.toml` file to find all dependencies.
```
jhdcs, line 67:
```suggestion
However, this will not work out of the box, since the `Cargo.toml` files contain dependencies like `rclrs = "*"`, even though `rclrs` is not published on crates.io. This is intentional and follows ROS 2's principle for packages to reference their dependencies only with their name, and not with their path. At build-time, these dependencies are resolved to a path to the local package by `colcon`, and written into `.cargo/config.toml`. Therefore, the package in question should be built with `colcon` once, and after that `cargo` will be able to use the `.cargo/config.toml` file to find all dependencies.
```

jhdcs (dismissed):
Only one area where I thought re-wording might be good, but it's a subjective thing, so I won't require it.

Looks good!

README.md
nnmm: Done.

jhdcs (approved):

PR #80, opened 2022-03-05
Message generation refactoring

nnmm:
Previously, only messages consisting of basic types and strings were supported. Now, all message types will work, including those that have fields of nested types, bounded types, or arrays.

# Changes
- The "rsext" library is deleted
- Unused messages in "rosidl_generator_rs" are deleted
- There is a new package, "rosidl_runtime_rs", see below
- The RMW-compatible messages from C, which do not require an extra conversion step, are exposed in addition to the "idiomatic" messages
- Publisher and subscription are changed to work with both idiomatic and rmw types, through the unifying `Message` trait

On `rosidl_runtime_rs`: This package is the successor of `rclrs_msg_utilities` package, but doesn't have much in common with it anymore.
It provides common types and functionality for messages. The `String` and `Sequence` types and their variants in that package essentially wrap C types from the `rosidl_runtime_c` package and C messages generated by the "rosidl_generator_c" package.
A number of functions and traits are implemented on these types, so that they feel as ergonomic as possible, for instance, a `seq!` macro for creating a sequence. There is also some documentation and doctests.

The memory for the (non-pretty) message types is managed by the C allocator.

# Not yet implemented
- long double
- constants
- Services/clients
- @verbatim comments
- ndarray for sequences/arrays of numeric types
- implementing `Eq`, `Ord` and `Hash` when a message contains no floats

# How to review
I know that this is a really large PR. If desired, I can split it up for reviewing. Here are some steps I suggest to get familiar with the changes:
- Run the `message_demo` and read the code
- Generate `cargo doc`s for the `rosidl_runtime_rs` crate and read them
- Dive into `msg.rs.em` and `__init__.py` files


nnmm:
There is currently a limitaiton where I'm using copy functions (https://github.com/ros2/rosidl/pull/650/) that are not yet in Foxy and Galactic. Jacob wanted to take a look at backporting those. However, I could also try to do a polyfill for them if we don't want to wait that long.

jhdcs:
> There is currently a limitaiton where I'm using copy functions ([ros2/rosidl#650](https://github.com/ros2/rosidl/pull/650)) that are not yet in Foxy and Galactic. Jacob wanted to take a look at backporting those. However, I could also try to do a polyfill for them if we don't want to wait that long.

What would be involved in the creation of a "polyfill"? I'm not familiar with the process.

nnmm:
> > There is currently a limitaiton where I'm using copy functions ([ros2/rosidl#650](https://github.com/ros2/rosidl/pull/650)) that are not yet in Foxy and Galactic. Jacob wanted to take a look at backporting those. However, I could also try to do a polyfill for them if we don't want to wait that long.
> 
> What would be involved in the creation of a "polyfill"? I'm not familiar with the process.

I'd try implementing the functions from that PR in Rust, using `rcutils_get_default_allocator()`. So `rosidl_runtime_rs` would need to link against `rcutils` for the time being.

esteve:
@nnmm awesome work, thanks! CI seems to be failing because of a missing message, the file that's causing the problem is `NestedType.msg`, any reference to it should be removed if we're no longer using it.

nnmm:
> @nnmm awesome work, thanks! CI seems to be failing because of a missing message, the file that's causing the problem is `NestedType.msg`, any reference to it should be removed if we're no longer using it.

Oh, forgot to add it. CI will still fail because of the issue mentioned above (the copy functions being Rolling-only for now). I still opened this PR for review to parallelize reviewing + work on that issue.

jhdcs:
> > > There is currently a limitaiton where I'm using copy functions ([ros2/rosidl#650](https://github.com/ros2/rosidl/pull/650)) that are not yet in Foxy and Galactic. Jacob wanted to take a look at backporting those. However, I could also try to do a polyfill for them if we don't want to wait that long.
> > 
> > 
> > What would be involved in the creation of a "polyfill"? I'm not familiar with the process.
> 
> I'd try implementing the functions from that PR in Rust, using `rcutils_get_default_allocator()`. So `rosidl_runtime_rs` would need to link against `rcutils` for the time being.

I think implementing the functions in Rust for now might be good, provided the work isn't too difficult. That way, if I get pulled away to work on other projects you won't be stuck waiting on me for this to get merged/working.

rclrs/src/node/mod.rs
```diff
@@ -96,7 +97,7 @@ impl Node {
         qos: QoSProfile,
     ) -> Result<Publisher<T>, RclReturnCode>
     where
-        T: MessageDefinition<T>,
```
esteve, line 99: Why was the generic type parameter dropped?

rclrs_example_msgs/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+cmake_minimum_required(VERSION 3.5)
```
esteve, line 1: Instead of adding `rclrs_example_msgs`, I'd prefer using the `example_interfaces` or `test_interface_files` packages.

rosidl_generator_rs/cmake/rosidl_generator_rs_generate_interfaces.cmake
```diff
@@ -162,93 +155,30 @@ set_property(
...
-      @ONLY)
-
-    install(
-      FILES "${CMAKE_CURRENT_BINARY_DIR}/share/${PROJECT_NAME}/rust/Cargo.toml"
-      DESTINATION share/${PROJECT_NAME}/rust/
-    )
   endif()
 endforeach()
 
+configure_file("${rosidl_generator_rs_TEMPLATE_DIR}/Cargo.toml.in"
```
esteve, line 173: Not necessary to change this, but what do you think of using `.em` templates for `Cargo.toml` instead of CMake's limited template support?

esteve (changes requested):

esteve:
@nnmm I just did a first pass. This is really awesome, you truly outdid yourself dude! I'll do a more extensive review later, but this looks great.

rclrs_examples/src/minimal_publisher.rs
```diff
@@ -22,7 +22,7 @@ fn main() -> Result<(), Error> {
     while context.ok()? {
         message.data = format!("Hello, world! {}", publish_count);
         println!("Publishing: [{}]", message.data);
-        publisher.publish(&message)?;
+        publisher.publish(message.clone())?;
```
esteve, line 25: Would it be possible to implement the `Message` trait for references as well? `rclcpp` does not require developers to create a new copy of a message before publishing it, and it'd be cool if Rust developers had the same experience.

esteve (changes requested):

rclrs/src/node/mod.rs
nnmm: I just didn't see any reason why it should be there.

rosidl_generator_rs/cmake/rosidl_generator_rs_generate_interfaces.cmake
nnmm: Sure, why not

rclrs_examples/src/minimal_publisher.rs
nnmm: Maybe. But one thing I like about having it implemented only for owned values is that it encourages a more efficient coding style by making the clones explicit.

rclrs_example_msgs/msg/VariousTypes.msg
```diff
@@ -0,0 +1,34 @@
...
+
+# Array/sequence of primitive type
+float32[3] float_array [1.0, 2.0, 3.0]
+float32[<=3] float_seq_bounded [4.0, 5.0]
+float32[] float_seq_unbounded [6.0]
+
+# String types
+string string_member "Χαίρετε 你好"
+wstring wstring_member "αντίο σου 再见"
+string<=3 bounded_string_member "äöü"
```
nnmm, line 16: I was surprised by this. The length of a string member seems to be measured in unicode code points or something, not bytes ("äöü" has six bytes). I still need to find the exact criterion and replicate it in `BoundedString`.

rclrs_examples/src/minimal_publisher.rs
esteve: However, that's a performance problem. If a user has a large point cloud or an image, cloning it makes little sense, hence why `rclcpp` uses pointers and not instances. And one of the benefits of Rust is the borrow checker, you can guarantee that references don't go out of scope, so you don't have to waste memory.

rosidl_runtime_rs/src/sequence.rs
```diff
@@ -0,0 +1,618 @@
...
+        let num_remaining = it.size_hint().0;
+        if num_remaining > 0 {
+            let new_size = self.size.saturating_add(num_remaining);
+            resize(self, new_size);
+        }
+        while let Some(item) = it.next() {
+            // If there is no more capacity for the next element, resize to the
+            // next power of two.
+            //
+            // A pedantic implementation would check for usize overflow here, but
```
jhdcs, line 161: Should we be pedantic, though? If `ros2_rust` is being used on safety-critical systems, they may want this sort of check. Or is there a better value to test against? What would the performance impact be?

rclrs_example_msgs/msg/VariousTypes.msg
nnmm: A string. For the wstring, it would be expected that it can fit "äöü".

rclrs_examples/src/minimal_publisher.rs
nnmm: So, there's two different things we could do:
1. Implement `Message` for `&IdiomaticType`
2. Implement `Message` for `&RmwType`

1 is less efficient than the existing impl for an owned `IdiomaticType`. For instance, for an array, when the source array is owned, you can use `Array::map`, but when it's borrowed, you need to collect into a temporary `Vec` and call `try_into()` to create an array again. Bounded strings and sequences would need to be cloned from a borrowed message, instead of moved.

2  could be equally efficient as the existing impl for an owned `RmwType`, i.e. a no-op, at least I think it should be possible with `Cow` or something.

I mostly see the value for 2. Do you also want 1?

rosidl_runtime_rs/src/traits.rs
```diff
@@ -0,0 +1,133 @@
...
+///
+/// `create()` simply allocates space for the message itself, and calls `init()`.
+///
+/// `destroy()` simply deallocates the message itself, and calls `fini()`.
+///
+/// Memory ownership by C is achieved by calling `init()` when any string or sequence is created,
+/// as well as in the `Default` impl for messages.
+/// User code can still create messages explicitly, which will not call `init()`, but this is not a
+///  problem, since nothing is allocated this way.
+/// The `Drop` impl for any sequence or string will call `fini()`.
```
jhdcs, line 122: Exemplary documentation!

jhdcs (commented):
Mostly a few comments/clarifications that I'd like to know about. Looks excellent, though!

rosidl_generator_rs/resource/msg.rs.em
```diff
@@ -1,123 +1,219 @@
...
-@[    if isinstance(member.type, Array)]@
-@[    elif isinstance(member.type, AbstractGenericString)]@
-      let ptr = @(package_name)_@(subfolder)_@(convert_camel_case_to_lower_case_underscore(type_name))_@(member.name)_read_handle(_message_handle);
-      self.@(get_rs_name(member.name)) = CStr::from_ptr(ptr).to_string_lossy().into_owned();
-@[    elif isinstance(member.type, BasicType)]@
-      self.@(get_rs_name(member.name)) = @(package_name)_@(subfolder)_@(convert_camel_case_to_lower_case_underscore(type_name))_@(member.name)_read_handle(_message_handle);
-@[    elif isinstance(member.type, AbstractSequence)]@
-@[    end if]@
-@[end for]@
+@#    // This is safe since a zeroed bit pattern always forms a valid message.
```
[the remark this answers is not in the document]
nnmm: Good point. I can start by adding `// SAFETY: <reason>` annotations to every `unsafe` in this PR.

rclrs_example_msgs/CMakeLists.txt
nnmm: We need https://github.com/ros2/test_interface_files/pull/18 to make that work, I believe

rosidl_generator_rs/resource/msg.rs.em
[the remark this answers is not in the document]
nnmm: Done.

rosidl_runtime_rs/src/traits.rs
nnmm: Thanks!

rosidl_runtime_rs/src/string.rs
```diff
@@ -0,0 +1,451 @@
...
+                    panic!("Sinit failed");
+                }
+                msg
+            }
+        }
+
+        impl Clone for $string {
+            fn clone(&self) -> Self {
+                let mut msg = Self::default();
+                // SAFETY: This is doing the same thing as rosidl_runtime_c__String__copy.
```
jhdcs, line 143: I'm not sure this comment addresses _why_ this is safe. I feel like these safety comments should be written so that someone won't need to dive into an external C library to verify, unless they really wanted to.

Something like: "This is safe because neither `self.data` nor `self.size` are mutated, and `self.data` is guaranteed to be pointing to valid memory by this point, and `self.size` is guaranteed to be correct"?

jhdcs (approved):
Just another comment, mostly a stylistic thing. Otherwise, this looks good to me!

Thank you very much for your work!

rclrs_examples/src/minimal_publisher.rs
esteve: I'd like to avoid copies anywhere in the code, otherwise we'll have to readd the first item in README that says that messages are copied.

esteve: Also, idiomatic types in `rclcpp` are passed as shared pointers, not as copies, so giving a worse experience to Rust developers would be unfair when it's technically possible.

esteve:
@nnmm given (RFC344) [https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis ] is there a need for `BoundedSequence` and `BoundedString`? Checking for length is done in setters in other ROS2 bindings instead of using a separate structure.

rclrs_examples/src/minimal_publisher.rs
nnmm: So you're saying there should be a `Message` impl for `&IdiomaticType` too? I can do that.

nnmm: The only problem is that this will require s lifetime parameter on the trait, making it less ergonomic. But I'll try it out.

nnmm:
> @nnmm given [RFC344](`https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis`) is there a need for `BoundedSequence` and `BoundedString`? Checking for length is done in setters in other ROS2 bindings instead of using a separate structure.

I'd say that direct access to a field is the most flexible and ergonomic option, and only when that for some reason is not allowed (e.g. because it would allow violating some invariant) are getters and setters needed.

This PR also has `TryFrom<&str>` for `BoundedString` and I will also add `TryFrom<&[T]>` for bounded sequences. With that, the pattern of building up a string/sequence outside and checking the length when it's inserted into the message is also covered.

nnmm:
@esteve Do you know how to turn off these `cpplint_rosidl_generated_rs`, `cppcheck_rosidl_generated_rs`, `uncrustify_rosidl_generated_rs` tests in CI?

rosidl_runtime_rs/build.rs
```diff
@@ -0,0 +1,13 @@
+use std::env;
+use std::path::Path;
+
+const AMENT_PREFIX_PATH: &'static str = "AMENT_PREFIX_PATH";
+
+fn main() {
+    if let Ok(ament_prefix_path_list) = env::var(AMENT_PREFIX_PATH) {
```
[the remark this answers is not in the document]
nnmm: Yeah, good point! I added an `else` clause that panics with a helpful message.

rclrs_example_msgs/msg/VariousTypes.msg
nnmm: Sending "🔥🔥🔥" (one unicode code point per emoji, 4 bytes each in UTF-8) works fine.

When I send "🤷🏻‍♂️🤷🏻‍♂️🤷🏻‍♂️" (has more than one unicode code point per emoji) or "abcd" in a bounded string, `ros2 topic echo` doesn't receive the message anymore, but the `rclrs` node still does. I'm surprised there is no lower-level check about the length of the string, seem like it's the client library's job.

I changed it to `chars().count()` since that's what mirrors the above behavior and Python's `len()` function for strings.

rclrs_example_msgs/CMakeLists.txt
nnmm: I'd propose to merge the version with the `rclrs_example_msgs` if the PR I linked isn't approved by the time the rest of the review is finished, and then later remove it once it is.

rosidl_runtime_rs/src/sequence.rs
nnmm: Yeah, good question. I'm not worried about performance impact, but about code complexity/overengineering.

I think ROS 2 messages will never reach this limit – certainly not on 64-bit machines, and I don't know if anyone is building ROS 2 for 32-bit systems. There is also a 1 GB message size limit, so even if we have a 32-bit system, there won't be any sequences with more than `usize/2` elements. Besides, there are probably plenty of unchecked additions in rcl and rmw that would cause trouble too.

So I propose to not be pedantic about usize overflow, but to just leave comments like this one when such an issue is encountered. If anyone ever really wants to use `ros2_rust` in a safety-critical context, they'll need to go through the code and refactor a whole lot of stuff anyway, and then at least there's a comment to point to potential issues.

rclrs_examples/src/minimal_publisher.rs
nnmm: The only ergonomic way I've found so far is to split up the `Message` trait:

```rust
pub trait Message {
    /// The corresponding RMW-compatible message type.
    type RmwMsg: RmwMessage;

    /// Converts the idiomatic message into a RMW-compatible message.
    fn into_rmw_message(self) -> Self::RmwMsg;
}

pub trait OwnedMessage: Message {
    /// Converts the RMW-compatible message into an idiomatic message.
}
``` 
`Message` is implemented by owned values and references, and `OwnedMessage` only by owned values. `Message` is used on the publishing side, `OwnedMessage` on the subscription side.

Does this architecture sound alright, and do you have suggestions for better names?

rosidl_generator_rs/cmake/rosidl_generator_rs_generate_interfaces.cmake
esteve: What does that mean? That you're going to switch to `.em` templates or not? Can you elaborate?

rclrs_examples/src/minimal_publisher.rs
esteve: > So you're saying there should be a Message impl for &IdiomaticType too? I can do that.

Yes, that's what I meant by implementing `Message` for references.

> Does this architecture sound alright, and do you have suggestions for better names?

A priori this look good :+1:

rclrs_example_msgs/msg/VariousTypes.msg
esteve: `chars().count()` makes sense to me

esteve (changes requested):

esteve:
> @esteve Do you know how to turn off these `cpplint_rosidl_generated_rs`, `cppcheck_rosidl_generated_rs`, `uncrustify_rosidl_generated_rs` tests in CI?

I'll look into it, neither of those tests should run on Rust files and there are no more C files to be checked.

rclrs_examples/src/minimal_publisher.rs
nnmm: Reworked it again and pushed it. Please see the last two commits.

rclrs/src/node/publisher.rs
```diff
@@ -94,17 +95,16 @@ where
         })
     }
 
-    pub fn publish(&self, message: &T) -> Result<(), RclReturnCode> {
```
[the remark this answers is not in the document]
nnmm: Done, in a sense.

README.md
```diff
@@ -33,9 +33,8 @@ Lots of things!
 
 ### Limitations
 
-- messages are deep-copied and this can be terribly inefficient for big messages like images; the current solution leverages C typesupport implementations and might benefit from a direct serialization/deserialization
```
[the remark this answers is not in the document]
nnmm: I hadn't pushed the changes to generalize publishing to borrowed messages yet. Now they're in, and I believe now the code is as efficient as it can be.

rclrs/src/node/mod.rs
```diff
@@ -109,7 +110,7 @@ impl Node {
         callback: F,
```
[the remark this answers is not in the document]
nnmm: Why would we have a redundant function like this? The current `Subscription` API can receive `Messages`, and both idiomatic and RMW-compatible messages are instances of that. Adding a second API that only supports the latter type is redundant and confuses only, imo. Uniformity is nice.

nnmm:
Thanks for fixing the linter @esteve. Now this PR is ready to merge from my side: I added the workaround for the `copy_sequence` functions in Foxy and Galactic until the backports are shipped, and changed the publishing + conversion logic to allow publishing of both borrowed and owned messages, for maximum efficiency.

rclrs/src/node/publisher.rs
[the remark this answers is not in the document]
esteve: What do you mean by "in a sense", are there any caveats?

rclrs/src/node/mod.rs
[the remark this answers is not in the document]
esteve: Precisely for uniformity with the other bindings, no other bindings accept these so-call rmw types, all of the support idiomatic types. Adding a second API clears that confusion by explicitly signaling that it's specific to `ros2-rust`. Rust, very much like Python, and unlike C++, encourages explicitness vs implicitness. Explicitness is nice too.

esteve:
> Thanks for fixing the linter @esteve. Now this PR is ready to merge from my side: I added the workaround for the `copy_sequence` functions in Foxy and Galactic until the backports are shipped, and changed the publishing + conversion logic to allow publishing of both borrowed and owned messages, for maximum efficiency.

Thanks, there is still some unanswered feedback, could you address it? CI is not passing either, do you know why?

rclrs/src/node/publisher.rs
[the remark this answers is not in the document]
nnmm: It's not `message: &T` now, it's `message: impl OwnedOrBorrowedMessage`, which accepts `T` and `&T`.

Suggestions for a better name welcome. In case you dislike this design, we could also split it up into two functions, `publish_owned()` and `publish_ref()`.

rosidl_generator_rs/cmake/rosidl_generator_rs_generate_interfaces.cmake
nnmm: I think it's a good idea, but I'd probably do it in the future instead of in this MR, since it's mostly unrelated.

esteve: Yeah, that sounds good for a future PR

rclrs_example_msgs/CMakeLists.txt
nnmm: Seems like bringing in `test_interface_files` is more complicated than we thought. Would you be fine with leaving in `rclrs_example_msgs` for now?

rclrs/src/node/mod.rs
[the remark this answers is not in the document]
esteve: > through less-than-canonical APIs

The RMW-compatible message API is already non-canonical, because it does not have any counterpart in the other bindings. APIs are not only about names, but also semantics, the arguments they accept and the values they return. Having a distinctly different name emphasizes that it behaves differently from the other bindings, and that users should expect differences.

> Besides, there is also lots of stuff that is specific to rclcpp, and none of it is marked as such (topic statistics, generic pub/sub, intraprocess, etc.).

Because the goal is that eventually that functionality might be incorporated into other bindings. For example, `rclc` could potentially get support for intraprocess.

Anyway, I rather not waste more time on this, but it's important to get APIs right early on, because changing them later is far more difficult

rclrs/src/node/publisher.rs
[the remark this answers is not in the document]
esteve: I don't get it, you seemed to be very keen on not having a separate create_subscription function for RMW-compatible types, but `publish_owned` and `publish_ref` are ok. Could you elaborate?

Let's keep it as is now, though, and we could refine the APIs later if needed, though it'll be more painful.

esteve (approved):

esteve:
@nnmm awesome job, thanks. I've approved these changes, but there are conflicts with `main` because of #81 I think removing `rclrs_msg_utilities` should suffice.

rclrs/src/node/publisher.rs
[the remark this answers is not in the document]
nnmm: Sure. The distinction between `publish_owned()` and `publish_ref()` is of course orthogonal to the question of which message type is published. But you're right that both are a "have two functions for doing the thing vs. having one more general function" question.

In this case here, I'm more on the fence between splitting vs unifying, since the more general function comes at the cost of a more complicated signature: The `MessageCow` trait. In the case of the extra `create_subscription` function, the `create_subscription` function would look the same whether or not we also have a `create_rmw_subscription`, so there's no extra complexity in the unified version. Being simpler is not the only argument in favor of one subscription function, but I'll keep that discussion in the other thread.

rclrs_examples/src/minimal_publisher.rs
nnmm: I'm resolving this thread since the approach of implementing `Message` for `&T` has been superseded by the `publish()` function taking a `T` and `&T` where `T: Message`, as discussed in this thread: https://github.com/ros2-rust/ros2_rust/pull/80#discussion_r824689568

esteve (approved):

rclrs/src/node/mod.rs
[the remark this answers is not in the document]
nnmm: As for counterparts in other bindings, here's my perspective for posterity:
- Python is slow anyway, and usability of a raw C message (= our "RMW-compatible" message) would be absymal from a Python perspective, so of course the messages in that language are an idiomatic version that are converted to/from the C message.
- C and C++ have the privilege of being natively RMW-compatible, so they can be both idiomatic and maximally efficient. They are the counterpart to the RMW-compatible *and* the idiomatic messages.
- I assume other language bindings also use their own idiomatic message types that are converted to/from C messages. However, I believe none of the other languages are quite as performance-focused as Rust/C/C++.

Hence, Rust is kind of unique, in that its users expect maximum performance but it can't achieve that (yet) by serializing/deserializing its standard library types to/from CDR directly. So you could argue that it would be canonical to use RMW-compatible messages because that's what the other performance-focused language bindings do. Or you could argue that it would be canonical to use idiomatic messages because that's what all language bindings do.

I think the solution in this PR, to offer both, isn't too bad. And I placed the idiomatic messages in the `<pkg>::msg` namespace and the RMW-compatible ones in `<pkg>::msg::rmw`, because they can't be both in the same namespace, but I believe we shouldn't hide the RMW-compatible messages more than that if it's not technically necessary. 

I don't think the time spent discussing on this is wasted. Hope I'm not too stubborn! If you insist, I'll add the separate function. Or I can also poll the ROS + Rust people I know to get more input. But since you approved, I take it that it's okay to merge as it is now and potentially discuss it again later.

nnmm:
Huge thanks for reviewing @esteve and also @jhdcs! :tada:

rclrs/src/node/mod.rs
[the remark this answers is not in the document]
mosteo: > * I assume other language bindings also use their own idiomatic message types that are converted to/from C messages. However, I believe none of the other languages are quite as performance-focused as Rust/C/C++.

Ada reuses the C data layout in memory, so there is no overhead involved. The price is somewhat un-idiomatic use when sequences are involved that I hope to alleviate at some point in the future.

PR #81, opened 2022-03-15
Check code formatting in CI

nnmm:
This check is placed before build & test because it is cheap.


jhdcs (approved):
Nice, short, and simple.

Looks good to me! I'm assuming the change in `rclrs_msg_utilities` is to bring it into compliance with the new check?

nnmm:
Precisely. Thanks for reviewing!

PR #82, opened 2022-03-16
Add CONTRIBUTING.md

nnmm:


CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+- `struct A {…}`
+- `struct B {…}`
+- `impl Bar for A {…}`
+- `impl Foo for A {…}`
+- `impl A {…}`
+- `impl Bar for B {…}`
+- `impl Foo for B {…}`
+- `impl B {…}`
+
+Try to not exceed a line length of 100 characters.
```
jhdcs, line 31: I thought the standard was 120 characters? Or is that only for C++?

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+
+Use `cargo clippy` to help with writing idiomatic code.
+
+### Unsafe code
+Annotate every `unsafe` block with a `// SAFETY:` comment explaining why its safety preconditions are met.
+
+### Formatting
+Use `cargo fmt` to format code.
+
+Put a struct's definition first, then its trait impls in alphabetical order, then its inherent impls. When there are several structs in one file, one possibility is to put all struct definitions at the top of the file, like this:
```
jhdcs, line 21:
```suggestion
Put a `struct`'s definition first, then its trait `impl`s in alphabetical order, then its inherent `impl`s. When there are several `struct`s in one file, one possibility is to put all `struct` definitions at the top of the file, like this:
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+Contributions to `ros2_rust` are very welcome!
+
+Work is coordinated in the [Matrix chat](https://matrix.to/#/+rosorg-rust:matrix.org). It's a good idea to check in with the other contributors there before starting development on a new feature.
+
+There are also occasional ROS 2 Rust WG meetings that are announced on the [ROS discourse forum](https://discourse.ros.org/).
+
+## Coding guidelines
+Aim for well documented and tested code.
+
+Use declarative sentences, not imperative sentences, for documenting functions. For instance, `Creates a by-value iterator.`, not `Create a by-value iterator`.
```
jhdcs, line 11:
```suggestion
For documenting functions, use declarative sentences, not imperative sentences. For instance, `Creates a by-value iterator.` should be used instead of `Create a by-value iterator`. 
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Coding guidelines
+Aim for well documented and tested code.
+
+Use declarative sentences, not imperative sentences, for documenting functions. For instance, `Creates a by-value iterator.`, not `Create a by-value iterator`. 
+
+Use `cargo clippy` to help with writing idiomatic code.
+
+### Unsafe code
+Annotate every `unsafe` block with a `// SAFETY:` comment explaining why its safety preconditions are met.
```
jhdcs, line 16: Should we also mention that `unsafe` blocks are to be kept as limited in size as possible? I don't know that `unsafe` programming is very common in Rust, so newer contributors might not know best practices for it.

We can also link to [The Rustonomicon](https://doc.rust-lang.org/nomicon/) for some guidance to new `unsafe` programmers, if that would be useful.

jhdcs (changes requested):
Looks very good! Just a few thoughts/suggestions on my first pass.

esteve:
@nnmm looks good. Could you add a section about licensing? Copying the relevant parts of https://github.com/ros2/rclcpp/blob/master/CONTRIBUTING.md is enough. Thanks.

esteve (changes requested):
Looks good. Could you add a section about licensing? Copying the relevant parts of https://github.com/ros2/rclcpp/blob/master/CONTRIBUTING.md is enough. Thanks.

CONTRIBUTING.md
nnmm: 100 is the default in rustfmt. For comments, it's even 80: https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#comment_width

I edited this statement to make it only apply to comments, since code is handled by `cargo fmt`.

CONTRIBUTING.md
nnmm: Done.

CONTRIBUTING.md
nnmm: Done. I didn't link to the Rustonomicon because then I'd feel like I need to link against every resource that's equally important as the Rustonomicon.

CONTRIBUTING.md
nnmm: Done.

esteve (approved):

jhdcs (approved):

PR #83, opened 2022-03-18
Enable Clippy in CI

nnmm:


esteve (approved):
If this makes Clippy happy, so be it. Thanks @nnmm !

PR #88, opened 2022-03-25
Refactor context.rs

nnmm:
* Delete Handle wrapper
* Add SAFETY comments
* Rename default() to new() and accept Strings for usability
* Fix a few issues, such as return values not being checked

### DIFF
--- a/rclrs/src/context.rs
+++ b/rclrs/src/context.rs
@@ -3,76 +3,102 @@ use crate::rcl_bindings::*;
 use crate::Node;
 use alloc::sync::Arc;
 use alloc::vec::Vec;
-use cstr_core::{c_char, CString};
+use std::string::String;
 
 #[cfg(not(feature = "std"))]
-use spin::{Mutex, MutexGuard};
+use cstr_core::{c_char, CString};
+#[cfg(not(feature = "std"))]
+use spin::Mutex;
 
 #[cfg(feature = "std")]
-use parking_lot::{Mutex, MutexGuard};
+use cty::c_char;
+#[cfg(feature = "std")]
+use parking_lot::Mutex;
+#[cfg(feature = "std")]
+use std::ffi::CString;
 
-pub(crate) struct ContextHandle(Mutex<rcl_context_t>);
-
-impl ContextHandle {
-    pub fn get_mut(&mut self) -> &mut rcl_context_t {
-        self.0.get_mut()
-    }
-
-    pub fn lock(&self) -> MutexGuard<rcl_context_t> {
-        self.0.lock()
-    }
-}
-
-impl Drop for ContextHandle {
+impl Drop for rcl_context_t {
     fn drop(&mut self) {
         unsafe {
-            rcl_shutdown(&mut *self.get_mut() as *mut _);
+            rcl_shutdown(self as *mut _);
+            rcl_context_fini(self as *mut _);
         }
     }
 }
 
 pub struct Context {
-    pub(crate) handle: Arc<ContextHandle>,
+    pub(crate) handle: Arc<Mutex<rcl_context_t>>,
 }
 
 impl Context {
-    fn init(&self, context_env_args: Vec<CString>) -> Result<(), RclReturnCode> {
-        let c_args: Vec<*const c_char> = context_env_args.iter().map(|arg| arg.as_ptr()).collect();
-        let handle = &mut *self.handle.lock();
-
-        unsafe {
-            let allocator = rcutils_get_default_allocator();
-            let mut init_options = rcl_get_zero_initialized_init_options();
-            rcl_init_options_init(&mut init_options as *mut _, allocator);
-            rcl_init(
-                c_args.len() as i32,
-                c_args.as_ptr(),
-                &init_options as *const _,
-                handle as *mut _,
-            )
-            .ok()?;
-            rcl_init_options_fini(&mut init_options as *mut _).ok()?;
-        }
-
-        Ok(())
-    }
-
-    pub fn default(args: Vec<CString>) -> Self {
+    /// Creates a new context.
+    ///
+    /// Usually, this would be called with `std::env::args()`, analogously to `rclcpp::init()`.
+    /// See also the official "Passing ROS arguments to nodes via the command-line" tutorial.
+    ///
+    /// Creating a context can fail in case the args contain invalid ROS arguments.
+    ///
+    /// # Example
+    /// ```
+    /// # use rclrs::Context;
+    /// assert!(Context::new([]).is_ok());
+    /// let invalid_remapping = ["--ros-args", "-r", ":=:*/]"].map(String::from);
+    /// assert!(Context::new(invalid_remapping).is_err());
+    /// ```
+    ///
+    /// # Panics
+    /// When there is an interior null byte in any of the args.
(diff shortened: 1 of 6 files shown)


rclrs/src/context.rs
```diff
@@ -3,73 +3,92 @@ use crate::rcl_bindings::*;
...
-    }
-}
+use cty::c_char;
+#[cfg(feature = "std")]
+use parking_lot::Mutex;
+#[cfg(feature = "std")]
+use std::ffi::CString;
 
-impl Drop for ContextHandle {
+impl Drop for rcl_context_t {
```
nnmm, line 20: This change is meant to start a discussion, I'm not dead set on it. But a priori it makes sense to me to implement Drop directly on the rcl structs and not to wrap them if they're not going to be exposed to users anyway.

rclrs/src/context.rs
```diff
@@ -3,73 +3,92 @@ use crate::rcl_bindings::*;
 use crate::Node;
 use alloc::sync::Arc;
 use alloc::vec::Vec;
-use cstr_core::{c_char, CString};
+use std::string::String;
 
 #[cfg(not(feature = "std"))]
-use spin::{Mutex, MutexGuard};
+use cstr_core::{c_char, CString};
+#[cfg(not(feature = "std"))]
```
nnmm, line 10: no_std is currently broken – do we still want to spend effort on no_std? If so, I can create a CI job for it.

rclrs_examples/src/minimal_publisher.rs
```diff
@@ -17,13 +13,11 @@ fn main() -> Result<(), Error> {
 
     let mut publish_count: u32 = 1;
 
-    while context.ok()? {
+    loop {
```
mxgrey, line 16: I was going to say something here about how this change doesn't allow a graceful wind down for the node, but I just tested the `master` branch and found that it doesn't gracefully wind down either. I wonder if `rclcpp` does something special that `rcl` itself doesn't do to allow its `Context::ok()` to know whether `SIGINT` has been triggered.

mxgrey (approved):
I left a few remarks, but I think these changes are great!

rclrs/src/context.rs
esteve: I'm leaning towards no, the idea is that we'd merge efforts with `r2r`, but I think it's not worth the work and both projects are currently going in their own direction. @jhdcs are you ok with dropping support for `no_std`? Thanks for working on it, and sorry if you feel your work was wasted.

rclrs/src/context.rs
esteve: I'm a bit decisive about it, on the one hand, I'd rather not expose the `rcl` API directly to Rust, but also it makes no sense to keep `ContextHandle` if users can't access it and we can implement traits directly. I'd go with removing `ContextHandle` and bring it back if needed.

rclrs_examples/src/minimal_publisher.rs
esteve: Yes, `rclcpp` registers a signal handler (see https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/signal_handler.hpp) to make sure the context is shutdown correctly (and also for other stuff, like logging). I'd keep the `ok` function and improve it to work correctly with `SIGINT`

mxgrey: In that case I wonder whether observing `SIGINT` should really be done with `Context::ok` or if we should consider a different way of observing `SIGINT`.

My current impression is that there are only two things that can make `rclcpp::Context::ok` false:
1. `SIGINT` is received
2. `rclcpp::shutdown(context)` is called

If that's the case then I might float the idea of having a singleton independent of `rclrs::Context` whose entire purpose is to reflect whether `SIGINT` has occurred.

But if there are other potential reasons that `Context::ok` could turn out to be false (e.g. issues from the RMW layer?) then I think bringing back `Context::ok` and having it reflect `SIGINT` status makes sense.

nnmm: Would it be a good idea to do signal handling in the `spin()` method?

esteve (changes requested):
I left my feedback about `Context::ok` as a comment and now I can't turn it into a request for changes.

rclrs_examples/src/minimal_publisher.rs
mxgrey: A more general thing to do for `spin` would be to have an argument like

```
keep_spinning: Box<dyn FnMut() -> bool>
```

and have that callback triggered on each cycle to decide whether to keep spinning. Then the user can customize the decision however they'd like.

But if that's considered too complicated for the basic `spin` function, then yes I think it should at least be listening for `SIGINT`.

esteve: The `spin` function will removed eventually to support executors. I'd put the signal handling in the context (like `rclcpp` and `rclpy` do), so that we can register callbacks and do logging from there for example.

nnmm: So would you like me to add signal handling in this PR?

If not, I can add back the `Context::ok()` function and make it always return `true`, since `rcl_context_is_valid()` is redundant (we know that it hasn't been shutdown since the value hasn't been dropped). That doesn't feel too useful though.

esteve: Don't worry about signal handling now, just add `ok` back and we can improve that later. Thanks!

rclrs/src/context.rs
jhdcs: If `no_std` is broken, I'm okay with dropping support for it.

Perhaps someday, if/when we want to see about making this work with embedded devices, we can revisit `no_std` support. But for now, I think that getting the client library working for (probably) most use cases would be the best call.

So, in a nutshell, yes. Feel free to drop `no_std` support.

nnmm: It wouldn't be too much work to restore `no_std` support, so that probably shouldn't be weighted too heavily. For me, it's more about the extra work required by contributors in the future. But in the end, I don't have an opinion on whether to keep it, since I don't have a good grasp of how in-demand this feature would be.

Anyway, good to know this issue is not a blocker for this PR.

rclrs_examples/src/minimal_publisher.rs
nnmm: Added it back.

esteve: Can you add it back as it was not just a skeleton? Thanks.

nnmm: That's what I was talking about above. The call to `rcl_context_is_valid()` is entirely redundant. Only `rcl_shutdown()` can cause that call to return `false`, but we know shutdown hasn't been called (since shutdown only happens when the context is dropped).

esteve: Still, that's not a reason for removing it, the signal handler will most likely registered outside the `ok` funcion, which will end up being like it is currently on `main`. Please add the code back.

esteve: Yes, hence why the call to `rcl_context_is_valid` still needs to be in `ok`

nnmm: Done.

esteve (approved):

PR #89, opened 2022-03-29
Remove unnecessary dependencies and make small cleanups

nnmm:


keenanjohnson (dismissed):
This all looks good to me! I like that we're removing some dependencies here.

mxgrey (dismissed):
The cleanup looks good!

I also built this branch and ran the pub+sub examples on my machine, and everything is working :+1:

rclrs/src/rcl_bindings.rs
```diff
@@ -1,13 +1,7 @@
-#[allow(dead_code)]
+#![allow(dead_code)]
```
esteve, line 1: Since you're changing this file, could you rename it to something other than `rcl_bindings.rs`? Or rename the `bindgen` file. Either way, it's sometimes a bit confusing when referring to `rcl_bindings.rs`, which is of the two.

esteve (changes requested):

rclrs/src/rcl_bindings.rs
nnmm: Sure, I renamed the bindgen one to `rcl_bindings_generated.rs`.

esteve (approved):

PR #90, opened 2022-03-29
Improve the QoS module

nnmm:
Improve the QoS module

* Add documentation
* Make depth a u32
* Put depth into enum variants
* Derive useful traits

### DIFF
--- a/rclrs/src/qos.rs
+++ b/rclrs/src/qos.rs
@@ -1,77 +1,126 @@
 use crate::rcl_bindings::*;
 
-pub enum QoSReliabilityPolicy {
-    SystemDefault = 0,
-    Reliable = 1,
-    BestEffort = 2,
-}
-
+/// The `HISTORY` DDS QoS policy.
+///
+/// A subscription internally maintains a queue of messages (called "samples" in DDS) that have not been processed yet
+/// by the application, and likewise a publisher internally maintains a queue.
+///
+/// If the history policy is `KeepAll`, this queue is unbounded, and if it is `KeepLast`, it is bounded and old values are discarded when the queue is overfull.
+///
+/// # Compatibility
+/// | Publisher | Subscription | Compatible |
+/// | -- | -- | -- |
+/// | KeepLast | KeepLast | yes |
+/// | KeepLast | KeepAll | yes |
+/// | KeepAll | KeepLast | yes |
+/// | KeepAll | KeepAll | yes |
+///
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
 pub enum QoSHistoryPolicy {
-    SystemDefault,
+    /// Use the default policy of the RMW layer.
+    ///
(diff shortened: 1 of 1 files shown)


rclrs/src/qos.rs
```diff
@@ -137,3 +177,45 @@ impl From<QoSDurabilityPolicy> for rmw_qos_durability_policy_t {
...
+
+pub const QOS_PROFILE_PARAMETER_EVENTS: QoSProfile = QoSProfile {
+    history: QoSHistoryPolicy::KeepAll,
+    reliability: QoSReliabilityPolicy::Reliable,
+    durability: QoSDurabilityPolicy::Volatile,
+    avoid_ros_namespace_conventions: false,
+};
+
+pub const QOS_PROFILE_SYSTEM_DEFAULT: QoSProfile = QoSProfile {
+    history: QoSHistoryPolicy::SystemDefault { depth: 0 },
```
mxgrey, line 217: I assume `depth: 0` means the depth is unspecified (decided by the system defaults), and not that there is zero depth in the message buffer. It might be good to leave a comment about that so users don't get alarmed or confused.

rclrs/src/qos.rs
```diff
@@ -95,10 +135,10 @@ impl From<QoSProfile> for rmw_qos_profile_t {
 impl From<QoSHistoryPolicy> for rmw_qos_history_policy_t {
     fn from(policy: QoSHistoryPolicy) -> Self {
         match policy {
-            QoSHistoryPolicy::SystemDefault => {
+            QoSHistoryPolicy::SystemDefault { .. } => {
```
mxgrey, line 138: Just for my own edification, does the destructuring that you added here change anything? Since this isn't actually placing any constraints on the pattern matching or extracting any information from them, my understanding is this changed version is equivalent to the original version, is that right?

mxgrey (approved):
The changes look great! I just left a couple remarks.

esteve (approved):

esteve:
@nnmm one improvement I think would make `QoS` more idiomatic is to use the builder pattern to create `QoS`s. In fact, `rclcpp` uses the same pattern, but builder is even more prevalent in Rust APIs.

rclrs/src/qos.rs
nnmm: Yep, the braces are just there because Rust expects them.

mxgrey:
I've stumbled across some crates in the past that help with the builder pattern: [[1]](https://crates.io/crates/derive_builder), [[2]](https://crates.io/crates/typed-builder).

rclrs/src/qos.rs
nnmm: I don't know if that's the case, so I don't want to promise things I'm not sure about myself. But this was 0 already before this change and even in rclcpp/rmw (in `rmw_qos_profile_system_default`).

nnmm:
I merged it for now, we can flesh out the builder idea in another PR.

PR #92, opened 2022-04-01
Give subscription callback the owned message

nnmm:
This enables efficiency and ergonomics improvements: for instance, think of a node that takes incoming messages and mutates them, then re-publishes them. Or alternatively, a node could move (part of) the message into a new message to be published. When the received message is owned, this can be done with fewer copies. There is no drawback to this, since the message is already not used anymore after the callback. 

Also remove unused dependency from rosidl_runtime_rs


esteve (approved):
@nnmm looks good to me. Thanks for updating the description.

PR #94, opened 2022-04-02
Document all public items

nnmm:
It would be super nice if we could have example blocks for creating and using publishers, subscriptions etc. However, that would require adding a dependency on `std_msgs` or some other message package, because it's not so easy to implement `Message` manually for an example message struct. And I'm not sure if the extra dependency would be worth it.


nnmm:
@esteve I've decided that the WaitSet is not yet ready for making public, but that I'll take care of that in a separate PR and leave it private for now. So this PR is ready from my side.

jhdcs (dismissed):

nnmm:
@esteve I'm interpreting your silence as tacit approval and will merge this tomorrow :)

nnmm:
@jhdcs Could you re-approve please?

jhdcs (approved):

rclrs/src/node/mod.rs
```diff
@@ -38,18 +38,33 @@ impl Drop for NodeHandle {
...
+/// A processing unit that can communicate with other nodes.
+///
+/// Nodes are a core concept in ROS 2. Refer to the official ["Understanding ROS 2 nodes"][1]
+/// tutorial for an introduction.
+///
+/// Ownership of the node is shared with all [`Publisher`]s and [`Subscription`]s created from it.
+/// That means that even after the node itself is dropped, it will continue to exist and be
+/// displayed by e.g. `ros2 topic` as long as its publishers and subscriptions are not dropped.
+///
+/// [1]: https://docs.ros.org/en/foxy/Tutorials/Understanding-ROS2-Nodes.html
```
esteve, line 50: I'd link the code documentation, not the tutorials

esteve (changes requested):

rclrs/src/node/mod.rs
esteve: Also, this links to `foxy`, this would need to be updated every time there's a new release

rclrs/src/node/publisher.rs
```diff
@@ -35,7 +35,16 @@ impl Drop for PublisherHandle {
     }
 }
 
-/// Main class responsible for publishing data to ROS topics
+/// Struct for sending messages of type `T`.
+///
+/// Multiple publishers can be created for the same topic, in different nodes or the same node.
+///
+/// The underlying RMW will decide on the concrete delivery mechanism (network stack, shared
```
esteve, line 42: Not entirely accurate, intraprocess may happen above the RMW layer

rclrs/src/qos.rs
```diff
@@ -178,41 +183,47 @@ impl From<QoSDurabilityPolicy> for rmw_qos_durability_policy_t {
     }
 }
 
+/// Equivalent to `rmw_qos_profile_sensor_data`.
```
esteve, line 186: It might be useful to link to `rmw_qos_profile_sensor_data` from `rmw`

esteve (changes requested):

rclrs/src/node/publisher.rs
nnmm: I thought that is rclcpp-specific?

esteve: Yes, that's what I meant, above the RMW layer. `rclcpp` implements intraprocess on top of RMW, whereas the documentation says:

> The underlying RMW will decide on the concrete delivery mechanism (network stack, shared memory, or **intraprocess**).

The RMW can't decide to use intraprocess.

rclrs/src/qos.rs
nnmm: Which link do you have in mind? GitHub master, GitHub pinned revision, https://docs.ros2.org/galactic/api/rmw/qos__profiles_8h_source.html?

rclrs/src/node/publisher.rs
nnmm: Maybe I'm wrong, but I'd assume that the RMW or DDS layer can detect when writer and reader are in the same process and use an optimized intraprocess delivery mechanism (independently from rclcpp having its own intraprocess mechanism). This is also how I'd read e.g. [this post](https://discourse.ros.org/t/ros2-foxy-rmw-fast-dds-improved-intra-process-inter-process-performance/17422).

rclrs/src/node/mod.rs
nnmm: Which specific code documentation? I like the tutorial because it explains the concept in a language-agnostic manner. We can link to the Rolling version to avoid the staleness problem: https://docs.ros.org/en/rolling/Tutorials/Understanding-ROS2-Nodes.html.

nnmm: I changed the link to the Rolling version.

nnmm:
I added `#![warn(missing_docs)]`, and fixed all places where they were still missing.

rclrs/src/node/publisher.rs
esteve: I thought you meant [`rclcpp`'s intra-process](https://docs.ros.org/en/galactic/Tutorials/Intra-Process-Communication.html), in that case, you're right about the docstring.

rclrs/src/qos.rs
esteve: I'd use master, what do you think?

nnmm: Ok. I added a link to the `qos_profiles.h` file to the docstring, but didn't hyperlink `rmw_qos_profile_sensor_data` directly, because the specific line where it's defined in might change.

jhdcs (approved):

nnmm:
@esteve Could you do another (maybe final) round of review?

esteve (approved):
@nnmm great job!

PR #96, opened 2022-04-03
Fix spin taking messages from non-ready subscriptions

nnmm:
Only call `rcl_take` on ready subscription. Only trigger callbacks if subscriptions have consumed any data

Fixes #96


esteve (approved):
@nnmm thanks for fixing this. I've updated the description in this PR, please can you add a description and a link to their issue in future subsequent PRs? Thanks.

nnmm:
Oh dang, I accidentally committed my changes to the minimal pubsub that I was using to test this. I'll open a PR to revert.

PR #98, opened 2022-04-04
Remove unnecessary WaitSetErrorResponse

nnmm:
By doing the upgrade outside of the wait set, this struct can be entirely omitted.

### DIFF
--- a/rclrs/src/lib.rs
+++ b/rclrs/src/lib.rs
@@ -27,15 +27,15 @@ pub use self::node::*;
 pub use self::qos::*;
 
 use self::rcl_bindings::*;
-use wait::{WaitSet, WaitSetErrorResponse};
+
+use wait::WaitSet;
 
 /// Wrapper around [`spin_once`]
-pub fn spin(node: &node::Node) -> Result<(), WaitSetErrorResponse> {
+pub fn spin(node: &node::Node) -> Result<(), RclReturnCode> {
     while unsafe { rcl_context_is_valid(&mut *node.context.lock() as *mut _) } {
         if let Some(error) = spin_once(node, 500).err() {
             match error {
-                WaitSetErrorResponse::DroppedSubscription
-                | WaitSetErrorResponse::ReturnCode(RclReturnCode::Timeout) => continue,
+                RclReturnCode::Timeout => continue,
                 error => return Err(error),
             };
         }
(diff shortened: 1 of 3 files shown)


rclrs/src/node/mod.rs
```diff
@@ -110,4 +110,14 @@ impl Node {
             .push(Arc::downgrade(&subscription) as Weak<dyn SubscriptionBase>);
         Ok(subscription)
     }
+
+    /// Returns the subscriptions that have not been dropped yet.
+    pub(crate) fn live_subscriptions(&self) -> Vec<Arc<dyn SubscriptionBase>> {
+        let live_subs = self
```
esteve, line 116: This can be made more succinct by dropping the `live_subs` variable

rclrs/src/lib.rs
```diff
@@ -81,45 +81,37 @@ pub fn spin(node: &node::Node) -> Result<(), WaitSetErrorResponse> {
...
-            Err(err) => return Err(err),
-        };
+    let live_subscriptions = node.live_subscriptions();
+    for sub in &live_subscriptions {
+        wait_set.add_subscription(&**sub)?;
     }
 
     wait_set.wait(timeout_ns)?;
-    for (i, sub) in node.subscriptions.iter().enumerate() {
+    for (i, sub) in live_subscriptions.iter().enumerate() {
```
esteve, line 108: Please avoid shortened names, rename `sub` to `live_subscription`

esteve (changes requested):

rclrs/src/lib.rs
```diff
@@ -81,45 +81,37 @@ pub fn spin(node: &node::Node) -> Result<(), WaitSetErrorResponse> {
...
+        &mut *node.context.lock(),
     )?;
 
-    for subscription in &node.subscriptions {
-        match wait_set.add_subscription(subscription) {
-            Ok(()) => (),
-            Err(WaitSetErrorResponse::DroppedSubscription) => (),
-            Err(err) => return Err(err),
-        };
+    let live_subscriptions = node.live_subscriptions();
```
esteve, line 102: Should this be guarded by a lock in case the stored subscriptions change?

nnmm: No, it's a vector of `Arc`s, so as long as this variable is in scope they won't change.

rclrs/src/lib.rs
nnmm: Done (though I personally don't mind the shorter names as it is visually more distinct from the `live_subscriptions` variable).

rclrs/src/node/mod.rs
nnmm: Yep, I at first considered also pruning dead subscriptions in here, but then decided against it. Done.

esteve (approved):

PR #99, opened 2022-04-10
Refactor wait set

nnmm:
- Properly ensure lifetime of context in waitset by storing a refcounted `rcl_context_t` pointer
  - This makes it natural to take a `Context` in `WaitSet::new` instead of `rcl_context_t`
- Make `WaitSet::wait` return a list of ready entities
  - This requires storing the subscriptions in the wait set
- Change timeout to a proper `Duration` type
- Remove unnecessary Result from `WaitSet::clear()`
- Add SAFETY comments to all unsafe functions
- Improve documentation
- Make `wait` module public


nnmm:
This is a complicated topic and I'm not 100% sure about everything.

For instance, is it okay if the node being spun is created from a different context than the wait set?

rclrs/src/lib.rs
```diff
@@ -25,95 +25,4 @@ pub use self::context::*;
...
-///                   |
-///         +---------v----------+
-///         |     rcl_wait       |
-///         +---------+----------+
-///                   |
-///         +---------v----------+
-///         | rcl_wait_set_fini  |
-///         +--------------------+
-///
-///
```
jhdcs, line 83: I rather liked the ASCII art representation of how the underlying RCL functions were supposed to work/be called. Would it be worth adding this back in to where `spin_once` now lives?

rclrs/src/wait.rs
```diff
@@ -35,16 +52,11 @@ impl Drop for rcl_wait_set_t {
...
-        number_of_clients: usize,
-        number_of_services: usize,
-        number_of_events: usize,
-        context: &mut rcl_context_t,
-    ) -> Result<Self, RclReturnCode> {
+    /// Creates a new wait set.
+    ///
+    /// The given number of subscriptions is a capacity, corresponding to how often
+    /// [`WaitSet::add_subscription`] may be called.
+    pub fn new(number_of_subscriptions: usize, context: &Context) -> Result<Self, RclReturnCode> {
```
jhdcs, line 59: Is the plan here to add back in the other parameters as we implement them?

rclrs/src/wait.rs
```diff
@@ -53,100 +65,115 @@ impl WaitSet {
...
-    /// # Errors
-    /// - `WaitSetError::DroppedSubscription` if the passed weak pointer refers to a dropped subscription
-    /// - `WaitSetError::RclError` for any `rcl` errors that occur during the process
+    /// It is possible, but not useful, to add the same subscription twice.
+    ///
+    /// This will return an error if the number of subscriptions in the wait set is larger than the
+    /// capacity set in [`WaitSet::new`].
+    ///
+    /// The same subscription must not be added to multiple wait sets, because that would make it
+    /// unsafe to simultaneously wait on those wait sets.
```
jhdcs, line 108: Perhaps someday we can brainstorm some way to make sure that something is not added to multiple wait sets - but I'm not sure it's possible without significant performance overhead, and is outside the scope of this PR anyways.

jhdcs (changes requested):
Just a couple documentation changes that I saw. One in `context.rs`, line 141, the other is mentioned in `lib.rs` on the old line 83.

Still, looks pretty good to me! Hopefully I haven't missed anything!

rclrs/src/context.rs
```diff
@@ -107,4 +107,46 @@ impl Context {
...
+    /// Polls the node for new messages and executes the corresponding callbacks.
+    ///
+    /// See [`WaitSet::wait`] for the meaning of the `timeout` parameter.
+    ///
+    /// This may under some circumstances return
+    /// [`SubscriptionTakeFailed`][1] when the wait set spuriously wakes up.
+    /// This can usually be ignored.
+    ///
+    /// [1]: crate::SubscriberErrorCode
+    pub fn spin_once(&self, node: &Node, timeout: Option<Duration>) -> Result<(), RclReturnCode> {
```
esteve, line 120: I'll elaborate more later, but the `spin` functions can't be in the context. The context is only used for sharing data between nodes, but once the executors are implemented, spin will be moved there.

rclrs/src/wait.rs
nnmm: Yes. Or maybe do some kind of builder pattern.

rclrs/src/lib.rs
nnmm: To be honest, I don't find it that helpful:
- All of the functions have different names in `rclrs` than in `rcl`
- `rcl_get_zero_initialized_wait_set` and `rcl_wait_set_fini` are implementation details and should not concern users
- `rcl_wait_set_clear` doesn't need to be called after `init`. If anything, it should be an arrow back from a filled wait set to an empty one.

So basically, it just says "add subscriptions before you call `wait()`", and that is mentioned in a comment already. 

If we want something like this for contributors to learn about the typical `get zero-initialized value -> init -> do something -> fini` sequence, that's a fair point, but I think I'd add it to `CONTRIBUTING.md` since it's not specific to the wait set.

rclrs/src/context.rs
```diff
@@ -107,4 +107,46 @@ impl Context {
...
+        }
+
+        Ok(())
+    }
+
+    /// Convenience function for calling [`Context::spin_once`] in a loop.
+    ///
+    /// This function additionally checks that the context is still valid.
+    pub fn spin(&self, node: &Node) -> Result<(), RclReturnCode> {
+        while unsafe { rcl_context_is_valid(&mut *node.context.lock() as *mut _) } {
```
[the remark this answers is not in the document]
nnmm: Done, thanks.

rclrs/src/context.rs
nnmm: Ok, I moved them into a new `executor` module (`pub use`d from lib.rs).

rclrs/src/wait.rs
nnmm: Yep.

rclrs/src/context.rs
esteve: Not sure about that, there's no executor infrastructure yet, so having an `executor` module is a bit misleading. I'd rather those functions are moved back to `lib.rs`, like other bindings do (e.g. [rclpy](https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/__init__.py#L184=))

esteve:
@nnmm looks good to me, my only feedback is moving the `spin*` functions back to `lib.rs`

rclrs/src/context.rs
nnmm: Moved back to `lib.rs`.

esteve:
@nnmm thanks!

jhdcs (approved):

PR #100, opened 2022-04-10
Bump every package to version 0.2

nnmm:
The reason for the "bump" is to have a consistent version numbering across packages, not because it's needed for a release.

### DIFF
--- a/rclrs/Cargo.toml
+++ b/rclrs/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rclrs"
-version = "0.1.0"
+version = "0.2.0"
 authors = ["Esteve Fernandez <esteve@apache.org>"]
 edition = "2021"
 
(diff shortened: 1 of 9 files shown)


jhdcs:
This appears good to me! The version "regression" in `rclrs_examples` is just a correction, right?

jhdcs (approved):

esteve (changes requested):
Leaving a comment here before this gets merged and I'll add more later, but this PR needs to be split into multiple ones and not ok with the author field. The tradition in ROS is to keep the author field intact, but add maintainers. For example, rclcpp still has Dirk as the sole author, despite William being the most active committee and me being the third. Nevertheless, Jacob should be added as a maintainer as well.

nnmm:
Ok, sure @esteve. I wasn't clear about the meaning of the author field. I'm making this PR just about the version numbers then.

esteve:
@nnmm thanks!

nnmm:
@esteve Done.

nnmm:
@esteve Could you approve if no further changes are required?

esteve (approved):

Issue #33, opened 2020-11-07
Missing benchmark tests


ruffsl:
A topic that emerged from the previous Rust WG meeting is the fact that rclrs currently lacks benchmarks for testing performance or efficiency of the client library. Adding such benchmarks could go a long way in vetting and quantifying PR improvements, such as when validating changes to add array IDL support for more advanced message types. This may also be a good first issue.

https://discourse.ros.org/t/ros-rust-wg-meeting-2020-11-05/17101

PR #101, opened 2022-04-11
Add benchmarks for Subscriber and Publisher

AlexKaravaev:
Hi everyone!

It's a PR for #33, however, I don't really know which of the parts of the library should be benchmarked at first, currently I only have written tests for simple publishing and subscribing for simple string and creating of publisher and subscriber instances. 

Ideas that I have about what can be benchmarked:

- Publishing and subscribing with more heavy payload(E.g progressing increase of message size)
- Message creation

Please feel free to add some ideas

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -83,6 +83,6 @@ jobs:
         for path in $(colcon list | awk '$3 == "(ament_cargo)" { print $2 }'); do
         cd $path
         echo "Checking $path"
-        cargo clippy -- -D warnings
+        cargo clippy --all-targets --all-features -- -D warnings
         cd -
         done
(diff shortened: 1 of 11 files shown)


jhdcs:
Would benchmarking the creation and sending/receiving of messages with allocated portions (such as Strings) vs trivially-copyable portions (such as i32) be of use?

Another thing that _might_ be good to benchmark are all the various ROS 2 object creation functions: creating nodes, creating contexts, creating publishers, creating subscribers, etc. More for catching performance regressions than anything else. While these functions mostly involve external ROS 2 code, it would give us insight into the health of the external system.

nnmm:
@AlexKaravaev Want to revive this to gather some numbers for sending messages in the same process vs between processes?

We could construct a message, and then time the duration for
- sending the message
- receiving the message on the other end, and
- sending back a very small response (e.g. `std_msgs/msg/Empty`)

for different message types. See also https://gitlab.com/ApexAI/performance_test

esteve:
Closed due to inactivity. @AlexKaravaev let us know if you'd like to revisit this. Thanks.

PR #102, opened 2022-04-12
Remove NodeHandle struct

nnmm:
This is a small simplification that makes it (imo) easier to see what's going on, besides making the code a little smaller.


esteve (approved):
I like this simplification. Should we go further and remove all the `*Handle` structs?

PR #103, opened 2022-04-14
Add SAFETY comments to all unsafe blocks in rclrs

nnmm:


jhdcs (changes requested):
Looks pretty good to me! I only noticed one spot that was missed! Unfortunately GitHub won't let me link it, because it's not close enough to a changed line (I think that's the reason, anyway), but it's in `mod.rs`, line 37, in the Drop implementation for NodeHandle.

nnmm:
@jhdcs Right, I left that out because that part is changed in https://github.com/ros2-rust/ros2_rust/pull/102/files, but it makes sense to add it here already and then transfer that comment over. Done.

esteve (dismissed):
@nnmm awesome work! Thanks

esteve (dismissed):

nnmm:
@jhdcs Could you approve this PR?

jhdcs (dismissed):

nnmm:
Sorry, I'd need another approval because I just merged #102 and rebased.

esteve (approved):

nnmm:
Thanks!

PR #105, opened 2022-04-15
Fix timeout_ns set value when None is passed to timeout param

### LINKED ISSUE #104: WaitSet::wait does not wait indefinitely when timeout is None
Hi.
I am new to ros2 with rust and tried to run minimal_publisher and minimal_subscriber.
However minimal_subscriber does not work in expect ( no output is displayed ).
After investigation, I found that there are some problems in `WaitSet::wait()`.

**Current Behavior**
Even if passing `None` to `WaitSet::wait`, it does not wait until some messages are passed and return immediately (i.e. non-blocking).

**Expected Behavior** 
Waiting (blocking) indefinitely until some messages are passed.

As described [here](https://docs.ros2.org/beta1/api/rcl/wait_8h.html#a732278988c802fe2c9d8ec24752f9dd9), `timeout_ns` should be minus if procedure wants to wait indefinitely, but current implementation make rcl_wait non-blocking function by passing zero.

Soya-Onishi:
Fix #104 

`rcl_wait` requires negative value if procedure wants to
block thread indefinitely until some messages are passed.
However, in current implementation, `timeout_ns` is set to zero,
even if `timeout` param is `None`, and this results in
making `rcl_wait` non-blocking function ( returns immediately ).

To make it correctly, I fixed to set `timeout_ns` to `-1` if `timeout` param is `None`.

### DIFF
--- a/rclrs/src/wait.rs
+++ b/rclrs/src/wait.rs
@@ -150,7 +150,7 @@ impl WaitSet {
     /// [1]: std::time::Duration::ZERO
     pub fn wait(&mut self, timeout: Option<Duration>) -> Result<ReadyEntities, RclReturnCode> {
         let timeout_ns = match timeout.map(|d| d.as_nanos()) {
-            None => 0,
+            None => -1,
             Some(ns) if ns <= i64::MAX as u128 => ns as i64,
             _ => {
                 return Err(RclReturnCode::InvalidArgument);


nnmm:
Very good catch, thanks!

nnmm (approved):

PR #107, opened 2022-04-15
Fix to collect ready subscription entities

### LINKED ISSUE #106: WaitSet::wait does not return ready subscription entities
Related to #104.
`WaitSet::wait()` does not collect entities that have messages, and this causes `minimal_subscriber` does not display any messages from `minimal_publisher` .

**Expected Behavior**
correct and return entities that have messages from publisher.

Related code is [here](https://github.com/ros2/rcl/blob/35a31b00a12f259d492bf53c0701003bd7f1745c/rcl/include/rcl/wait.h#L424)
Above code check *this is **not** null*. i.e. `wait_set_entry.is_null()` in `WaitSet::wait()` should attach not operator.

Soya-Onishi:
Fix #106

`minimal_subscriber` in `rclrs_examples` does not display anything.
This is because `ready_entities` in `WaitSet::wait()` collect
null `wait_set_entry` (i.e. not ready entries).

In actual, `ready_entries` should return array of "not" null `wait_set_entry`.
This fix makes so.

### DIFF
--- a/rclrs/src/wait.rs
+++ b/rclrs/src/wait.rs
@@ -170,7 +170,7 @@ impl WaitSet {
             // equivalent to
             // https://github.com/ros2/rcl/blob/35a31b00a12f259d492bf53c0701003bd7f1745c/rcl/include/rcl/wait.h#L419
             let wait_set_entry = unsafe { *self.handle.subscriptions.add(i) };
-            if wait_set_entry.is_null() {
+            if !wait_set_entry.is_null() {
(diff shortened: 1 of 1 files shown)


nnmm (approved):
Thank you very much for this fix! It's an oversight that slipped in in https://github.com/ros2-rust/ros2_rust/pull/99, and apparently I did not test sufficiently. We definitely need to add automatic tests for that.

PR #109, opened 2022-04-19
Removed support for no_std

esteve:
This PR removes support for `no_std`, by removing the dependency for the `spin` crate and the corresponding `cfg()` blocks.

I recall @nnmm saying that `no_std` is broken (https://github.com/ros2-rust/ros2_rust/pull/88#discussion_r835641156) and I thought we might as well just remove support for it and revisit this in the future if we want to.

### DIFF
--- a/rclrs/Cargo.toml
+++ b/rclrs/Cargo.toml
@@ -13,8 +13,7 @@ lock_api = "0.4.5"
 cstr_core = "0.2"
 cty = "0.2"
 core-error = "0.0.0"
-parking_lot = {version = "0.11.2", optional = true}
-spin = "0.9.2"
+parking_lot = "0.11.2"
 downcast = "0.10.0"
 
 [dependencies.rosidl_runtime_rs]
@@ -22,7 +21,3 @@ version = "*"
 
 [build-dependencies]
 bindgen = "0.59.1"
-
-[features]
-default = ["std"]
-std = ["parking_lot"]
--- a/rclrs/Cargo.toml
+++ b/rclrs/Cargo.toml
@@ -10,7 +10,6 @@ path = "src/lib.rs"
 [dependencies]
 libc = "0.2.43"
 parking_lot = "0.11.2"
-downcast = "0.10.0"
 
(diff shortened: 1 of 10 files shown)


esteve:
Kudos to @jhdcs for implementing support for `no_std` in the first place.

jhdcs (dismissed):
Too bad things aren't working, but if we ever decide to re-add it, I hope that what I did could serve as a template.

nnmm:
I think so, if we want to add it back we know how to do that.

rclrs/Cargo.toml
```diff
@@ -13,16 +13,11 @@ lock_api = "0.4.5"
 cstr_core = "0.2"
 cty = "0.2"
```
nnmm, line 14: Aren't `cstr_core` and `cty` also no_std-only dependencies?

rclrs/src/error.rs
```diff
@@ -613,7 +611,7 @@ impl ToResult for rcl_ret_t {
 
 #[cfg(test)]
 mod tests {
-    use core::convert::TryFrom;
+    use std::convert::TryFrom;
```
nnmm, line 614: Should not be needed iirc, since we're on the 2021 Rust edition

nnmm (approved):

rclrs/src/error.rs
```diff
@@ -1,9 +1,7 @@
 use crate::rcl_bindings::*;
-use core::{
-    convert::TryFrom,
-    fmt::{self, Display},
-};
-use core_error::{self, Error};
+use std::convert::TryFrom;
```
jhdcs, line 2: If @nnmm is correct about us not needing this for Rust 2021, then this line can be removed as well.

jhdcs (approved):

rclrs/Cargo.toml
esteve: Removed

nnmm (approved):
Thanks for making the code cleaner :)

nnmm (approved):

jhdcs (approved):

PR #117, opened 2022-04-21
Added rustdoc check step to CI

### LINKED ISSUE #113: Add rustdoc check step to CI
We can check with `cargo rustdoc -- -D warnings` whether a package's documentation contains any warnings or errors. This could be another step in the CI pipeline.

mohitsaxenaknoldus:
Closes #113

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -90,6 +90,7 @@ jobs:
     - name: Rustdoc check
       run: |
         cd ${{ steps.build.outputs.ros-workspace-directory-name }}
+        . /opt/ros/${{ matrix.ros_distro }}/setup.sh
         for path in $(colcon list | awk '$3 == "(ament_cargo)" { print $2 }'); do
         cd $path
         cargo rustdoc -- -D warnings
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -89,8 +89,8 @@ jobs:
 
     - name: Rustdoc check
       run: |
-        cd rclrs && cargo rustdoc -- -D warnings
-        cd .. && cd rclrs_example_msgs && cargo rustdoc -- -D warnings
-        cd .. && cd rosidl_generator_rs && cargo rustdoc -- -D warnings
-        cd .. && cd rclrs_examples && cargo rustdoc -- -D warnings
-        cd .. && cd rosidl_runtime_rs && cargo rustdoc -- -D warnings
+        for path in $(colcon list | awk '{ print $2 }'); do
+        cd $path
+        cargo rustdoc -- -D warnings
+        cd -
+        done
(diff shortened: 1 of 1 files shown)


nnmm:
I think this won't work yet – you need to execute this in each package's directory, not the top-level directory.

mohitsaxenaknoldus:
> I think this won't work yet – you need to execute this in each package's directory, not the top-level directory.

Oh! My bad...check now?

nnmm:
Could you do it the same way as the clippy step? That should work.

mohitsaxenaknoldus:
> Could you do it the same way as the clippy step? That should work.

Done.

nnmm:
You removed the filter for `ament_cargo` packages, so now it tries to run rustdoc on Python and CMake packages too.

mohitsaxenaknoldus:
> You removed the filter for `ament_cargo` packages, so now it tries to run rustdoc on Python and CMake packages too.

Oh! I thought we wanted this for every package, my bad! Fixed.

nnmm:
Now we're getting somewhere – though now it can't find the dependencies, because the `.cargo` folder is in a different folder.

If you add back the `cd ${{ steps.build.outputs.ros-workspace-directory-name }}` from the clippy step, it should probably work.

mohitsaxenaknoldus:
> Now we're getting somewhere – though now it can't find the dependencies, because the `.cargo` folder is somewhere else.
> 
> If you add back the `cd ${{ steps.build.outputs.ros-workspace-directory-name }}` from the clippy step, it should probably work. That's where the `colcon build` command ran, so it contains the `.cargo` folder.

Added

nnmm:
Looks like we also need the last command from clippy, sourcing the ROS 2 installation. I wasn't sure about that.

mohitsaxenaknoldus:
> Looks like we also need the last command from clippy, sourcing the ROS 2 installation. I wasn't sure about that.

This one?
`. /opt/ros/${{ matrix.ros_distro }}/setup.sh`

nnmm:
Yep

nnmm:
Looks like `rclrs_examples` is failing the rustdoc step, because it contains multiple binaries.

I'd suggest to just filter out `rclrs_examples` for now.

mohitsaxenaknoldus:
> Looks like `rclrs_examples` is failing the rustdoc step, because it contains multiple binaries.
> 
> I'd suggest to just filter out `rclrs_examples` for now.

Done.

nnmm:
I can't see where you filtered it out?

mohitsaxenaknoldus:
> I can't see where you filtered it out?

`for path in $(colcon list | awk '$3 == "(ament_cargo)" { print $2 }' | sed -n 'p;n'); do`

mohitsaxenaknoldus:
> Ah, thanks. That works, but it's not obvious what it's intended to do. Could you do this instead? `awk '$3 == "(ament_cargo)" && $1 != "rclrs_examples" { print $2 }'`

Done

nnmm (approved):

nnmm:
Thanks!

PR #119, opened 2022-04-22
Remove unnecessary "extern crate" declarations

nnmm:
This was only needed in Rust 2015.


jhdcs (approved):

jhdcs:
Nice and simple. I'll merge this on in.

PR #122, opened 2022-04-22
Add comments explaining each dependency

nnmm:
I got the idea from this reddit post: https://www.reddit.com/r/rust/comments/u5cjmu/dependency_justification_section_may_need_more/

### DIFF
--- a/rclrs/Cargo.toml
+++ b/rclrs/Cargo.toml
@@ -7,12 +7,16 @@ edition = "2021"
 [lib]
 path = "src/lib.rs"
 
+# Please keep the list of dependencies alphabetically sorted,
+# and also state why each dependency is needed.
 [dependencies]
+# Needed for FFI
 libc = "0.2.43"
+# Provides better concurrency primitives than std
 parking_lot = "0.11.2"
-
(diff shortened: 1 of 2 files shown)


jhdcs (changes requested):
Looks good, but should there be comments for the examples crate as well? To fully document everything going on for ROS 2 users not used to Rust (and vice-versa)? Or would this be repeating ourselves too much?

nnmm:
> Looks good, but should there be comments for the examples crate as well? To fully document everything going on for ROS 2 users not used to Rust (and vice-versa)? Or would this be repeating ourselves too much?

I looked at the `rclrs_examples/Cargo.toml` but decided not to comment it for now. It's just anyhow + message packages, so there's nothing interesting to say about these, and I also think this convention applies more to libraries than binaries. If you'd like me to, I can add it though.

jhdcs (approved):
If you feel that adding the extra comments will be too confusing/verbose, then I will defer to your judgement. Otherwise, everything looked good to me!

nnmm:
Thanks!

PR #125, opened 2022-04-23
Fix executable name in README

GeauxEric:


esteve (approved):
@GeauxEric nice catch, thanks!

PR #128, opened 2022-04-26
Install cargo-ament-build with --debug

nnmm:
This is faster than the release build.

Also print the current path in check steps.


nnmm:
Well, if anything, using the debug build makes it easier to spot errors (debug_assert etc.).

jhdcs (approved):

PR #131, opened 2022-04-28
Add serde support to messages

### LINKED ISSUE #115: Optionally make messages implement Serde traits
Messages could have a serde feature (in the Cargo.toml sense of "feature") which makes them derive `Deserialize` and `Serialize`.

nnmm:
The current implementation is not the most efficient one possible, except for sequence serialization: It uses `std::string::String` and `Vec` as intermediate formats instead of directly deserializing/serializing to the `rosidl_runtime_rs` datatype.

Closes #115


jhdcs (approved):
This seems to look all right to me, though I'm hoping @esteve will be able to give a look-through and approval as well before merging. I'm not 100% confident on my CMake skills, for example.

I do like the `quickcheck` crate - I'll have to look into what all it can do...

esteve (approved):

Issue #116, opened 2022-04-21
Use version number from package.xml in the generation of messages


nnmm:
The `Cargo.toml` of a generated message crate should probably contain a version number that matches the version number in its `package.xml`. Currently it is a hard-coded number.

Related to https://github.com/ros2-rust/ros2_rust/issues/114

PR #138, opened 2022-04-29
Generate Cargo.toml of message crate with an EmPy template, not CMake

### LINKED ISSUE #114: Generate Cargo.toml of message crate with an EmPy template, not CMake
Currently, the source files are generated with an EmPy template, but Cargo.toml is created from a CMake template. It would be good to create Cargo.toml from an EmPy template as well, for simplicity.

nnmm:
Closes #114


rosidl_generator_rs/resource/Cargo.toml.em
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "@(package_name)"
+version = "0.2.0"
```
esteve, line 3: Can you add an EmPy comment with a reference to https://github.com/ros2-rust/ros2_rust/issues/116 ?

esteve (changes requested):

rosidl_generator_rs/resource/Cargo.toml.em
nnmm: Good idea, done.

esteve: Thanks.

esteve (approved):
Really nice, the less CMake magic we use, the better.

PR #140, opened 2022-04-30
Add build.rs to messages to automatically find the message libraries

nnmm:
I think the relative path will be always correct, but even if not, this is only an _additional_ search dir.

The impact is that you can build packages depending on message packages much more easily with `cargo`.

### DIFF
--- a/rosidl_generator_rs/resource/build.rs.em
+++ b/rosidl_generator_rs/resource/build.rs.em
@@ -1,0 +1,10 @@
+use std::path::Path;
+
+fn main() {
+	let lib_dir = Path::new("../../../lib")
+		.canonicalize()
+		.expect("Could not find '../../../lib'");
+	// This allows building Rust packages that depend on message crates without
+	// sourcing the install directory first.
+	println!("cargo:rustc-link-search={}", lib_dir.display());
+}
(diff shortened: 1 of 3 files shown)


esteve:
@nnmm can you elaborate a bit more what issue this PR fixes? Thanks.

nnmm:
@esteve Even if you have a `.cargo/config.toml` so that dependencies can be found, you cannot currently build packages depending on messages with `cargo`. You also need to source the install beforehand, because otherwise you'll get a linker error about not finding `librclrs_example_messages_typesupport.so` or somesuch.

This build script fixes that, so that building with pure cargo becomes easier.

esteve:
@nnmm thanks for the explanation, makes sense.

rosidl_generator_rs/resource/build.rs.em
```diff
@@ -0,0 +1,10 @@
+use std::path::Path;
+
+fn main() {
+	let lib_dir = Path::new("../../../lib")
```
esteve, line 4: Is this expecting a specific directory structure? What would happen with a merge install? i.e. everything installed into the install directory, no separate packages.

nnmm: Yep, the relative path is still the same. I tested it.

Issue #141, opened 2022-04-30
Call rcl functions through a macro


nnmm:
Calling a fallible rcl function involves several steps:
- Calling the function
- Converting the error to our `RclReturnCode` type (see also #111)
- Calling the `rcutils_reset_error()` function if needed (currently we never do this)

Additionally, all unsafe calls should have a `SAFETY` comment.

I suggest we make a macro `ffi_call!` that does these steps above and takes an additional string literal argument with the safety justification. That way none of these steps, nor the safety justification, can be forgotten.

Potentially, several different macros would be needed, for different return types.

nnmm:
I don't think this is needed very much anymore, after I realized I can do the `rcutils_reset_error()` call inside the conversion function to `RclReturnCode`.

PR #143, opened 2022-05-01
Add new node name-related functions

nnmm:
- Add Node::name, Node::namespace, Node::fully_qualified_name getters
- Add Context::create_node_with_namespace function
- Swap order of namespace and name argument in Node::new_with_namespace, so that it's less confusing

### DIFF
--- a/rclrs/src/node/mod.rs
+++ b/rclrs/src/node/mod.rs
@@ -8,10 +8,11 @@ mod subscription;
 pub use self::publisher::*;
 pub use self::subscription::*;
 
-use std::ffi::CString;
+use std::ffi::{CStr, CString};
 use std::sync::{Arc, Weak};
 use std::vec::Vec;
 
+use libc::c_char;
 use parking_lot::Mutex;
 
 use rosidl_runtime_rs::Message;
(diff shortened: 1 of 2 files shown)


rclrs/src/node/mod.rs
```diff
@@ -88,6 +89,87 @@ impl Node {
...
+    /// let context = Context::new([])?;
+    /// let node = context.create_node_with_namespace("/my/namespace", "my_node")?;
+    /// assert_eq!(node.fully_qualified_name(), String::from("/my/namespace/my_node"));
+    /// # Ok::<(), RclReturnCode>(())
+    /// ```
+    pub fn fully_qualified_name(&self) -> String {
+        self.get_string(rcl_node_get_fully_qualified_name)
+    }
+
+    fn get_string(
```
esteve, line 155: This looks like a good candidate for a macro, IMHO

nnmm: I don't care much either way in this instance, but the approach using a function could be said to have a few advantages:
- Less code size (obviously doesn't matter much for such a small function)
- No need to know `macro_rules!`
- Better [IDE support](https://rust-analyzer.github.io/blog/2021/11/21/ides-and-macros.html)
- You can see the type of the function supported by the helper
- Definition of the helper is closer to the call site (since macro would typically be at the top level)

nnmm: It's a little less efficient though, but I think for these functions that won't matter.

esteve: Yeah, that makes sense. I was just thinking of https://github.com/ros2-rust/ros2_rust/issues/141 and how to integrate it with this, though.

esteve (approved):

rclrs/src/node/mod.rs
nnmm: Yes, that's something we'll need to think about. Maybe `call_ffi!(getter, …)` will already work if the right token type is used.

Issue #126, opened 2022-04-26
Port executor API from ROS2


esteve:
We currently use simple `spin` functions to interact with waitsets and any waitable entities (subscriptions, clients, services, etc.), porting the [executor API](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/executor.hpp) would help us implement more waitable entities (e.g. timers) and support running more than one node in a single program.

jhdcs:
Would we be able to use some crates to help implement this solution (such as `tokio` or `rayon`), or do we want to make this a sort of 1:1 port of what `rclcpp` does?

esteve:
@jhdcs for now I'd just port the `rclcpp` API 1:1 as a first step, but then offer the possibility to integrate with `async_std`, `tokio`, etc. so that existing Rust applications can integrate with ROS2. Not sure how that'd look like, for example via an implementation of the ROS2 executor API that can call both the RMW constructs (e.g. waitsets) and the Rust executor crates. The `Future` implementation is highly dependent on the executor, so it'll require some work.

nnmm:
This is an issue where I feel there is a lot to learn for us about what a good design would be.

For instance, if we tried to port the `rclcpp` API 1:1, we'd run into [`CallbackGroup`](https://docs.ros.org/en/humble/How-To-Guides/Using-callback-groups.html) being used a lot in the executor. Does the concept of a callback group make sense in Rust?  There is no need to use a callback group to "guard access to shared critical or non-thread-safe resources" since the language forces you to be thread-safe.

Async is tricky too. One observation is that it would probably be possible to make all things that can be waited on, like nodes, subscriptions, timers etc., instances of `std::task::Future`, and then we would not need to write our own executor, we could simply use the `tokio` or `smol`executor in the single-threaded or multi-threaded configuration. This is not a suggestion yet to actually do this.

> The Future implementation is highly dependent on the executor, so it'll require some work.
I thought a `Future` is generally independent of the async executor?

nnarain:
I believe the wake implementation ends up being executor specific.

PR #146, opened 2022-05-02
Added support for clients and services

esteve:
This PR adds support for clients and services. Includes a very primitive implementation of a `Future` which will surely evolved once we support the executor API (see https://github.com/ros2-rust/ros2_rust/issues/126).


i-am-chauhan:
Hi @esteve , My name Shubham Chauhan. My team is using ros2 in our project. We are considering to use rust client for writing ros2 nodes. Currently, we are using CPP client. So I was looking into this repo. Although we found out it doesn't support services and clients yet. But as I can see, that will be supported once this PR has been merged. 

So when can we expect this to be merged in the master and be officially supported. And if we can help you with anything, please let us know.

nnmm:
@esteve Could you rebase this onto master? Also, what design decisions did you make along the way?

nnmm:
> @esteve Could you rebase this onto master? Also, what design decisions did you make along the way?

Actually, nevermind with rebasing the whole thing - I think it would be better to open separate PRs for message generation and service (without client) first.

nnmm:
> @esteve Could you rebase this onto master? Also, what design decisions did you make along the way?

Actually, nevermind with rebasing the whole thing - I think it would be better to open separate PRs for message generation and service (without client) first.

esteve:
@nnmm I've rebased anyway, I'd rather get all the changes in, generating clients and services by itself won't be of much use.

esteve:
@i-am-chauhan this should be almost ready for merging. In any case, your team can help by trying out this branch (or the rest of ros2-rust), we always welcome feedback!

nnmm:
@esteve I was thinking that splitting it up into PRs just for ease of reviewing. This PR will gain many comments otherwise, since it's so large, which could get hard to keep an overview of. You can also make the modules private if you don't want to have the partial features exposed on master.

esteve:
@nnmm I see what you mean and I'd agree, but given that messages are in sync with the API, splitting them wouldn't help much, so I prefer to keep both in the same PR to make sure that any changes in either side match.

nnmm:
Great to see this PR moving forward :)

rclrs/src/future.rs
```diff
@@ -0,0 +1,63 @@
...
+        if let Some(value) = &self.value {
+            Poll::Ready(value.clone())
+        } else {
+            Poll::Pending
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct RclWaker {}
```
nnmm, line 39: I don't quite understand this part, at least the waker seems to have no functionality.

`r2r` and `rclrust` both implement the client future by storing one side of a `oneshot::channel` (that is itself a `Future`) inside the client, and the other inside the  response future. Would something like that maybe be more straightforward?

rosidl_generator_rs/resource/srv.rs.em
```diff
@@ -0,0 +1,37 @@
+@{
+for subfolder, service in srv_specs:
+    msg_specs.append((subfolder, service.request_message))
+    msg_specs.append((subfolder, service.response_message))
+
+TEMPLATE(
```
nnmm, line 6: Wouldn't it be sufficient to only generate the `service.request_message` and `service.response_message` here and not regenerating all of the messages?

rosidl_runtime_rs/src/traits.rs
```diff
@@ -143,3 +144,17 @@ pub trait Message: Clone + Debug + Default + 'static + Send + Sync {
     /// Converts the RMW-native message into an idiomatic message.
     fn from_rmw_message(msg: Self::RmwMsg) -> Self;
 }
+
+pub trait RmwService {
```
nnmm, line 148: This trait shouldn't be needed. Instead, we can generate two structs (idiomatic and RMW-native) that both implement `Service`.

rclrs/src/future.rs
esteve: Yeah, for now the waker is just boilerplate so that `RclFuture` can implement the `Future` trait. The rationale for this is such that once we support ROS2 executors, the `RclFuture` can take reference to the current executor and use it to resolve any pending computation. But your suggestion of using channels is interesting, thanks. I'll explore that, it might be clearer and easier to read than implementing a custom `Future`

rosidl_generator_rs/resource/srv.rs.em
esteve: Not sure I'm following, but you also need a reference to the service type support, hence the rest of the structures in this template.

rosidl_runtime_rs/src/traits.rs
esteve: Messages must have a function that provides a pointer to the underlying type support, which is a different kind of type support from the type support that services provide, hence the need for a separate trait.

nnmm: I know, what I'm getting at is that you have two service traits here, `RmwService` and `Service`, where one would be sufficient. This is what it the generated code should look like imo:
```rust
mod rmw {
  pub struct @(type_name) {...}
  impl Service for @(type_name) { type Request = crate::srv::rmw::@(type_name)Request; ...}
}
pub struct @(type_name) {...}
impl Service for @(type_name) {type Request = crate::srv::@(type_name)Request; ...}
```

rosidl_generator_rs/resource/srv.rs.em
nnmm: You append the `srv_specs` to `msg_specs` and then re-instantiate the `msg.rs.em` template with the extended `msg_specs`. Instead, couldn't you instantiate it only for `srv_specs`?

rclrs/src/future.rs
nnmm: Well, a waker is not required to implement `Future`, it's required for an async executor like `tokio` (and the `spin_until_future_complete` function is kind of an async executor, but only for `RclFuture`).

But my proposal would be, let's first get an MVP of clients merged that doesn't use futures at all, but only callbacks.
After that, let's do a follow-up for futures. The reason being that I don't have the context behind `rclcpp::spin_until_future_complete`, and futures in C++ in general yet to understand what led to this design. Rust's async story seems very different from C++'s, so I can see this being a wider discussion than appropriate for an (already large) PR.

esteve: > Well, a waker is not required to implement Future, it's required for an async executor like tokio (and the spin_until_future_complete function is kind of an async executor, but only for RclFuture).

It is required if you want `.wait` to actually do anything, but that's semantics.

> But my proposal would be, let's first get an MVP of clients merged that doesn't use futures at all, but only callbacks.

That's why I added `async_send_request_with_callback`, which just accepts a callback. I'm not too happy with the future stuff in this PR, but I wanted to get something in that felt familiar to `rclcpp` users.

> The reason being that I don't have the context behind `rclcpp::spin_until_future_complete`

The rationale for `rclcpp::spin_until_future_complete` is that the only way for getting the value out of a future is to call `.get`, which blocks and unless you add threads to the mix, it won't do anything. Given that the first ROS2 executor was single threaded (and it is in fact, the default one), this model wouldn't work, so what `rclcpp::spin_until_future_complete` does is it waits for it to be resolved for a certain time, if it's not resolved, call spin, wait again, etc. until it's resolved and so calling `.get` won't block and will return the value to the caller.

nnmm: Ok. I'd guess `rclcpp` users will probably be fine with just a callback API at first – @i-am-chauhan, as a requester of this feature, let us know if you have any opinion on this. In any case, I'm not suggesting to keep the "future" stuff out of the crates.io release entirely, only out of this PR.

If you're against that suggestion, as a compromise, we could also keep `RclFuture` in this PR, but not implement `Future` on it. The trait is not actually necessary (so you can just implement `poll()` on the type, and it doesn't even need `task::Context`), and might lead people to believe they can use it with a `tokio` runtime, which is not the case.

esteve: I've reworked the futures stuff to use `Future`s from `future-rs` and tokio as the runtime in the examples. I haven't looked into `r2r` yet, but from what I've seen, `rclrust` has implemented executors incorrectly, as each node has its own executor thread which spins (https://github.com/rclrust/rclrust/blob/main/rclrust/src/node.rs#L219-L222=), with this model, you'd have as many threads as nodes.

The `RclFuture`/`rclrs::spin_until_future_complete` implementation has the advantage that no threads are used anywhere. With the latest changes, there's only one thread, that simply calls `rclrs::spin`. With ROS2's executor API, we could hide the thread logic and make it as seamless as possible, but for now we only have `rclrs::spin*`

I still need to do some cleanups, but with the latest changes the API may be easier to understand for people who are used to the Rust async API, but not the ROS2 API.

nnmm: It doesn't matter whether `Future` is from `futures` or `std::task`, `futures` just re-exports the latter one.

Your `Future` implementation is also incorrect, I think. As mentioned above, the waker is provided by the async executor, not the implementation of a future, so no need to define one at all. And the future should, upon first being polled, store the waker so as to notify the async executor when it's ready (i.e. the future would need to give the waker to a wait set, and the wait set needs to use it to wake up the future when the response arrives).

I've started a discussion in the Matrix chat about this, since I think it will be more efficient..

esteve: If look at the latest commits, RclFuture is not use anywhere any more, as I said, there were still some cleanups pending, one of them is removing RclFuture from this PR.

> Your Future implementation is also incorrect, I think. As mentioned above, the waker is provided by the async executor, not the implementation of a future, so no need to define one at all.

And therefore, that's why there's a waker for the ROS2 executor model. Anyway, is it worth discussing this further? I've already replaced RclFuture with Rust's futures.

> I've started a discussion in the Matrix chat about this, since I think it will be more efficient..

Please don't do that in the future, I prefer not moving this discussion to the Matrix channel, it's much less efficient. If you want to have a more general futures vs callbacks discussion, please create a discussion on GitHub so that there's a trail of whatever decisions we make.

nnmm: I wasn't aware you were removing `RclFuture`. But do you intend to keep `spin_until_future_complete`?

esteve: I don't see the point of keeping `spin_until_future_complete` if `RclFuture` is not included, the former only works with `RclFuture`s

i-am-chauhan:
Hi @esteve , I was trying to write some service nodes using this. Things were working then. But when I upgraded to latest of this branch, It's not compiling. And I can see there is lot of changes being made here. So I guess I'll have to wait till it's a bit stable.

esteve:
@i-am-chauhan the latest commit should fix all the compilation issues, there will be more internal cleanups, but I don't expect the API to change much. Could you give it a try again? Thanks!

rclrs/src/node/service.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+impl<T> Service<T>
+where
+    T: rosidl_runtime_rs::Service,
+{
+    pub fn new<F>(node: &Node, topic: &str, callback: F) -> Result<Self, RclrsError>
+    where
+        T: rosidl_runtime_rs::Service,
+        F: FnMut(&rmw_request_id_t, &T::Request, &mut T::Response) + 'static + Send,
```
mxgrey, line 66: I understand this callback signature is meant to match the callback signature used in the original `create_service` definition used by `rclcpp` which has the benefit of being familiar for long-term users of `rclcpp`. However, the `rclcpp` API has come a long way since then and can now support async services with the [defer response](https://github.com/ros2/rclcpp/blob/86a9d5882ed31d1fa6197a1f5d4b388f167bd23e/rclcpp/include/rclcpp/any_service_callback.hpp#L214-L224) callback signatures.

This older fashioned API that uses `&mut T::Response` is unable to support non-blocking async services since it demands that a response be provided by the time the service callback exits. It also comes with the issue that `T::Response` must be getting implicitly default-initialized before it is passed in as a mutable reference. Implicit initialization is (rightly imo) seen as an anti-pattern in Rust.

I would recommend a callback signature that gives the benefits of supporting async services while still allowing ease of use:

```
Fn(&rmw_request_id_t, &T::Request, FnOnce(T::Response))
```

The `FnOnce` given to the callback will ensure
1. A service response is only provided once
2. The correct response is sent for the correct request

Changing to `Fn` instead of `FnMut` for the service callback trait should also help the signature to support multi-threaded executors for single services in the future.

Users that are okay with the service being sync can simply call the `FnOnce(T::Response)` before the function exits. Right now we unfortunately don't have an obvious way to force them to call it, but if the [function parameter attributes RFC](https://rust-lang.github.io/rfcs/2565-formal-function-parameter-attributes.html) makes it into the language then we'll be able to put `#[must_use]` on the response callback in the future. Meanwhile users that want an async service can pass the response callback off to some async function which can trigger the response callback when its work is finished.

Since I'm still not sure if `FnOnce` is usable here, we could alternatively have the response callback be something like

```
enum ResponseErr {
    AlreadyResponded,
}

FnMut(T::Response) -> Result<(), ResponseErr>
```

Inside of `FnMut` we would track whether a response has already been received by this callback and issue a `ResponseErr::AlreadyResponded` and ignore the input if it gets called a second time.

If any of this would be of interest, I'm happy to try prototyping it.

mxgrey: Also if we want a callback signature that is explicitly sync and guarantees an immediate response, I would recommend making it

```
Fn(&rmw_request_id_t, &T::Request) -> T::Response
```

to move away from the implicit initialization problem.

Having both a sync and optionally async callback signature and allowing the user to choose is always nice.

mxgrey (commented):
Apologies for the long delay in reviewing this. I finally had some spare time to look it over.

Getting services into ros2-rust is super exciting! I have some feedback on the callback signatures which loosely amounts to "I wish `rclcpp` had done it this way from the start, so maybe we can have ros2-rust do things in a 'better' (subjective, I know) way from the start".

If my suggestions sound interesting I'll be happy to try making the changes and submit a PR that targets this one.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,255 @@
...
+    ///
+    /// Hence, when a message will not be needed anymore after publishing, pass it by value.
+    /// When a message will be needed again after publishing, pass it by reference, instead of cloning and passing by value.
+    pub fn async_send_request_with_callback<'a, M: MessageCow<'a, T::Request>, F>(
+        &self,
+        message: M,
+        callback: F,
+    ) -> Result<(), RclrsError>
+    where
+        F: FnMut(&T::Response) + 'static + Send,
```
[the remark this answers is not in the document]
esteve: Good catch! Done in 783040c

rclrs/src/node/service.rs
esteve: Good idea about avoiding implicit initialization, I'm rewriting the callback signature

esteve: Done in 4607232

mxgrey:
Makes sense, I'll be happy to do a follow-up PR in the future :+1:

rosidl_generator_rs/resource/srv.rs.em
esteve: Ah, you're right, good catch! I don't remember why I used `msg_specs`, I guess I assumed it was empty, anyway fixed in b0fb8ca

rosidl_runtime_rs/src/traits.rs
nnmm: It should be a different module: `msg::rmw` and `srv::rmw` can coexist, no?

nnmm: I.e. we are already in the `srv` module by virtue of being in `srv.rs.em`.

esteve: Done in ec36503

examples/minimal_client_service/src/minimal_client.rs
```diff
@@ -0,0 +1,31 @@
+use anyhow::{Error, Result};
+use std::env;
+
+fn main() -> Result<(), Error> {
+    let context = rclrs::Context::new(env::args()).unwrap();
```
jhdcs, line 5: Could this be re-written to use `?` instead? I thought we moved away from using `unwrap` in the tutorials.

examples/minimal_client_service/src/minimal_client_async.rs
```diff
@@ -0,0 +1,34 @@
+use anyhow::{Error, Result};
+use std::env;
+
+#[tokio::main]
+async fn main() -> Result<(), Error> {
+    let context = rclrs::Context::new(env::args()).unwrap();
```
jhdcs, line 6: Could this be re-written to use `?` instead? I thought we moved away from using `unwrap` in the tutorials.

examples/minimal_client_service/src/minimal_client_async.rs
```diff
@@ -0,0 +1,34 @@
...
+    println!("Waiting for response");
+
+    let spin_thread = std::thread::spawn(move || rclrs::spin(&node).map_err(|err| err));
+
+    let response = future.await;
+    println!(
+        "Result of {} + {} is: {}",
+        request.a,
+        request.b,
+        response.unwrap().sum
```
jhdcs, line 29: Could this be re-written to use `?` instead? I thought we moved away from using `unwrap` in the tutorials.

Though this would probably require actually rewriting line 24.

examples/minimal_client_service/src/minimal_service.rs
```diff
@@ -0,0 +1,24 @@
...
+    request: &example_interfaces::srv::AddTwoInts_Request,
+) -> example_interfaces::srv::AddTwoInts_Response {
+    println!("request: {} + {}", request.a, request.b);
+    example_interfaces::srv::AddTwoInts_Response {
+        sum: request.a + request.b,
+    }
+}
+
+fn main() -> Result<(), Error> {
+    let context = rclrs::Context::new(env::args()).unwrap();
```
jhdcs, line 15: Could this be re-written to use `?` instead? I thought we moved away from using `unwrap` in the tutorials.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+            rcl_client_fini(handle as *mut _, rcl_node_mtx as *mut _);
+        }
+    }
+}
+
+impl From<Canceled> for RclrsError {
+    fn from(_: Canceled) -> Self {
+        RclrsError::RclError {
+            code: RclReturnCode::Error,
+            msg: None,
```
jhdcs, line 48: Should we actually have a message here? Something to let the user know that the request was cancelled when they're doing something like inspecting the bag files?

Would such a thing be useful?

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+{
+    /// Creates a new client.
+    pub fn new(node: &Node, topic: &str) -> Result<Self, RclrsError>
+    where
+        T: rosidl_runtime_rs::Service,
+    {
+        let mut rcl_client = unsafe { rcl_get_zero_initialized_client() };
+        let type_support = <T as rosidl_runtime_rs::Service>::get_type_support()
+            as *const rosidl_service_type_support_t;
+        let topic_c_string = CString::new(topic).unwrap();
```
jhdcs, line 87: Can this please be re-written without using `unwrap`? I think something similar to https://github.com/ros2-rust/ros2_rust/blob/346e2b5b5448fcb5df0b7b592664aebcd3b51f3d/rclrs/src/node/builder.rs#L238 should work...

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+        let rcl_node = &mut *node.rcl_node_mtx.lock();
```
jhdcs, line 88: Does the mutex need to be locked for the entire rest of the function? Or could we restrict the lock to a smaller block?

Would this be overkill? My thought is that the node is potentially having quite a few things trying to initialize at the same time, and there might be several locks being attempted. Shortening the length of time a lock is active might help to keep the initialization times down.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+
+        unsafe {
```
jhdcs, line 90: This needs a `// SAFETY:` comment.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+    pub(crate) fn lock(&self) -> MutexGuard<rcl_client_t> {
+        self.rcl_client_mtx.lock()
+    }
+}
+
+impl Drop for ClientHandle {
+    fn drop(&mut self) {
+        let handle = self.rcl_client_mtx.get_mut();
+        let rcl_node_mtx = &mut *self.rcl_node_mtx.lock();
+        unsafe {
```
jhdcs, line 38: This needs a `// SAFETY:` comment.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+    pub async fn call_async<'a, R: MessageCow<'a, T::Request>>(
+        &self,
+        request: R,
+    ) -> Result<T::Response, RclrsError>
+    where
+        T: rosidl_runtime_rs::Service + 'static,
+    {
+        let rmw_message = T::Request::into_rmw_message(request.into_cow());
+        let mut sequence_number = self.sequence_number.load(Ordering::SeqCst);
+        let ret = unsafe {
```
jhdcs, line 174: This needs a `// SAFETY:` comment.

rclrs/src/node/service.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub(crate) fn lock(&self) -> MutexGuard<rcl_service_t> {
+        self.handle.lock()
+    }
+}
+
+impl Drop for ServiceHandle {
+    fn drop(&mut self) {
+        let handle = self.handle.get_mut();
+        let node_handle = &mut *self.node_handle.lock();
+        unsafe {
```
jhdcs, line 36: This needs a `// SAFETY:` comment.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+impl<T> Client<T>
+where
+    T: rosidl_runtime_rs::Service,
+{
+    /// Creates a new client.
+    pub fn new(node: &Node, topic: &str) -> Result<Self, RclrsError>
+    where
+        T: rosidl_runtime_rs::Service,
+    {
+        let mut rcl_client = unsafe { rcl_get_zero_initialized_client() };
```
jhdcs, line 84: This needs a `// SAFETY:` comment.

rosidl_generator_rs/resource/srv.rs.em
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  pub struct @(type_name);
+
+  impl rosidl_runtime_rs::Service for @(type_name) {
+    type Request = crate::@(subfolder)::rmw::@(type_name)_Request;
+    type Response = crate::@(subfolder)::rmw::@(type_name)_Response;
+
+    fn get_type_support() -> libc::uintptr_t {
+      unsafe { rosidl_typesupport_c__get_service_type_support_handle__@(package_name)__@(subfolder)__@(type_name)() }
```
jhdcs, line 70: This needs a `// SAFETY:` comment.

jhdcs (changes requested):
This is absolutely fantastic! Thank you so much!

I have a bunch of style/documentation requests, though. Sorry!

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+pub trait ClientBase: Send + Sync {
+    /// Internal function to get a reference to the `rcl` handle.
+    fn handle(&self) -> &ClientHandle;
+    /// Tries to take a new response and run the callback or future with it.
+    fn execute(&self) -> Result<(), RclrsError>;
+}
+
+type RequestValue<Response> = Box<dyn FnOnce(&Response) + 'static + Send>;
+
+/// Main class responsible for publishing data to ROS topics
```
nnmm, line 64: This comment needs to be adjusted

rclrs/src/node/client.rs
nnmm: Am I correct to assume that an occurrence of `Canceled` would basically mean we have a logic error in our code?

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+pub struct Client<T>
+where
+    T: rosidl_runtime_rs::Service,
+{
+    pub(crate) handle: Arc<ClientHandle>,
+    requests: Mutex<HashMap<i64, RequestValue<T::Response>>>,
```
nnmm, line 70: Might be handy to have a comment or better yet type alias to explain what the `i64` means.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+        let handle = Arc::new(ClientHandle {
+            rcl_client_mtx: Mutex::new(rcl_client),
+            rcl_node_mtx: node.rcl_node_mtx.clone(),
+        });
+
+        Ok(Self {
+            handle,
+            requests: Mutex::new(HashMap::new()),
+            futures: Arc::new(Mutex::new(
+                HashMap::<i64, oneshot::Sender<T::Response>>::new(),
```
nnmm, line 112: You could just make this `HashMap::new()` if you want, I believe. But this way is also fine.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+    where
+        F: FnOnce(&T::Response) + 'static + Send,
+    {
+        let rmw_message = T::Request::into_rmw_message(message.into_cow());
+        let mut sequence_number = self.sequence_number.load(Ordering::SeqCst);
+        let ret = unsafe {
+            rcl_send_request(
+                &*self.handle.lock() as *const _,
+                rmw_message.as_ref() as *const <T::Request as Message>::RmwMsg as *mut _,
+                &mut sequence_number,
```
nnmm, line 144: Since this produces a new sequence number, is there any need for us to have the `sequence_number` field at all?

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+    /// |   rcl_take_response  |
+    /// +----------+-----------+
+    ///            |
+    ///            |
+    /// +----------v----------+
+    /// |      rmw_take       |
+    /// +---------------------+
+    pub fn take_response(&self) -> Result<(T::Response, rmw_request_id_t), RclrsError> {
+        let mut request_id_out = rmw_request_id_t {
+            writer_guid: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
```
nnmm, line 204:
```suggestion
            writer_guid: [0; 16],
```

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+        let mut response_out = RmwMsg::<T>::default();
+        let handle = &mut *self.handle.lock();
+        let ret = unsafe {
+            rcl_take_response(
+                handle as *const _,
+                &mut request_id_out,
+                &mut response_out as *mut RmwMsg<T> as *mut _,
+            )
+        };
+        ret.ok()?;
```
nnmm, line 218: In the existing code we have the `ok()?` tacked on right after the function call. I like it because it means there is no danger of forgetting, or accidentally inserting a new line before `ret.ok()?` that continues to do stuff even though an error occurred.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    fn execute(&self) -> Result<(), RclrsError> {
+        let (res, req_id) = match self.take_response() {
+            Ok((res, req_id)) => (res, req_id),
+            Err(RclrsError::RclError {
+                code: RclReturnCode::ClientTakeFailed,
+                ..
+            }) => {
+                // Spurious wakeup – this may happen even when a waitset indicated that this
+                // subscription was ready, so it shouldn't be an error.
```
nnmm, line 239:
```suggestion
                // client was ready, so it shouldn't be an error.
```

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,257 @@
...
+                return Ok(());
+            }
+            Err(e) => return Err(e),
+        };
+        let requests = &mut *self.requests.lock();
+        let futures = &mut *self.futures.lock();
+        if requests.contains_key(&req_id.sequence_number) {
+            requests.remove(&req_id.sequence_number).unwrap()(&res);
```
nnmm, line 247:
```suggestion
        if let Some(callback) = requests.remove(&req_id.sequence_number) {
            callback(&res);
        }
```

nnmm: Same for the futures.

nnmm (commented):
Cool to see the progress with this! I've reviewed up to `client.rs` and like what I've seen so far, you've also kept it nicely consistent with the existing code.

examples/minimal_client_service/src/minimal_client_async.rs
esteve: Done

examples/minimal_client_service/src/minimal_client_async.rs
esteve: Done in d2c72e3ab65e99fbe230ca4270003b1e4de3e225

examples/minimal_client_service/src/minimal_client.rs
esteve: Done in d2c72e3ab65e99fbe230ca4270003b1e4de3e225

rclrs/src/node/client.rs
esteve: Done in 20a09d6ec0798c598f5214c4913b765467d528e1 It used to be the case that the sequence number was not stored in `rcl_client_t` and client libraries needed to keep track of it themselves,, but that was changed.

rclrs/src/node/client.rs
esteve: Not always, for example publisher and subscription use `let ret = unsafe { ... }; ret.ok?'`, so that code should be updated.

rclrs/src/node/client.rs
jhdcs: I don't _think_ so... I think you can get `Canceled` requests in a variety of situations, though I'm not 100% positive how common they are. So it's not necessarily a logic error, just a way to stop the service from doing extra work if it isn't needed.

rclrs/src/node/client.rs
nnmm: Yes, it should.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,253 @@
...
+    ) -> Result<T::Response, RclrsError>
+    where
+        T: rosidl_runtime_rs::Service + 'static,
+    {
+        let rmw_message = T::Request::into_rmw_message(request.into_cow());
+        let mut sequence_number = -1;
+        let ret = unsafe {
+            rcl_send_request(
+                &*self.handle.lock() as *const _,
+                rmw_message.as_ref() as *const <T::Request as Message>::RmwMsg as *mut _,
```
nnmm, line 174: I think this const-to-mut cast can be avoided by making the `rmw_message` variable `mut` and using `as_mut()` instead of `as_ref()`. Same for the service.

nnmm (commented):
Thanks for working on this, looking better and better!

rclrs/src/node/client.rs
nnmm: Yes, definitely.

rclrs/src/node.rs
```diff
@@ -174,6 +180,23 @@ impl Node {
...
+    /// Creates a [`Client`][1].
+    ///
+    /// [1]: crate::Client
+    // TODO: make client's lifetime depend on node's lifetime
+    pub fn create_client<T>(
+        &mut self,
+        topic: &str,
+    ) -> Result<Arc<crate::node::client::Client<T>>, RclrsError>
+    where
+        T: rosidl_runtime_rs::Service + 'static,
```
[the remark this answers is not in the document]
esteve: Done in 8abddd1

rclrs/src/node/client.rs
esteve: Do you mean using `unwrap()` after the future is inserted into the hashmap? If that's the case, yes `unwrap()` should be safe there, otherwise please elaborate.

rclrs/src/node/client.rs
esteve: Done in a49d6caf54b7844d0f31bab71e150bfb23330aa7

rclrs/src/node/client.rs
esteve: Done in 64f0ce0

rclrs/src/node/client.rs
esteve: Done in 2786288

rclrs/src/node/client.rs
nnmm: Yes, `Cow` doesn't have an `AsMut` impl. I think it's fine to leave as it is, since the message should not be modified by `rcl_send_request` anyway.

rclrs/src/node/client.rs
nnmm: Exactly, in the last statement of `call_async()`, I'd write `rx.await.unwrap()`.

rclrs/src/node/client.rs
esteve: Done in 5c261fe

rclrs/src/node/client.rs
esteve: Done in 85d6ed7

rclrs/src/node/client.rs
esteve: Done in d336380

rclrs/src/node/service.rs
```diff
@@ -0,0 +1,173 @@
...
+                // Spurious wakeup – this may happen even when a waitset indicated that this
+                // subscription was ready, so it shouldn't be an error.
+                return Ok(());
+            }
+            Err(e) => return Err(e),
+        };
+        let res = (*self.callback.lock())(&req_id, &req);
+        let rmw_message = <T::Response as Message>::into_rmw_message(res.into_cow());
+        let handle = &mut *self.handle.lock();
+        let ret = unsafe {
```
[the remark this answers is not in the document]
esteve: Done in 9766c33

rclrs/src/node/service.rs
```diff
@@ -0,0 +1,173 @@
...
+        T: rosidl_runtime_rs::Service,
+        F: Fn(&rmw_request_id_t, &T::Request) -> T::Response + 'static + Send,
+    {
+        let mut service_handle = unsafe { rcl_get_zero_initialized_service() };
+        let type_support = <T as rosidl_runtime_rs::Service>::get_type_support()
+            as *const rosidl_service_type_support_t;
+        let topic_c_string = CString::new(topic).unwrap();
+        let node_handle = &mut *node.rcl_node_mtx.lock();
+
+        unsafe {
```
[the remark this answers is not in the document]
esteve: Done in 9766c33

rclrs/src/node/service.rs
```diff
@@ -0,0 +1,173 @@
...
+where
+    T: rosidl_runtime_rs::Service,
+{
+    /// Creates a new service.
+    pub fn new<F>(node: &Node, topic: &str, callback: F) -> Result<Self, RclrsError>
+    where
+        T: rosidl_runtime_rs::Service,
+        F: Fn(&rmw_request_id_t, &T::Request) -> T::Response + 'static + Send,
+    {
+        let mut service_handle = unsafe { rcl_get_zero_initialized_service() };
```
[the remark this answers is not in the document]
esteve: Done in 9766c33

rclrs/src/node/service.rs
esteve: Done in 9766c33

rosidl_generator_rs/resource/msg_rmw.rs.em
```diff
@@ -0,0 +1,85 @@
...
+pub struct @(type_name) {
+@[for member in msg_spec.structure.members]@
+    pub @(get_rs_name(member.name)): @(get_rmw_rs_type(member.type)),
+@[end for]@
+}
+
+impl Default for @(type_name) {
+  fn default() -> Self {
+    unsafe {
+@#    // This is safe since a zeroed bit pattern always forms a valid message.
```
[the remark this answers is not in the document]
esteve: Done in 643c555

rosidl_generator_rs/resource/srv.rs.em
```diff
@@ -0,0 +1,76 @@
...
+}
+
+pub struct @(type_name);
+
+impl rosidl_runtime_rs::Service for @(type_name) {
+  type Request = crate::@(subfolder)::@(type_name)_Request;
+  type Response = crate::@(subfolder)::@(type_name)_Response;
+
+  fn get_type_support() -> libc::uintptr_t {
+    unsafe { rosidl_typesupport_c__get_service_type_support_handle__@(package_name)__@(subfolder)__@(type_name)() }
```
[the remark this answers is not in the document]
esteve: Done in 643c555

rosidl_generator_rs/resource/msg_rmw.rs.em
```diff
@@ -0,0 +1,85 @@
...
+      let mut msg = std::mem::zeroed();
+@#    // This is safe since the precondititons for inti() are fulfilled by giving it a zeroed message.
```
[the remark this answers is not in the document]
esteve: Done in 643c555

rosidl_generator_rs/resource/srv.rs.em
esteve: Done in 643c555

rclrs/src/node/client.rs
esteve: Wanna open a PR for that or shall I do it myself?

examples/minimal_client_service/src/minimal_client_async.rs
```diff
@@ -0,0 +1,32 @@
+use anyhow::{Error, Result};
+use std::env;
+
+#[tokio::main]
```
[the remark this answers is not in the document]
esteve: > This is the thing that I fear is unidiomatic for Rust

Do you have a source for this?

It seems to me that this is a completely valid case, given that tokio itself has a specific API for this and even the example for it is quite similar:

https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

In any case, this is an example of how to integrate the `rclrs` event loop with tokio, users can adapt it to their liking. Other libraries (like r2r), follow a similar pattern (https://github.com/sequenceplanner/r2r/blob/master/examples/client.rs#L42-L52)

rclrs/src/node/client.rs
esteve: Done in 9d3ee714

rclrs/src/node.rs
```diff
@@ -189,6 +212,27 @@ impl Node {
...
+    /// [1]: crate::Service
+    // TODO: make service's lifetime depend on node's lifetime
+    pub fn create_service<T, F>(
+        &mut self,
+        topic: &str,
+        callback: F,
+    ) -> Result<Arc<crate::node::service::Service<T>>, RclrsError>
+    where
+        T: rosidl_runtime_rs::Service + 'static,
+        F: Fn(&rmw_request_id_t, &T::Request) -> T::Response + 'static + Send,
```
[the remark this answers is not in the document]
esteve: > * FnMut is more general than Fn, which is why it's used in Subscription. But it can't run concurrently with itself, which might come up when implementing executors. For now, I'd favor FnMut for consistency and generality.

If you check the feedback that @mxgrey gave (https://github.com/ros2-rust/ros2_rust/pull/146#discussion_r897549598), I originally had an `FnMut`, but I prefer the change to `Fn` as it makes the code simpler. If we end up with a use case where `FnMut` is necessary, we can always update this.

> * As far as I can see, there's no reason to not give an owned T::Request to the callback – it's strictly more powerful/useful. We also do this in subscriptions.

Done in 40cd43cd

esteve:
Thanks everyone for their feedback. I believe I've addressed it all, but please have another look. Thanks!

i-am-chauhan:
Hi @esteve , I tried transforming one of our python nodes in rust node using this. Things are working properly. We haven't checked it thoroughly yet. I'll let you know if we find any issues. 
One thing though, can't we get a function in the client to check whether a service is up and running(similar to rclpy)? 

Note: I'm using an older version of this pr(couple of days old)

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,268 @@
...
+                &mut sequence_number,
+            )
+        }
+        .ok()?;
+        let (tx, rx) = oneshot::channel::<T::Response>();
+        self.futures.lock().insert(sequence_number, tx);
+        Ok(rx.await.unwrap())
```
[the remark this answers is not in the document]
nnmm: Yes, this is the one discussed in https://github.com/ros2-rust/ros2_rust/pull/146#discussion_r899208744 – though I think a comment mentioning why it is infallible would be good.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,268 @@
...
+        };
+        let requests = &mut *self.requests.lock();
+        let futures = &mut *self.futures.lock();
+        if let Some(callback) = requests.remove(&req_id.sequence_number) {
+            callback(&res);
+        }
+        if let Some(future) = futures.remove(&req_id.sequence_number) {
+            future
+                .send(res)
+                .unwrap_or_else(|_| panic!("fail to send response via channel in Client::execute"));
```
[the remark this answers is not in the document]
nnmm: If it is logically guaranteed to not happen then I'm fine with panicking, but we should have a comment explaining why it won't happen. But here I think it can happen.

The docs say "If the receiving end was dropped before this function was called, however, then `Err(t)` is returned." In this case, the receiving end lives in the `call_async()` function, and the receiving end could be dropped if the user doesn't await that function, or drops a future which awaits that function. In these cases, I think we can assume that the user just isn't interested in the result, and do nothing instead of panicking.

examples/minimal_client_service/src/minimal_client_async.rs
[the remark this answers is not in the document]
nnmm: > Do you have a source for this?

I don't have one that addresses this head-on, but for instance the official async Rust book, particularly [the beginning of this chapter](https://rust-lang.github.io/async-book/02_execution/04_executor.html), sounds like it's the responsibility of the async executor (and only the async executor) to run them to completion. (Side note: I'm not saying we should build our own async executor)

I also asked this on the Rust discord whether it's unidiomatic to design a Future such that it can not be driven to completion by an async executor alone, but needs a separate function to run in the background to make progress. Just got one answer, which was "I'd probably be surprised by it, but it's not in principle very different from 'function deadlocks unless certain conditions, please satisfy those conditions' ".  I think the difference here is that a deadlock condition probably cannot be solved with a different design, but we *could* design the future in a way that the separate spin is not needed.

The purpose of `spawn_blocking` is to perform a heavy piece of blocking (=non-async) work inside a `Future`, I don't think it's related to the current question (or at least it needs to be explained to me how it is).

All in all, I still think it's not terrible, but maybe a bit weird. We can merge this and continue in a GitHub discussion about async stuff.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,268 @@
...
+        Ok(Self {
+            handle,
+            requests: Mutex::new(HashMap::new()),
+            futures: Arc::new(Mutex::new(
+                HashMap::<RequestId, oneshot::Sender<T::Response>>::new(),
+            )),
+        })
+    }
+
+    /// Send a requests with a callback as a parameter.
```
nnmm, line 128:
```suggestion
    /// Sends a request with a callback to be called with the response.
```

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,268 @@
...
+    ///
+    /// Hence, when a message will not be needed anymore after publishing, pass it by value.
+    /// When a message will be needed again after publishing, pass it by reference, instead of cloning and passing by value.
+    pub fn async_send_request_with_callback<'a, M: MessageCow<'a, T::Request>, F>(
+        &self,
+        message: M,
+        callback: F,
+    ) -> Result<(), RclrsError>
+    where
+        F: FnOnce(&T::Response) + 'static + Send,
```
nnmm, line 146: Could this get the owned response too?

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,268 @@
...
+            // SAFETY: The request type is guaranteed to match the client type by the type system.
+            rcl_send_request(
+                &*self.handle.lock() as *const _,
+                rmw_message.as_ref() as *const <T::Request as Message>::RmwMsg as *mut _,
+                &mut sequence_number,
+            )
+        };
+        let requests = &mut *self.requests.lock();
+        requests.insert(sequence_number, Box::new(callback));
+        ret.ok()
```
nnmm, line 160: The `ok()` should be called directly on the rcl function. Otherwise if there is an error, we will have this invalid entry in our requests map.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    /// Send a request with a callback as a parameter.
```
nnmm, line 163:
```suggestion
    /// Sends a request and returns the response as a `Future`.
```

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,268 @@
...
+    ///            |
+    ///            |
+    /// +----------v-----------+
+    /// |   rcl_take_response  |
+    /// +----------+-----------+
+    ///            |
+    ///            |
+    /// +----------v----------+
+    /// |      rmw_take       |
+    /// +---------------------+
```
nnmm, line 212: This diagram is nice for rclrs devs, but not that helpful for end users. I propose to put it in regular comments, not doc comments, like `Subscription::take()`.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,268 @@
...
+                &mut sequence_number,
+            )
+        }
+        .ok()?;
+        let (tx, rx) = oneshot::channel::<T::Response>();
+        self.futures.lock().insert(sequence_number, tx);
+        Ok(rx.await.unwrap())
+    }
+
+    /// Ask RMW for the data
```
nnmm, line 198:
```suggestion
    /// Fetches a new response.
    ///
    /// When there is no new message, this will return a
    /// [`ClientTakeFailed`][1].
    ///
    /// [1]: crate::RclrsError
```

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,268 @@
...
+            sequence_number: 0,
+        };
+        type RmwMsg<T> =
+            <<T as rosidl_runtime_rs::Service>::Response as rosidl_runtime_rs::Message>::RmwMsg;
+        let mut response_out = RmwMsg::<T>::default();
+        let handle = &mut *self.handle.lock();
+        unsafe {
+            // SAFETY: The three pointers are valid/initialized
+            rcl_take_response(
+                handle as *const _,
```
nnmm, line 225:
```suggestion
        let handle = &*self.handle.lock();
        unsafe {
            // SAFETY: The three pointers are valid/initialized
            rcl_take_response(
                handle,
```

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,268 @@
...
+        Ok((T::Response::from_rmw_message(response_out), request_id_out))
+    }
+}
+
+impl<T> ClientBase for Client<T>
+where
+    T: rosidl_runtime_rs::Service,
+{
+    fn handle(&self) -> &ClientHandle {
+        self.handle.borrow()
```
nnmm, line 240:
```suggestion
        &self.handle
```

rclrs/src/node/service.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// Internal function to get a reference to the `rcl` handle.
+    fn handle(&self) -> &ServiceHandle;
+    /// Tries to take a new request and run the callback with it.
+    fn execute(&self) -> Result<(), RclrsError>;
+}
+
+type ServiceCallback<Request, Response> =
+    Box<dyn Fn(&rmw_request_id_t, Request) -> Response + 'static + Send>;
+
+/// Main class responsible for responding to requests sent by ROS clients
```
nnmm, line 55:
```suggestion
/// Main class responsible for responding to requests sent by ROS clients.
```

nnmm: Optionally, you could add some more detail on services. See e.g. the docs for `Subscription<T>`. The same for clients.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,268 @@
...
+    fn handle(&self) -> &ClientHandle;
+    /// Tries to take a new response and run the callback or future with it.
+    fn execute(&self) -> Result<(), RclrsError>;
+}
+
+type RequestValue<Response> = Box<dyn FnOnce(&Response) + 'static + Send>;
+
+type RequestId = i64;
+
+/// Main class responsible for sending requests to a ROS service
```
nnmm, line 67:
```suggestion
/// Main class responsible for sending requests to a ROS service.
```

rclrs/src/node/service.rs
```diff
@@ -0,0 +1,185 @@
...
+
+impl<T> Service<T>
+where
+    T: rosidl_runtime_rs::Service,
+{
+    /// Creates a new service.
+    pub fn new<F>(node: &Node, topic: &str, callback: F) -> Result<Self, RclrsError>
+    where
+        T: rosidl_runtime_rs::Service,
+        F: Fn(&rmw_request_id_t, T::Request) -> T::Response + 'static + Send,
```
nnmm, line 73: Using `Fn` here is inconsistent with the use of `FnMut` in subscriptions. I'd like to keep it `FnMut` here too, and to open a follow-up issue for the proposal to change it to `Fn`.

rclrs/src/node/service.rs
```diff
@@ -0,0 +1,185 @@
...
+            sequence_number: 0,
+        };
+        type RmwMsg<T> =
+            <<T as rosidl_runtime_rs::Service>::Request as rosidl_runtime_rs::Message>::RmwMsg;
+        let mut request_out = RmwMsg::<T>::default();
+        let handle = &mut *self.handle.lock();
+        unsafe {
+            // SAFETY: The three pointers are valid/initialized
+            rcl_take_request(
+                handle as *const _,
```
nnmm, line 141:
```suggestion
        let handle = &*self.handle.lock();
        unsafe {
            // SAFETY: The three pointers are valid/initialized
            rcl_take_request(
                handle,
```

rclrs/src/node/service.rs
```diff
@@ -0,0 +1,185 @@
...
+        Ok((T::Request::from_rmw_message(request_out), request_id_out))
+    }
+}
+
+impl<T> ServiceBase for Service<T>
+where
+    T: rosidl_runtime_rs::Service,
+{
+    fn handle(&self) -> &ServiceHandle {
+        self.handle.borrow()
```
nnmm, line 156:
```suggestion
        &self.handle
```

rclrs/src/node/service.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    fn execute(&self) -> Result<(), RclrsError> {
+        let (req, mut req_id) = match self.take_request() {
+            Ok((req, req_id)) => (req, req_id),
+            Err(RclrsError::RclError {
+                code: RclReturnCode::ServiceTakeFailed,
+                ..
+            }) => {
+                // Spurious wakeup – this may happen even when a waitset indicated that this
+                // subscription was ready, so it shouldn't be an error.
```
nnmm, line 167:
```suggestion
                // service was ready, so it shouldn't be an error.
```

rclrs/src/node/service.rs
```diff
@@ -0,0 +1,185 @@
...
+            }
+            Err(e) => return Err(e),
+        };
+        let res = (*self.callback.lock())(&req_id, req);
+        let rmw_message = <T::Response as Message>::into_rmw_message(res.into_cow());
+        let handle = &mut *self.handle.lock();
+        unsafe {
+            // SAFETY: The response type is guaranteed to match the service type by the type system.
+            rcl_send_response(
+                handle as *mut _,
```
nnmm, line 178:
```suggestion
        let handle = &*self.handle.lock();
        unsafe {
            // SAFETY: The response type is guaranteed to match the service type by the type system.
            rcl_send_response(
                handle,
```

rclrs/src/rcl_bindings.rs
```diff
@@ -4,5 +4,6 @@
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
 #![allow(clippy::all)]
+#![allow(missing_docs)]
```
nnmm, line 7: This should be fine, but I'm curious – why is this now needed?

rclrs/src/wait.rs
```diff
@@ -126,6 +144,56 @@ impl WaitSet {
...
+    ///
+    /// This will return an error if the number of clients in the wait set is larger than the
+    /// capacity set in [`WaitSet::new`].
+    ///
+    /// The same client must not be added to multiple wait sets, because that would make it
+    /// unsafe to simultaneously wait on those wait sets.
+    pub fn add_client(&mut self, client: Arc<dyn ClientBase>) -> Result<(), RclrsError> {
+        unsafe {
+            // SAFETY: I'm not sure if it's required, but the subscription pointer will remain valid
+            // for as long as the wait set exists, because it's stored in self.subscriptions.
```
nnmm, line 159:
```suggestion
            // SAFETY: I'm not sure if it's required, but the client pointer will remain valid
            // for as long as the wait set exists, because it's stored in self.clients.
```

rosidl_generator_rs/resource/srv.rs.em
```diff
@@ -0,0 +1,78 @@
...
+@{
+type_name = srv_spec.namespaced_type.name
+}@
+
+#[link(name = "@(package_name)__rosidl_typesupport_c")]
+extern "C" {
+    fn rosidl_typesupport_c__get_service_type_support_handle__@(package_name)__@(subfolder)__@(type_name)() -> libc::uintptr_t;
+}
+
+pub struct @(type_name);
```
nnmm, line 29: Would be nice to add a little documentation here – for messages we have this line:
```
// Corresponds to @(package_name)__@(subfolder)__@(type_name)
```

nnmm (commented):
Mostly nitpicking

esteve:
@i-am-chauhan

> One thing though, can't we get a function in the client to check whether a service is up and running(similar to rclpy)?

This requires support for the ROS graph, which `rclrs` is currently lacking. Would you and your team be interested in contributing it to `rclrs`?

i-am-chauhan:
Hi @esteve , I can add this feature in client and give you PR if you want.

esteve:
@i-am-chauhan 

> Hi @esteve , I can add this feature in client and give you the PR if you want.

That sounds really cool, thanks! It won't be included in this PR since we want to make a first release of `rclrs` soon and will wait until this PR is merged for that.

However, not to discourage you, but adding a function to wait for a service to be available implies adding support for accessing the ROS graph, which requires many parts to be implemented (e.g. guard conditions, most likely a refactor of the executor API, etc.)

I'll be happy to guide you through the process, so feel free to start working on it and I'll give you feedback.

rclrs/src/rcl_bindings.rs
esteve: Because `rmw_request_id_t` is exposed to services, but the C struct is not documented.

rclrs/src/node/client.rs
esteve: Done in 284cecd

rclrs/src/node/client.rs
esteve: Done in 6f2aa36

rclrs/src/node/client.rs
esteve: `Subscription::take()` does have the similar diagram, not a regular comment.

rosidl_generator_rs/resource/srv.rs.em
esteve: Done in 70f41be

rclrs/src/node/client.rs
nnmm: What I mean is it uses double slashes, not triple slashes.

rclrs/src/node/client.rs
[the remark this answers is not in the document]
esteve: Removed panic in e99d1ec

rclrs/src/node/client.rs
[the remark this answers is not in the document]
esteve: Done in 34bdc08

esteve:
All feedback addressed, thanks everyone! Please have another look.

If there any improvements, but are not super urgent, please open follow up tickets, I'd like to merge this soon and wrap up a new release.

rclrs/src/node/client.rs
nnmm: Same for the service.

rclrs/src/node/service.rs
nnmm: Maybe, but don't you agree with the principle of doing changes like `FnMut` -> `Fn` atomically over the code base (services + subscriptions) so that it is never inconsistent across our code? I.e. I think we should open a separate issue (where we can maybe ensure that this is indeed the right thing to do – I don't want to stall this PR with this discussion) and a separate PR which switches to `Fn` (which would also involve removing the Mutex in line 61, for instance).

nnmm:
> All feedback addressed, thanks everyone! Please have another look.
> 
> If there any improvements, but are not super urgent, please open follow up tickets, I'd like to merge this soon and wrap up a new release.

Great! I agree we're really close to merging this. I commented on two threads where I think changes are needed.

rclrs/src/node/client.rs
esteve: Done in 4191195

rclrs/src/node/service.rs
esteve: I appreciate you caring about this so much, but I've already explained why I want to keep it as `Fn` for now, and that I agree with @mxgrey 's feedback (I originally had it as `FnMut`, in fact), but I don't want to block this PR. Feel free to open a new ticket and we can revisit this in the future. Thanks.

esteve:
@nnmm thanks for your feedback! I've applied the `//` change for the comments. Let's open a ticket to revisit the `Fn` / `FnMut` discussion.

nnmm:
Could you also add clients/services it to the list of features in `README.md`?

rclrs/src/node/service.rs
nnmm: I've created a follow-up issue: https://github.com/ros2-rust/ros2_rust/issues/217

i-am-chauhan:
> @i-am-chauhan
> 
> > Hi @esteve , I can add this feature in client and give you the PR if you want.
> 
> That sounds really cool, thanks! It won't be included in this PR since we want to make a first release of `rclrs` soon and will wait until this PR is merged for that.
> 
> However, not to discourage you, but adding a function to wait for a service to be available implies adding support for accessing the ROS graph, which requires many parts to be implemented (e.g. guard conditions, most likely a refactor of the executor API, etc.)
> 
> I'll be happy to guide you through the process, so feel free to start working on it and I'll give you feedback.

Hi @esteve , I have created a [PR](https://github.com/ros2-rust/ros2_rust/pull/218) for this with minimal changes. Please have a look if something needs to change.

examples/minimal_client_service/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+
+[[bin]]
+name = "minimal_service"
+path = "src/minimal_service.rs"
+
+[dependencies]
+anyhow = {version = "1", features = ["backtrace"]}
+tokio = { version = "1", features = ["macros", "rt", "rt-multi-thread", "time"] }
+# Needed for clients
+futures = "0.3"
```
nnmm, line 23: This can be removed

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,275 @@
...
+    {
+        // SAFETY: Getting a zero-initialized value is always safe.
+        let mut rcl_client = unsafe { rcl_get_zero_initialized_client() };
+        let type_support = <T as rosidl_runtime_rs::Service>::get_type_support()
+            as *const rosidl_service_type_support_t;
+        let topic_c_string = CString::new(topic).map_err(|err| RclrsError::StringContainsNul {
+            err,
+            s: topic.into(),
+        })?;
+        let rcl_node = { &mut *node.rcl_node_mtx.lock() };
```
nnmm, line 93:
```suggestion
        let rcl_node = &mut *node.rcl_node_mtx.lock();
```

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,275 @@
...
+                code: RclReturnCode::ClientTakeFailed,
+                ..
+            }) => {
+                // Spurious wakeup – this may happen even when a waitset indicated that this
+                // client was ready, so it shouldn't be an error.
+                return Ok(());
+            }
+            Err(e) => return Err(e),
+        };
+        let requests = &mut *self.requests.lock();
```
nnmm, line 266: No need to change anything, just a remark: The `&mut *` to convert the MutexGuard into a reference is mostly just needed when passing something to an rcl function, here and in a few other places we can also write `self.futures.lock().remove()`.

rclrs/src/node/client.rs
```diff
@@ -0,0 +1,275 @@
...
+        let futures = &mut *self.futures.lock();
+        if let Some(callback) = requests.remove(&req_id.sequence_number) {
+            callback(res);
+        } else if let Some(future) = futures.remove(&req_id.sequence_number) {
+            future.send(res).unwrap_or(());
```
nnmm, line 271: I believe the canonical way to ignore an error is:

```suggestion
            let _ = future.send(res);
```

but no need to change it if you like this way more.

rclrs/src/node/service.rs
```diff
@@ -0,0 +1,191 @@
...
+        let handle = self.handle.get_mut();
+        let node_handle = &mut *self.node_handle.lock();
+        // SAFETY: No preconditions for this function
+        unsafe {
+            rcl_service_fini(handle as *mut _, node_handle as *mut _);
+        }
+    }
+}
+
+/// Trait to be implemented by concrete Service structs
```
nnmm, line 42:
```suggestion
/// Trait to be implemented by concrete Service structs.
///
```

nnmm (commented):
Did my final review, only some nitpicks remaining.

examples/minimal_client_service/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+
+[dependencies.rclrs]
+version = "*"
+
+[dependencies.rosidl_runtime_rs]
+version = "*"
+
+[dependencies.std_msgs]
```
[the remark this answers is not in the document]
esteve: Done in a87318e

examples/minimal_client_service/package.xml
```diff
@@ -0,0 +1,25 @@
...
+  <name>examples_rclrs_minimal_client_service</name>
+  <version>0.2.0</version>
+  <description>Package containing an example of the client-service mechanism in rclrs.</description>
+  <maintainer email="esteve@apache.org">Esteve Fernandez</maintainer>
+  <license>Apache License 2.0</license>
+
+  <build_depend>example_interfaces</build_depend>
+  <build_depend>rclrs</build_depend>
+  <build_depend>rosidl_runtime_rs</build_depend>
+  <build_depend>std_msgs</build_depend>
```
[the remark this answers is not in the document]
esteve: Done in 051b828

rclrs/src/node/client.rs
esteve: Done in bf61831

nnmm (approved):
🚀🚀🚀🚀

jhdcs (approved):

rosidl_generator_rs/resource/srv.rs.em
```diff
@@ -0,0 +1,79 @@
...
+@{
+type_name = srv_spec.namespaced_type.name
+}@
+
+  #[link(name = "@(package_name)__rosidl_typesupport_c")]
+  extern "C" {
+      fn rosidl_typesupport_c__get_service_type_support_handle__@(package_name)__@(subfolder)__@(type_name)() -> libc::uintptr_t;
+  }
+
+  pub struct @(type_name);
```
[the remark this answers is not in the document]
esteve: Done in 51141b4

rclrs/src/node/client.rs
esteve: I can't find the thread, but this was changed based on @jhdcs 's suggestion to not lock the `Node`'s mutex for the entirety of the function and only for the handle instead. We should revisit the other handles in the codebase to make sure we're not locking too much.

nnmm: I agree with the idea, but I believe as long as you hold the mutable reference, the mutex stays locked.

nnmm (approved):

jhdcs (approved):

nnmm:
@esteve Don't you want to merge it? :)

PR #149, opened 2022-05-02
Use ? instead of .unwrap() when creating context in rclrs_examples

nnmm:


jhdcs (approved):
Looks like a simple enough change. Thanks for the PR!

PR #129, opened 2022-04-27
Add detailed guide for building ros2_rust packages

nnmm:
Also extend contributing.md and condense the readme a bit. See [here](https://github.com/ros2-rust/ros2_rust/blob/doc_articles/README.md) for a rendered version.

PR #151, opened 2022-05-04
Removes cd to /src from instructions in README.

jpace121:
Hi!

I recently started trying out ros2-rust and was following the instructions in the README to get started. 

I noticed this line that seemingly tried to cd to /src, which I don't think is the intended instruction...

Thanks!

James

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,6 @@ mkdir src
 git clone https://github.com/ros2-rust/ros2_rust.git src/ros2_rust
 vcs import src < src/ros2_rust/ros2_rust_foxy.repos
 . /opt/ros/foxy/setup.sh
-cd /src
 colcon build --packages-up-to rclrs_examples
 ```
 


nnmm:
I think the intent was to `cd src`, without the leading slash. Then it should work, though the naming of the directory isn't optimal. If this line is removed altogether, the instructions won't work if the working directory is '/'. 

https://github.com/ros2-rust/ros2_rust/pull/129 solves these problems, but it hasn't been merged yet. So, I suggest to either make this `cd src`, or wait for the other PR to get merged.

jpace121:
I’m cool waiting for #129. 

I don’t think the intent was to call colcon build in src though. That seems like a weird way to call colcon that doesn’t really lineup with how it’s used other places.

nnmm:
Agree, it is quite weird, but it was added as a quick fix for the case where the working directory is `/` - and I didn't look closely enough during the review. You can see it in the git history.

OK, then let's close this. Thanks for submitting the fix though.

esteve:
@jpace121 thanks for submitting the fix, running from `src` was definitely not intended. @nnmm until #129 is merged, this one will do, having the workspace on `/` is a quite weird scenario, I'm assuming that's from a Docker container, in which case the workspace should still be in a different folder, not the root

esteve (approved):

nnmm:
#129 is merged, closing this.

PR #152, opened 2022-05-05
Fix unit tests to match againt up to date master branch implementation

Soya-Onishi:
Current this branch unit tests is out of date in some points.
This commit contributes to run unit tests on current master branch implementation.
Also, there are some changes I thought it shoud be done like below points.

- Delete some method tests that are no longer exist (e.g. context::text_context_init())
- Delete default_context in some test module. 
- Add code into existing code (e.g. test_spin_once()).

`default_context()` uses `std::env::args()`, but I think this can cause test opaque in host environment.
Instead of using `std::env::args()`, using empty vec (`vec![]`) is enough for current test I think.
If there are cases to use `std::env::args()`, please teach me the cases.

In previous implementation, `test_spin_once` just create subscriber and spin for 500ms.
To check more comprehensively, publisher and publishing message to subscriber is added.
I also add assertions to check subscribed message is same content compared to published one and check callback of subscriber is called just only one time.


esteve:
@Soya-Onishi thanks for your PR, however the unit tests branch is still marked as draft, I'll rebase it before I submit it for review. I prefer rebasing over merging from master because it keeps the commit history linear, but thanks again for your contribution.

Issue #142, opened 2022-04-30
Add support for node options


nnmm:
Perhaps not all options should be exposed yet (e.g. we don't have rosout logging yet), but something like `use_global_arguments` could definitely be useful.

PR #155, opened 2022-05-06
Add get_domain_id method in Node struct

### LINKED ISSUE #145: Wrap rcl_node_get_domain_id() function
This function can be exposed as an inherent function of `Node` in `node/mod.rs`.

This is a small task, good for getting familiar with ros2_rust development.

Soya-Onishi:
Close #145 

This PR add helper method getting ROS domain ID via rcl_get_node_domain_id().

There is TODO in doc, but this should be done after support for passing node option, I think.

### DIFF
--- a/rclrs/src/node/mod.rs
+++ b/rclrs/src/node/mod.rs
@@ -219,11 +219,10 @@ impl Node {
     ///
     /// # Example
     /// ```
-    /// use rclrs::{Context, RclReturnCode};
-    ///
+    /// # use rclrs::{Context, RclReturnCode};
     /// // set default ROS domain ID to 10 here
     /// std::env::set_var("ROS_DOMAIN_ID", "10");
-    /// let context = Context::new(std::vec![])?;
+    /// let context = Context::new([])?;
     /// let node = context.create_node("domain_id_node")?;
     /// let domain_id = node.get_domain_id()?;
     ///
@@ -234,16 +233,14 @@ impl Node {
     /// TODO: If node option is supported,
     /// add description about this function is for getting actual domain_id
     /// and about override of domain_id via node option
-    pub fn get_domain_id(&self) -> Result<usize, RclReturnCode> {
+    pub fn get_domain_id(&self) -> usize {
         let handle = &*self.handle.lock();
         let mut domain_id: usize = 0;
         unsafe {
-            // SAFETY: Node handler is valid as expected by this function.
-            // If invalid node handler is passed, returns NodeErrorCode::NodeInvalid
-            rcl_node_get_domain_id(handle, &mut domain_id)
+            // SAFETY: No preconditions for this function.
+            rcl_node_get_domain_id(handle, &mut domain_id);
         }
-        .ok()?;
 
-        Ok(domain_id)
+        domain_id
     }
 }
--- a/rclrs/src/node/mod.rs
+++ b/rclrs/src/node/mod.rs
@@ -224,23 +224,22 @@ impl Node {
     /// std::env::set_var("ROS_DOMAIN_ID", "10");
     /// let context = Context::new([])?;
     /// let node = context.create_node("domain_id_node")?;
-    /// let domain_id = node.get_domain_id()?;
-    ///
+    /// let domain_id = node.domain_id();    
     /// assert_eq!(domain_id, 10);
     /// # Ok::<(), RclReturnCode>(())
     /// ```
-    ///
-    /// TODO: If node option is supported,
-    /// add description about this function is for getting actual domain_id
-    /// and about override of domain_id via node option
-    pub fn get_domain_id(&self) -> usize {
+    // TODO: If node option is supported,
+    // add description about this function is for getting actual domain_id
+    // and about override of domain_id via node option
+    pub fn domain_id(&self) -> usize {
         let handle = &*self.handle.lock();
         let mut domain_id: usize = 0;
-        unsafe {
+        let ret = unsafe {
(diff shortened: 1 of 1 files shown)


rclrs/src/node/mod.rs
```diff
@@ -212,4 +212,38 @@ impl Node {
             .filter_map(Weak::upgrade)
             .collect()
     }
+
+    /// Returns the ROS domain ID that the node is using.
+    ///    
+    /// returned values depend on ROS_DOMAIN_ID environment variable.
```
nnmm, line 218: I always like to include a general explanation, like this:

```suggestion
    /// The domain ID controls which nodes can send messages to each other, see the [ROS 2 concept article][1].
    /// It can be set through the `ROS_DOMAIN_ID` environment variable.
    ///
    /// [1]: https://docs.ros.org/en/rolling/Concepts/About-Domain-ID.html
```

rclrs/src/node/mod.rs
```diff
@@ -212,4 +212,38 @@ impl Node {
...
             .collect()
     }
+
+    /// Returns the ROS domain ID that the node is using.
+    ///    
+    /// returned values depend on ROS_DOMAIN_ID environment variable.   
+    ///
+    /// # Example
+    /// ```
+    /// use rclrs::{Context, RclReturnCode};
```
nnmm, line 222: This line is always hidden in the other examples, using `#`. Would be nice to do this here too for consistency.

rclrs/src/node/mod.rs
```diff
@@ -212,4 +212,38 @@ impl Node {
...
+    ///
+    /// // set default ROS domain ID to 10 here
+    /// std::env::set_var("ROS_DOMAIN_ID", "10");
+    /// let context = Context::new(std::vec![])?;
```
nnmm, line 226: Another nitpick for consistency: The other examples use an array for the context, not a vector.

rclrs/src/node/mod.rs
```diff
@@ -212,4 +212,38 @@ impl Node {
...
+    /// let domain_id = node.get_domain_id()?;
+    ///
+    /// assert_eq!(domain_id, 10);
+    /// # Ok::<(), RclReturnCode>(())
+    /// ```
+    ///
+    /// TODO: If node option is supported,
+    /// add description about this function is for getting actual domain_id
+    /// and about override of domain_id via node option
+    pub fn get_domain_id(&self) -> Result<usize, RclReturnCode> {
```
nnmm, line 237: I think this might actually be a candidate for a function that just returns `usize`, see https://github.com/ros2-rust/ros2_rust/blob/doc_articles/docs/Contributing.md#wrapping-rcl-functions

nnmm (commented):
Nice work! I like the example.

rclrs/src/node/mod.rs
Soya-Onishi: Thanks for your more comfortable comments!

rclrs/src/node/mod.rs
Soya-Onishi: nice pick! Thanks!

rclrs/src/node/mod.rs
Soya-Onishi: Do you mean using `#` before `use` statement?

rclrs/src/node/mod.rs
Soya-Onishi: You are right. I thought RCL_RET_ERROR may be returned here before, but I realized that there is no chance by reading rcl source code, so fix to return `usize` merely.

rclrs/src/node/mod.rs
Soya-Onishi: You mean use `#` leading `use` statement?

rclrs/src/node/mod.rs
Soya-Onishi: nice pick! Thanks!

nnmm (commented):
I left a final round of small comments. Once they are done, I'd say this PR is ready.

rclrs/src/node/mod.rs
```diff
@@ -212,4 +212,35 @@ impl Node {
...
+    /// std::env::set_var("ROS_DOMAIN_ID", "10");
+    /// let context = Context::new([])?;
+    /// let node = context.create_node("domain_id_node")?;
+    /// let domain_id = node.get_domain_id()?;
+    ///
+    /// assert_eq!(domain_id, 10);
+    /// # Ok::<(), RclReturnCode>(())
+    /// ```
+    ///
+    /// TODO: If node option is supported,
```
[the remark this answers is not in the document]
Soya-Onishi: I don't know pushed commit will be what you desired format, I try to fix it.

nnmm (approved):

rclrs/src/node/mod.rs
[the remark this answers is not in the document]
nnmm: Yep, exactly.

Soya-Onishi:
Thanks for your patient and meaningful suggestions!

nnmm:
Thanks for your contribution @Soya-Onishi! If you want to do something a bit more complex next, how about the node options (https://github.com/ros2-rust/ros2_rust/issues/142)?

Soya-Onishi:
@nnmm I'll try it, but I think there are what I don't know well , so PR may take more time than this PR.

nnmm:
There is no rush at all, and this issue is just a suggestion, of course. You can also join our Matrix chat, or ask questions in that issue.

PR #158, opened 2022-05-09
Add NodeBuilder for node instantiation

### LINKED ISSUE #157: Node constructor method become verbose?
Through some works for #142 , there is a point I want to ask about.

`Context` and `Node` have constructor methods for `Node` and current implementation is aware of each patterns like only name and with namespace. However, after supporting node options, these constructor method can be 4 patterns like below.

- only name (`Node::new`)
- name and namespace (`Node::new_with_namespace`)
- name and options (need implementation if need)
- name, namespace and options (implement now in [here](https://github.com/Soya-Onishi/ros2_rust/blob/15dad51a979d617ec09e7edc0676253ebc3e1ff5/rclrs/src/node/mod.rs#L64))

For now, I only implement name, namespace and options pattern for passing custom options to node constructor.

Inferring from rclcpp [Node constructor parameters](https://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Node.html), there will not be more parameters, but I feel current implementation has already be about to become verbose.

For simpleness, should those constructors be replaced to builder pattern?

Soya-Onishi:
Close #157 

This PR add builder only for replacing current `Node` constructor, and still not implement node options for #142.
Current implmentation still remain node constructor. 
However, after builder implementation, does normal constructor have demand?
If no, I'll remove constructor, and I also fix some docs and implementations.


Soya-Onishi:
As a milestone, current NodeBuilder supports only namespace and no node options are supported.
After this PR is merged, I'll work on node options support. 

This PR also has current implementation change.
Methods related to `Node` instantiation like `Context::create_node` and `Node::new` only delegate responsivility for instantiation of `Node` to `NodeBuilder`

nnmm:
@Soya-Onishi there seems to have been a mixup with the recently merged PRs, could you undo the accidental changes?

rclrs/src/node/mod.rs
```diff
@@ -40,6 +40,28 @@ pub struct Node {
...
+/// let context = Context::new([])?;
+/// let node = NodeBuilder::new("foo_node", &context).namespace("/bar").build()?;
+/// assert_eq!(node.name(), "foo_node");
+/// assert_eq!(node.namespace(), "/bar");
+/// # Ok::<(), RclReturnCode>(())
+/// ```
+pub struct NodeBuilder<'a> {
+    pub(self) context: &'a Context,
+    pub(self) name: &'a str,
+    pub(self) namespace: &'a str,
```
nnmm, line 62: How about this? Then it doesn't need a lifetime, which I think is nice.

```suggestion
pub struct NodeBuilder<'a> {
    context_handle: Arc<Mutex<rcl_context_t>>,
    name: String,
    namespace: String,
```

I'd also remove `pub(self)`, since leaving it out has the same meaning.

Note that I'm not suggesting to change the function signatures – i.e. `NodeBuilder::new()` would still take a `&Context`, and `NodeBuilder::namespace()` a `&str` etc.

rclrs/src/node/mod.rs
```diff
@@ -312,3 +306,62 @@ impl Node {
...
 }
+
+impl<'a> NodeBuilder<'a> {
+    /// Creates new builder for Node.
+    pub fn new(name: &'a str, context: &'a Context) -> NodeBuilder<'a> {
+        let mut default_domain_id = 0;
+        // SAFETY: No preconditions for this function.
+        let ret = unsafe { rcl_get_default_domain_id(&mut default_domain_id) };
+        debug_assert_eq!(ret, 0);
+
```
nnmm, line 317: These lines about the domain id seem to be currently unused, correct?

rclrs/src/node/mod.rs
```diff
@@ -40,6 +40,28 @@ pub struct Node {
     pub(crate) subscriptions: Vec<Weak<dyn SubscriptionBase>>,
 }
 
+/// A builder for node instantiation.
+///
+/// default value in each fields are
```
nnmm, line 45:
```suggestion
/// The default values for each field are
```

nnmm (commented):
Thanks for the PR! I left some initial remarks.

Also, feel free to explicitly request a review on your PRs when they're ready, otherwise it's easy for us to miss them.

Soya-Onishi:
Sorry for lack of knowledge about git.  Is this form desired commits which @nnmm mentioned?

rclrs/src/node/mod.rs
Soya-Onishi: I misunderstood field visibility of public struct.
I also agree not to use reference.
Is it ok to think that performance does not particularly important for builder?

rclrs/src/node/mod.rs
Soya-Onishi: I forgot that. I'll fix it.

rclrs/src/node/mod.rs
Soya-Onishi: Thanks for your suggestion!

nnmm (commented):
Functionality-wise, this looks good to me. I'd only delete the `Node::new_with_namespace()` and `Context::create_node_with_namespace()` functions. 

However, I think we should aim for making the documentation excellent as well.
- The "Naming" section of the `Node::new_with_namespace()` function's docs should probably be moved to the `Node` struct
- The `NodeBuilder::namespace()` function should reference the naming rules for namespaces, and mention that validation only happens later, in the `build()` function.
- The `NodeBuilder::new()` function should do the same for the node name, since it takes the node name as an argument
- Every function should have an example/doctest. The example from `Node::new_with_namespace()` can be reused for these.
- Convenience functions should reference the original functions they're equivalent to (in this case, that's only `Context::create_node_builder()`, which I've left a suggestion for)

rclrs/src/context.rs
```diff
@@ -138,7 +138,30 @@ impl Context {
...
+    /// # Example
+    /// ```
+    /// # use rclrs::{Context, RclReturnCode};
+    /// let context = Context::new([])?;
+    /// let node_builder = context.create_node_builder("foo_node");
+    /// let node = node_builder.build()?;
+    /// assert_eq!(node.name(), "foo_node");
+    /// # Ok::<(), RclReturnCode>(())
+    /// ```
+    pub fn create_node_builder(&self, node_name: &str) -> NodeBuilder {
```
esteve, line 163: In the same way as before, `NodeBuilder`s should be created via the `Context`. Moreover, `rclcpp` uses `NodeFactory` to instantiate `Node`s, so we could replicate that API to make it more conventional.

rclrs/src/node/mod.rs
```diff
@@ -312,3 +305,57 @@ impl Node {
         domain_id
     }
 }
+
+impl NodeBuilder {
+    /// Creates new builder for Node.
+    pub fn new(name: &str, context: &Context) -> NodeBuilder {
+        NodeBuilder {
+            context: context.handle.clone(),
+            name: name.to_string(),
```
esteve, line 314: These member variables get converted from `&str` to `String` and then back to `&str` when building the `Node`, it'd be cleaner if either `&str` or `String` are used consistently everywhere.

esteve (changes requested):

esteve:
@Soya-Onishi thank you so much for your contribution! :slightly_smiling_face:

rclrs/src/context.rs
```diff
@@ -115,7 +115,7 @@ impl Context {
     /// # Ok::<(), RclReturnCode>(())
     /// ```
     pub fn create_node(&self, node_name: &str) -> Result<Node, RclReturnCode> {
```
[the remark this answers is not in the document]
nnmm: Let's maybe create a new issue for this. I think it matters that we don't have a default global context like `rclcpp` and `rclpy`. That means that these methods offer a significant advantage in convenience.

rclrs/src/node/mod.rs
nnmm: I'm leaning in the other direction. The cost is irrelevant here, and makes the API easier (not having to keep the owner of the `&str` alive, not having to type the lifetime parameter).

esteve: It feels wrong that there are so many conversions back and forth. Instead, `&str` can be declared as `static` not avoid keeping the lifetimes.

rclrs/src/context.rs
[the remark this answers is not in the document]
esteve: I don't like deviating too much from other APIs, one of the advantages of `rclrs` versus other bindings is that the developer experience is more consistent with the rest of the ecosystem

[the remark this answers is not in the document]
nnmm: I know you don't like that :) I feel like there are also good arguments for keeping these functions, so let's discuss this elsewhere to not derail this PR. I created https://github.com/ros2-rust/ros2_rust/issues/160.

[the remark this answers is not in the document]
esteve: Filed in https://github.com/ros2-rust/ros2_rust/issues/161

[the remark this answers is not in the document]
esteve: @nnmm we both filed it at the same time :smile:

rclrs/src/node/mod.rs
nnmm: That would limit the API to not handle user-provided or randomly-generated node names.

I have the impulse to avoid conversions wherever possible as well – in most case it makes the code cleaner and smaller, but I think the impulse should be dismissed when it conflicts with a better API and performance is not a concern.

It's irrelevant from a practical perspective, but if it helps with this "feeling wrong", there is only one real conversion per field (the conversion from `String` to `&str` is free), and the `String` will not live forever and take up stack space, since the builder will be dropped pretty much immediately.

rclrs/src/context.rs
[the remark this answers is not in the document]
nnmm: Okay, let's use yours since it has a better description.

rclrs/src/node/mod.rs
jhdcs: Forgive me if this is a silly suggestion, but would a `cow` be of use in this instance?

Soya-Onishi: I think this struct will not affect performance because, in many case, this builder is used at initialization part of process lifetime. 

If @jhdcs's suggestion, using Cow, means for more user friendliness, I agree that and I think it us better to use impl Into<String>  instead, but if it means performance reason, it is excess I think.

Anyway, This should be discussed at another refactoring issue or PR. Especially, struct field is inner implementation and out of scope on user perspective.

nnmm: I think this struct is perfect as it is. Making the struct member and function parameter a `Cow` would avoid conversions, but the conversions are irrelevant and it would have similar ergonomics drawbacks (or worse) as a `&str` struct member. 

Making the function parameter an `impl Into<String>`, or `AsRef<str>` would allow also passing in `String` directly, but you can also just put an `&` before your String to make it coerce into a `&str`, so it doesn't buy us much. And in reality most people will pass in a string literal, so the extra complexity of the generic function signature would only be distracting to them.

rclrs/src/node/mod.rs
```diff
@@ -312,3 +305,57 @@ impl Node {
...
+            namespace: "/".to_string(),
+        }
+    }
+
+    /// Sets node namespace.
```
nnmm, line 319: I suggest to move the `Naming` documentation section from the `Node::new_with_namespace` function to the `Node` struct, to delete the `Node::new_with_namespace` function, and to add the following documentation here:

```suggestion
    /// Sets the node namespace.
    ///
    /// The node namespace needs to fulfill the criteria of
    /// [`rmw_validate_namespace()`][1], otherwise an error will be returned by the
    /// [`NodeBuilder::build`] function.
    ///
    /// A namespace without a leading forward slash is automatically changed to have a leading
    /// forward slash.
    ///
    /// See also the [section on node naming][2].
    ///
    /// # Example
    /// ```
    /// # use rclrs::{Context, NodeBuilder, RclReturnCode, NodeErrorCode};
    /// let context = Context::new([])?;
    /// let node_builder = NodeBuilder::new(&context, "my_node");
    /// // This is a valid namespace
    /// assert!(node_builder.namespace("/my/nested/namespace").build().is_ok());
    /// // This namespace contains invalid characters
    /// assert_eq!(
    ///     node_builder.namespace("/röböt").build(),
    ///     Err(RclReturnCode::NodeError(NodeErrorCode::NodeInvalidNamespace))
    /// );
    /// // This namespace starts with a number
    /// assert_eq!(
    ///     node_builder.namespace("/123/4").build(),
    ///     Err(RclReturnCode::NodeError(NodeErrorCode::NodeInvalidNamespace))
    /// );
    /// # Ok::<(), RclReturnCode>(())
    /// ```
    ///
    /// [1]: https://docs.ros2.org/latest/api/rmw/validate__namespace_8h.html
    /// [2]: crate::Node#naming
```

nnmm: The example won't compile yet – you probably need to at least derive `Clone` for `NodeBuilder` and add `.clone()` calls in there.

nnmm:
@Soya-Onishi Seems like there are again some accidental changes from resolving conflicts with master.

Soya-Onishi:
From some searches, I cannot found the way to rebase cleanly and merge is clearner than rebase in this case(conflict appear after creating PR).

nnmm:
Also, thanks for fixing the error type in the doctests!

rclrs/src/node/mod.rs
```diff
@@ -40,6 +40,28 @@ pub struct Node {
...
+/// # use rclrs::{Context, Node, NodeBuilder, RclrsError};
+/// let context = Context::new([])?;
+/// let node = NodeBuilder::new("foo_node", &context).namespace("/bar").build()?;
+/// assert_eq!(node.name(), "foo_node");
+/// assert_eq!(node.namespace(), "/bar");
+/// # Ok::<(), RclrsError>(())
+/// ```
+pub struct NodeBuilder {
+    context: Arc<Mutex<rcl_context_t>>,
+    name: String,
```
nnmm, line 61: We could make `name` and `namespace` public, that way not only chaining style is supported, but also assignment style.

rclrs/src/node/mod.rs
Soya-Onishi: As for now, struct field types use current ones, not reference and others suggested here. It is the best to discuss in another issue dedicates for this topic, I think.

nnmm:
@soya-onishi Would you be okay with me pushing some documentation improvements directly to this branch (no code changes)? That would be easier than commenting here.

rclrs/src/node/mod.rs
nnmm: Well, I don't think encapsulation matters much here: Everybody already knows that `name` and `namespace` are part of the builder, and there is no invariant of `NodeBuilder` that needs to be ensured by the public API (every `String` is valid here since validation only happens in `build()`).

But I'm also fine with leaving it this way.

Soya-Onishi:
@nnmm Sure, please push your commits.

nnmm:
@Soya-Onishi Just an update, I didn't manage to do the doc improvements yet, but probably tomorrow.

nnmm:
@Soya-Onishi Done.

From my side, this is now almost good to go. I have three suggestions:
- Could you create a new file for the node builder (`node/builder.rs`)? It might require marking a few fields in `node/mod.rs` as `pub(super)`, but that's okay.
- Could you swap the arguments to `NodeBuilder::new()`?
- We could have a `Node::builder(context: &Context, node_name: &str) -> NodeBuilder` method instead of `NodeBuilder::new()`. That is a common pattern because it's easier to discover when looking at the `Node` struct, and it's also slightly shorter,

Soya-Onishi:
@nnmm Should I swap arguments of `Node::new()`

nnmm:
> @nnmm Should I swap arguments of Node::new() also?

Yes, please.

nnmm:
@esteve Could you maybe re-review and see if you feel like you can remove the *changes requested* status? Then I'd do a final review of documentation.

rclrs/src/node/mod.rs
```diff
@@ -311,4 +255,23 @@ impl Node {
         debug_assert_eq!(ret, 0);
         domain_id
     }
+
+    /// Creates [`NodeBuilder`][1]
+    ///
+    /// This is convienience function equivalent to [`NodeBuilder::new()`][2].
```
nnmm, line 261:
```suggestion
    /// Convenience function equivalent to [`NodeBuilder::new()`][2].
```

rclrs/src/node/mod.rs
```diff
@@ -311,4 +255,23 @@ impl Node {
         debug_assert_eq!(ret, 0);
         domain_id
     }
+
+    /// Creates [`NodeBuilder`][1]
```
nnmm, line 259:
```suggestion
    /// Creates a [`NodeBuilder`][1] with the given name.
```

rclrs/src/node/builder.rs
```diff
@@ -0,0 +1,190 @@
...
+    pub fn namespace(mut self, namespace: &str) -> Self {
+        self.namespace = namespace.to_string();
+        self
+    }
+
+    /// Builds the node instance.
+    ///
+    /// Node name and namespace validation is performed in this method.
+    ///
+    /// For example usage, see the [`Node::builder()`][1] docs.
```
nnmm, line 149:
```suggestion
    /// For example usage, see the [`NodeBuilder`][1] docs.
```

rclrs/src/node/builder.rs
```diff
@@ -0,0 +1,190 @@
...
+    ///
+    /// # Panics
+    /// When the node name or namespace contain null bytes.
+    ///
+    /// [1]: crate::Node::builder
```
nnmm, line 154:
```suggestion
    /// [1]: crate::NodeBuilder
```

nnmm (commented):
Only four small improvements, then this is good to go from my side!

nnmm (approved):
Looks good to me!

@esteve ptal!

jhdcs (approved):

esteve (approved):

esteve:
@Soya-Onishi thanks for iterating and addressing our feedback!

Soya-Onishi:
Thanks for your suggestions and cooperation!!

PR #162, opened 2022-05-13
Reset error state when converting to RclReturnCode

nnmm:
This can be tested by making two failing calls to rcl after another, e.g. 

```
Context::new(["--ros-args", "-p", "foo&=3"].map(String::from));
Context::new(["--ros-args", "-p", "foo&=3"].map(String::from));
```

Before this change, an error message would be logged about overwriting the error state.


jhdcs (dismissed):
Ooh! Good catch! Once CI passes (formatting) this should be good to go!

nnmm:
Thx, appreciate the quick review :) Done.

jhdcs (approved):

PR #163, opened 2022-05-13
Move rclrs_examples to examples/minimal_pub_sub

nnmm:
We want to add new packages for further examples, which can then go into the `examples` directory. I also renamed the package to better follow the naming schema of `rclcpp` and `rclpy` examples.


jhdcs (approved):

nnmm:
Thanks for the quick review! I'm giving @esteve a day or so to leave comments since I think he might have some.

nnmm:
Forgot to adapt the README too, done now.

nnmm:
@jhdcs Could you re-review? I realized that I should create a new package for the message demo, since it's not related to pub-sub.

esteve (approved):

esteve:
@jhdcs @nnmm what do you think of moving the examples to a separate repo? Like https://github.com/ros2/examples does

jhdcs (approved):

jhdcs:
I think that might be a good idea! It would keep the crate size down, and follow how ROS does things.

esteve:
I've created https://github.com/ros2-rust/examples, we can move the examples there later once this PR is merged.

nnmm:
I'm fine with a separate repo.

Issue #111, opened 2022-04-20
Get error messages from rcl layer


nnmm:
The `rcl` layer does not only return an error code, but often also sets the error message (with `RCL_SET_ERROR_MSG`). `rclrs` currently ignores this error message, but it could be valuable for developers.

It should be made part of the error type, and populated with `rcutils_get_error_string()` or `rcutils_get_error_state()`.

PR #164, opened 2022-05-13
Get error string from rcl layer

nnmm:
Storing the error string requires a new struct, which I've named `RclrsError`.

This also makes `Display` impls use a period instead of an exclamation mark.

Closes https://github.com/ros2-rust/ros2_rust/issues/111

### DIFF
--- a/rclrs/src/context.rs
+++ b/rclrs/src/context.rs
@@ -1,5 +1,5 @@
 use crate::rcl_bindings::*;
-use crate::{Node, RclReturnCode, ToResult};
+use crate::{Node, RclrsError, ToResult};
 
 use std::ffi::CString;
 use std::os::raw::c_char;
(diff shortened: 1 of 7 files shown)


nnmm:
As an example (I added `.context("Failed to create context")` and `.context("Failed to do awesome thing")` in appropriate places in the minimal subscriber, so the chain is longer than just two entries):

```
cargo run --bin minimal_subscriber -- --ros-args -p plem
   Compiling rclrs_examples v0.2.0 (/home/user/workspace/src/ros2_rust/rclrs_examples)
    Finished dev [unoptimized + debuginfo] target(s) in 2.72s
     Running `target/debug/minimal_subscriber --ros-args -p plem`
[ERROR] [1652473951.278531464] [rcl]: Failed to parse global arguments
Error: Failed to do awesome thing

Caused by:
    0: Failed to create context
    1: RclReturnCode::ParsingError: Found invalid ROS argument while parsing.
    2: Couldn't parse parameter override rule: '-p plem'. Error: Expected lexeme type (22) not found, search ended at index 4, at /tmp/binarydeb/ros-galactic-rcl-3.1.2/src/rcl/lexer_lookahead.c:233
```

`2` is the error string from the rcl layer.

jhdcs (approved):

esteve (approved):

PR #169, opened 2022-05-16
168 add humble hawksbill to ci

### LINKED ISSUE #168: Update CI with humble and also update rolling image to jammy
- rolling was switched over to Ubuntu jammy so we should update the docker image in CI
- we can add humble as well to the list of distros to test

flynneva:
- update rolling to ubuntu jammy in CI
- add humble to CI

Closes #168


esteve (approved):

esteve:
@flynneva thanks!

esteve:
Looks like I pulled the trigger too quickly, @flynneva something must have changed in Jammy that makes apt fail. I can look into it when I get home, but I'd appreciate if you can have a look too, thanks!

PR #171, opened 2022-05-19
Make all the things Send, and messages Sync as well

nnmm:
This is required to make multithreading work. For instance, calling publish() on a separate thread doesn't work without these impls.

This is the first time I've needed to implement these traits, but I read up on them and I think this is correct. I believe `rcl` types don't generally care which thread they're on, and therefore can be `Send`.

### DIFF
--- a/rosidl_runtime_rs/src/string.rs
+++ b/rosidl_runtime_rs/src/string.rs
@@ -213,6 +213,11 @@ macro_rules! string_impl {
             }
         }
 
+        // SAFETY: A string is a simple data structure, and therefore not thread-specific.
+        unsafe impl Send for $string {}
+        // SAFETY: A string does not have interior mutability, so it can be shared.
+        unsafe impl Sync for $string {}
+
         impl SequenceAlloc for $string {
             fn sequence_init(seq: &mut Sequence<Self>, size: libc::size_t) -> bool {
                 // SAFETY: There are no special preconditions to the sequence_init function.
--- a/rclrs/src/context.rs
+++ b/rclrs/src/context.rs
@@ -24,6 +24,10 @@ impl Drop for rcl_context_t {
     }
 }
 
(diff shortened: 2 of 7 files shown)


rosidl_runtime_rs/src/string.rs
```diff
@@ -213,6 +213,11 @@ macro_rules! string_impl {
             }
         }
 
+        // SAFETY: A string is a simple data structure, and therefore not thread-specific.
+        unsafe impl Send for $string {}
```
jhdcs, line 217: Forgive the potentially silly question, but I am not familiar with the `$string` syntax. What does it mean?

jhdcs (changes requested):
Is there a way that we can set up some tests to make sure that our assumptions about `Send` and `Sync` on the `rcl` types are accurate? As in, we set up a few scenarios that can only be done if the external type is truly `Send` or `Sync`, and make them part of CI? That way, if something changes within `rcl` that we don't expect, we don't get caught by surprise - or we can potentially warn `rcl` if that was not an intended change.

nnmm:
> Is there a way that we can set up some tests to make sure that our assumptions about Send and Sync on the rcl types are accurate? As in, we set up a few scenarios that can only be done if the external type is truly Send or Sync, and make them part of CI? That way, if something changes within rcl that we don't expect, we don't get caught by surprise - or we can potentially warn rcl if that was not an intended change.

No, not easily. I think maybe (though I haven't checked) it's something that [Miri](https://github.com/rust-lang/miri) can detect, but setting that up would be a larger undertaking.

rosidl_runtime_rs/src/string.rs
nnmm: This is inside a macro. `$string` can be either the `String` or `WString` type.

esteve (approved):

jhdcs (approved):
Well, in that case, I'll approve the changes as-is. We can always look into it in a new PR.

PR #174, opened 2022-05-21
Tutorial on writing a node with rclrs

nnmm:
This is inspired by a question @AlexKaravaev had a while ago.


nnmm:
@tylerjw feel free to leave comments if you like.

docs/writing-your-first-rclrs-node.md
```diff
@@ -0,0 +1,214 @@
...
+
+Additionally, create a new `package.xml` if you want your node to be buildable with `colcon`. See the official [tutorial on creating a ROS 2 package](https://docs.ros.org/en/rolling/Tutorials/Creating-Your-First-ROS2-Package.html) for the structure of this file. Make sure to change the build type to `ament_cargo` and the list of dependencies to the two packages mentioned above, as such:
+
+```xml
+<depend>rclrs</depend>
+<depend>std_msgs</depend>
+
+<export>
+  <build_type>ament_cargo</build_type>
+</export>
```
tylerjw, line 25: I wonder how hard it would be to extend `ros2 pkg` to have a `cargo` build type that would generate all this for you.

Either way, for the ease of going through this tutorial would you mind if this block was extended to include the rest of a minimal `package.xml` for this tutorial?

```xml
<package format="3">
 <name>republisher_node</name>
 <version>0.0.0</version>
 <description>TODO: Package description</description>
 <maintainer email="user@todo.todo">user</maintainer>
 <license>TODO: License declaration</license>

 <depend>rclrs</depend>
 <depend>std_msgs</depend>

 <export>
  <build_type>ament_cargo</build_type>
 </export>
</package>
```

docs/writing-your-first-rclrs-node.md
```diff
@@ -0,0 +1,214 @@
...
+
+If you couldn't follow the explanation involving borrowing, closures etc. above, an explanation of these concepts is unfortunately out of scope of this tutorial. There are many good Rust books and tutorials that can help you understand these crucial features.
+
+## Periodically run a republishing function
+
+The node still doesn't republish the received messages. First, let's add a publisher to the node:
+
+```rust
+// Add this new field to the RepublisherNode struct:
+publisher: rclrs::Publisher<StringMsg>,
```
tylerjw, line 147: Tiny detail, I noticed based on how you construct the struct later that this needs to be placed between the _subscription and data fields. Maybe update this to show the whole new struct with this line added to make it clearer?

docs/writing-your-first-rclrs-node.md
```diff
@@ -0,0 +1,214 @@
...
+```rust
+fn republish(&self) -> Result<(), rclrs::RclrsError> {
+    if let Some(s) = &*self.data.lock().unwrap() {
+        self.publisher.publish(s)?;
+    }
+    Ok(())
+}
+```
+
+What's left to do is to call this function every second. `rclrs` doesn't have built-in functionality to do run a function at a fixed interval, but it's easy enough to achieve with a thread and the sleep function. Add `use std::time::Duration;` to the imports, and change your main function to spawn a separate thread:
```
tylerjw, line 170: All other code changes in this tutorial are called out with small blocks (instead of being inline) even when they are one-line changes. For consistency maybe consider changing this addition to also be a code block?

tylerjw (commented):
Thank you for creating this tutorial. I was able to follow the tutorial and get it to work. I left some comments on some minor things I found. Thank you again for creating this, this answered several of my questions about the interfaces in rclrs and how to set up a minimal project.

docs/writing-your-first-rclrs-node.md
```diff
@@ -0,0 +1,214 @@
...
+  <build_type>ament_cargo</build_type>
+</export>
+```
+
+
+## Writing the basic node structure
+
+Since Rust doesn't have inheritance, it's not possible to inherit from `Node` like you would in `rclcpp` or `rclpy`.
+
+Instead, you can store the node as a regular member. Let's add a struct that contains the node, a subscription, and a field for the last message that was received to `main.rs`:
```
tylerjw, line 33: This is also consistent with modern guidance for writing C++. I don't understand why the official tutorials use inheritance for the Node object when composition works just as well.

Rust doesn't have inheritance (yet), but even once it does I think it would be advisable to avoid encouraging people to use inheritance in this case.

docs/writing-your-first-rclrs-node.md
nnmm: Yep, good suggestion, it's not that long.

docs/writing-your-first-rclrs-node.md
nnmm: Ok, I rephrased it to say "as is common practice in `rclcpp` or `rclpy`", so as to not endorse (or discourage) the inheritance approach in these languages.

docs/writing-your-first-rclrs-node.md
nnmm: I agree. I moved it into the main function, so that the changes aren't so scattered. I like these "local" `use` statements better anyway when there's only one place that it will be used in.

tylerjw (dismissed):
lgtm, thank you for the updates and making this.

docs/writing-your-first-rclrs-node.md
```diff
@@ -0,0 +1,224 @@
+# Writing your first `rclrs` node
+
+This tutorial is intended to point out some of the differences of the Rust client library with the other client libraries. It assumes knowledge of Rust, and is also not intended to be an introduction to ROS 2.
+
+As a semi-realistic example, let's create a node that periodically republishes the last message it received. It's limited to only one specific message type – `std_msgs/msg/String` in this example.
+
+## Create a package
+
```
esteve, line 8: Would be good to use the standard ROS2 tools for creating packages:

```
ros2 pkg create --build-type ament_cargo <package_name>
```

More info at https://docs.ros.org/en/galactic/Tutorials/Creating-Your-First-ROS2-Package.html#create-a-package

Though this may require some sort of integration with `ros2` so that it's gets expanded with actual Rust code. Haven't looked up how to wire up `ros2` with build types.

esteve (changes requested):

docs/writing-your-first-rclrs-node.md
esteve: @tylerjw The reason the `rclcpp` and `rclpy` docs encourage inheritance is because this way initialization is homogeneous as it's delegated to the base `Node` class, and thus enables creating components without requiring to provide a custom `main` function.

It's highly unlikely that Rust will ever have inheritance, however as an improvement for the `rclrs` API we could provide default implementations for node traits:

https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations

tylerjw: In order to create a class that has the interface of a node in rclcpp and therefore can be a component node all you have to do is provide one method: 

```
rclcpp::node_interfaces::NodeBaseInterface::SharedPtr get_node_base_interface()
  {
    return node_->get_node_base_interface();
  }
```

All the rest of the node interface is unnecessary for making a component node.

tylerjw: It is fair to say that the details of how to use composition to achieve a component node are not documented and technically harder to understand for new developers than just inheriting from the Node... I do think this sort of thinking bleeds into all sorts of the ROS interfaces and that results in some sad tradeoffs that have been made in various interfaces.

When writing tutorials I try really hard to write idiomatic code for the language I'm using while still striving to make it as simple as possible. I don't think beginners should be encouraged to use inheritance in this way though even in tutorials.

Anyway, I don't know much about rclpy but I have been trying to fix PickNik and MoveIt's usage of C++ to be more inline with modern guidelines for using the language. I'll get off my soap box there. I did read somewhere that rust might get inheritance soon but now I can't find a link. If it doesn't, I'll be plenty happy with that as I really like the type trait system.

docs/writing-your-first-rclrs-node.md
nnmm: @esteve I agrere that would be good, but let's focus on the tutorial article here please. Doing that integration isn't the goal of this PR.

docs/writing-your-first-rclrs-node.md
nnmm: I always learn something new in these reviews :) However, I'll resolve this thread since it's a bit of a tangent.

docs/writing-your-first-rclrs-node.md
esteve: @nnmm I didn't suggest working on this integration in this PR, but thought it was appropriate to point out that the ROS2 way of creating a new package is via `ros2 pkg create`, not via creating files manually. Please add a comment to the tutorial referencing `ros2 pkg create`, but pointing out that until the integration is done, users will have to edit files manually.

esteve: @nnmm thanks :slightly_smiling_face:

esteve (changes requested):

docs/writing-your-first-rclrs-node.md
```diff
@@ -0,0 +1,226 @@
...
+},
+```
+
+This is a standard pattern in C++, but doesn't work in Rust. Why not?
+
+Written like this, `data` is *borrowed* by the callback, but `data` is a local variable only exists in its current form until the end of `RepublisherNode::new()`. The subscription callback is required to not borrow any variables, because the subscription, and therefore the callback, can live indefinitely.
+
+> 💡 As an aside, this requirement is expressed by the `'static` bound on the generic parameter `F` for the callback in `Node::create_subscription()`.
+
+You might think "I don't want to borrow from the local variable `data` anyway, I want to borrow from the `data` field in `RepublisherNode`!" and you would be right. That variable lives considerably longer, but also not forever, so it wouldn't be enough. A secondary problem is that it would be a *self-referential struct*, which is not allowed in Rust.
```
jhdcs, line 102: Would it be useful for people seeing this to be directed to [this series](https://rust-unofficial.github.io/too-many-lists/) as to why self-referential structs (and trying to squeeze them into Rust in any way the type system will let you) is not a good thing?

docs/writing-your-first-rclrs-node.md
```diff
@@ -0,0 +1,226 @@
...
+            _subscription,
+            data,
+        })
+    }
+}
+```
+
+If that seems needlessly complicated – maybe it is, in the sense that `rclrs` could potentially introduce new abstractions to improve the ergonomics of this use case. This is to be discussed.
+
+If you couldn't follow the explanation involving borrowing, closures etc. above, an explanation of these concepts is unfortunately out of scope of this tutorial. There are many good Rust books and tutorials that can help you understand these crucial features.
```
jhdcs, line 149: Should we mention the relevant chapters of The Book as a place to start?

jhdcs (changes requested):
Just a few minor changes. However, it looks really good! I like the detail of it!

docs/writing-your-first-rclrs-node.md
```diff
@@ -177,7 +177,7 @@ fn republish(&self) -> Result<(), rclrs::RclrsError> {
 }
 ```
 
-What's left to do is to call this function every second. `rclrs` doesn't yet have built-in functionality to do run a function at a fixed interval, but it's easy enough to achieve with a thread and the sleep function. Change your main function to spawn a separate thread:
+What's left to do is to call this function every second. `rclrs` doesn't yet have ROS timers, which run a function at a fixed interval, but it's easy enough to achieve with a thread, a loop and the sleep function. Change your main function to spawn a separate thread:
```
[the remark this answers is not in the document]
nnmm: No problem, done.

docs/writing-your-first-rclrs-node.md
```diff
@@ -0,0 +1,226 @@
...
+
+```rust
+|msg: StringMsg| {
+    data = Some(msg);
+},
+```
+
+This is a standard pattern in C++, but doesn't work in Rust. Why not?
+
+Written like this, `data` is *borrowed* by the callback, but `data` is a local variable only exists in its current form until the end of `RepublisherNode::new()`. The subscription callback is required to not borrow any variables, because the subscription, and therefore the callback, can live indefinitely.
```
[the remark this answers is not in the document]
nnmm: Thanks, done.

docs/writing-your-first-rclrs-node.md
nnmm: I linked to the book, but don't want to link to individual chapters – people should just read the whole thing :)

jhdcs (approved):

esteve:
@nnmm awesome work! Thanks for writing this up 🙂

nnmm:
One possible way for avoiding all this `Arc<Mutex<_>>` stuff would be to make `rclrs::Node` have a field `data: Arc<Mutex<T>>`, and make subscriptions callbacks take an additional `&mut T` argument which is supplied by the node, which would basically play the role of `this` in member function callbacks in C++. I'm not sure yet it's worth doing, so let me know if you have any comments.

PR #176, opened 2022-05-23
Implement functions to get parameters from global arguments

nnmm:
The parameters are not accessible by users yet – I'll do that in a separate PR, since I want to keep them easy to review.

### DIFF
--- a/rclrs/src/parameter/override_map.rs
+++ b/rclrs/src/parameter/override_map.rs
@@ -1,0 +1,131 @@
+use crate::rcl_bindings::*;
+use crate::{ParameterValue, RclrsError, ToResult};
+use libc::c_char;
+use std::collections::BTreeMap;
+use std::ffi::CStr;
+
+// Internal helper struct, iterator for rcl_params_t
+struct RclParamsIter<'a> {
+    node_name_ptrs: &'a [*mut c_char],
+    rcl_node_params: &'a [rcl_node_params_t],
+}
+
+// Internal helper struct, iterator for rcl_node_params_t
+struct RclNodeParamsIter<'a> {
+    param_name_ptrs: &'a [*mut c_char],
+    rcl_variants: &'a [rcl_variant_t],
+}
+
+impl<'a> Iterator for RclParamsIter<'a> {
+    type Item = (String, &'a rcl_node_params_t);
+    fn next(&mut self) -> Option<Self::Item> {
+        let (next_node_param, rest) = self.rcl_node_params.split_first()?;
+        self.rcl_node_params = rest;
+        let (next_node_name_ptr, rest) = self.node_name_ptrs.split_first()?;
+        self.node_name_ptrs = rest;
+        debug_assert!(!next_node_name_ptr.is_null());
+        // SAFETY: Pointer can be assumed to be valid. No lifetime issues due to immediate
+        // conversion to owned string.
+        let mut next_node_name = unsafe {
+            CStr::from_ptr(*next_node_name_ptr)
+                .to_string_lossy()
+                .into_owned()
+        };
+        if !next_node_name.starts_with('/') {
+            next_node_name.insert(0, '/');
+        }
+        Some((next_node_name, next_node_param))
+    }
+}
+
+impl RclParamsIter<'_> {
+    // This function is unsafe since the rcl_params argument might contain incorrect array sizes or
+    // dangling pointers.
+    pub unsafe fn new(rcl_params: *const rcl_params_t) -> Self {
+        // The params will be null if none were given on the command line.
+        // So, semantically, this can be treated like a struct with no parameters.
+        if rcl_params.is_null() {
+            Self {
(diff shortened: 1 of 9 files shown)


rclrs/src/parameter/value.rs
```diff
@@ -0,0 +1,173 @@
...
+            !var.byte_array_value.is_null(),
+            !var.bool_array_value.is_null(),
+            !var.integer_array_value.is_null(),
+            !var.double_array_value.is_null(),
+            !var.string_array_value.is_null(),
+        ]
+        .into_iter()
+        .map(u8::from)
+        .sum();
+        assert_eq!(num_active, 1);
```
nnmm, line 71: This assertion will fail on empty list parameters, but neither of the big 2 client libraries can handle empty lists either.

rclrs/src/parameter/value.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// A string.
+    ///
+    /// YAML examples: `""`, `"björk"`, `"42"`.
+    ///
+    /// Unquoted strings are also possible, but not recommended,
+    /// because they may be interpreted as another data type.
+    String(String),
+    /// An array of u8.
+    ///
+    /// YAML example: Not possible to specify as YAML.
```
nnmm, line 31: Correct me if I'm wrong with this comment.

esteve: Do you mean bytes (i.e. binary data) or unsigned integers? The former needs to be encoded as base64, whereas the latter is not possible to enforce signed integers in YAML.

rclrs/src/node/mod.rs
```diff
@@ -316,3 +315,20 @@ impl Node {
         domain_id
     }
 }
+
+// Helper used to implement run_string_getter(), but also used to get the FQN in the Node::new()
+// function, which is why it's not merged into Node::run_string_getter().
+// This function is unsafe since it's possible to pass in an rcl_node_t with dangling
+// pointers etc.
+unsafe fn run_string_getter_with_handle(
```
esteve, line 323: Don't really know why, but the name seems odd. I'd name it `call_string_getter_with_handle`

rclrs/src/parameter/value.rs
esteve: No problem with failing on empty lists. We should get clarification from `rclcpp` and `rclpy` what the expected behavior should be.

esteve (changes requested):

rclrs/src/parameter/value.rs
nnmm: Yeah, I will try to find out more about this edge case. What I know is that launch_ros forbids empty lists because they're not supported (yet), so I think this is expected and we're definitely fine for now.

rclrs/src/parameter/value.rs
nnmm: Bytes. Can you specify how exactly to specify them? E.g. the following is base64-encoded, but parses to a string, not a byte array:

```
cargo run --bin minimal_subscriber -- --ros-args --param foo:="eW9sbw=="
```

rclrs/src/node/mod.rs
nnmm: Done.

nnmm:
@esteve This one could use another review too ;)

rclrs/src/parameter/override_map.rs
```diff
@@ -0,0 +1,134 @@
...
+            param_name_ptrs,
+            rcl_variants,
+        }
+    }
+}
+
+/// A map of parameters for this node.
+/// The key for this map is the parameter name.
+/// A parameter override "overrides" the default value set from within the node, hence the name.
+pub(crate) type ParameterOverrideMap = BTreeMap<String, ParameterValue>;
```
esteve, line 103: We can use a HashMap here since we don't need keys to be in any order

esteve (changes requested):

rclrs/src/parameter/value.rs
esteve: I meant that they can be encoded, but there's no way syntactically to differentiate bytes from strings.

rclrs/src/parameter/override_map.rs
nnmm: I intentionally used a BTreeMap since it will make it easy to fetch all parameters with a given prefix later.

nnmm:
I added a test to ensure that the precedence of `/**` vs named node and global vs node arguments is correct.

rclrs/src/parameter/override_map.rs
esteve: In that case, please add a comment describing the rationale for it

nnmm: Sure, done.

esteve: Thanks

esteve (approved):

PR #177, opened 2022-05-26
Add humble to CI

flynneva:
The setup-ros humble docker image is available now so we can add humble to the CI


esteve:
@flynneva looks like CI is failing because there's no `.repos` file for Humble, could you add one? Copying the one for rolling and updating the repos to point to humble should suffice. Thanks!

PR #182, opened 2022-05-29
Install cargo-ament-build with --debug

nnmm:
We already do this in CI, users should also not need an optimized cargo-ament-build

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Here are the steps for building the `ros2_rust` examples in a vanilla Ubuntu Foc
 # Assuming you installed the minimal version of ROS 2, you need these additional packages:
 sudo apt install -y git libclang-dev python3-pip python3-vcstool # libclang-dev is required by bindgen
 # Install these plugins for cargo and colcon:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,7 +35,7 @@ Here are the steps for building the `ros2_rust` examples in a vanilla Ubuntu Foc
 # Assuming you installed the minimal version of ROS 2, you need these additional packages:
 sudo apt install -y git libclang-dev python3-pip python3-vcstool # libclang-dev is required by bindgen
 # Install these plugins for cargo and colcon:
-cargo install cargo-ament-build
+cargo install --debug cargo-ament-build  # --debug is faster to install
```
esteve, line 38: Why is it faster?

esteve (approved):

README.md
nnmm: By default, `cargo install` compiles with optimizations enabled, which takes much longer. The time spent in cargo-ament-build itself is minimal, so I think having a snappier setup is worth not having these optimizations.

PR #186, opened 2022-05-31
Add methods related to node options to NodeBuilder

### LINKED ISSUE #142: Add support for node options

Soya-Onishi:
Closes #142 

This PR add node options related to [`rcl_node_options_t`](https://docs.ros2.org/latest/api/rcl/structrcl__node__options__t.html).
However, methods related to allocator options (e.g. NodeBuilder::allocators.allocate) are not implemented because there are no way to return allocated heap pointer as far as I know.


rclrs/src/node/builder.rs
```diff
@@ -159,11 +274,38 @@ impl NodeBuilder {
...
+                &mut arguments,
+            )
+        }
+        .ok()?;
+
+        node_options.domain_id = self.domain_id;
+        node_options.arguments = arguments;
+        node_options.use_global_arguments = self.use_global_arguments;
+        node_options.enable_rosout = self.enable_rosout;
+        //SAFETY: No preconditions for this function.
```
nnmm, line 303:
```suggestion
        // SAFETY: No preconditions for this function.
```

rclrs/src/node/builder.rs
```diff
@@ -159,11 +274,38 @@ impl NodeBuilder {
         // SAFETY: No preconditions for this function.
         let mut node_handle = unsafe { rcl_get_zero_initialized_node() };
```
nnmm, line 275: Can you move this line further down, close to `rcl_node_init()` where it is used?

rclrs/src/node/builder.rs
```diff
@@ -40,6 +44,10 @@ pub struct NodeBuilder {
     context: Arc<Mutex<rcl_context_t>>,
     name: String,
     namespace: String,
+    domain_id: usize,
```
nnmm, line 47: The domain ID isn't part of the node_options anymore from galactic onwards. I think it now is an attribute of the context, not the node: https://github.com/ros2/rcl/pull/708

So, probably better to remove this. Sorry, I think I explicitly suggested this on the issue!

nnmm (commented):
Thanks! I'll add more comments later.

rclrs/src/node/builder.rs
nnmm: Or, instead of removing, you can gate it with `#[cfg(ros_distro = "foxy")]`.

rclrs/src/node/builder.rs
```diff
@@ -40,6 +44,10 @@ pub struct NodeBuilder {
...
+    use_global_arguments: bool,
+    arguments: Vec<String>,
+    enable_rosout: bool,
```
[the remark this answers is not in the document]
Soya-Onishi: Thanks for your suggestions! I have one question. If making fields public, should context filed also be public? Current context field has raw type of rcl and should not expose those types against users as much as possible I think. If it should make context field private, it is better for other fields to make private because of the rule, `Either make all fields of a type public, or make none of them public`, from blog post that is linked from reddit you attached.  Another way to make all field public is usage of reference for Context, but this may make compiler more sensitive and cause lifetime errors which is hard to understand why, although I haven't inspect yet.

By the way, surely, current builder setters are smell ones if I do not misunderstand the means about smell( I read posts as smell setter is equal to trivial setter, and trivial setter is the setter that takes a parameter and only assigns to struct field). However I could not understand why setter is worse when there are smell setters. If there are comments says why smell setters are bad, please tell me.

Also, the post you attached does not says about assigning a value to public field as far as I know. I read that post as recommendation of using C-like struct instantiation pattern with Default when the struct just be data box. However, on another note, this pattern is possible only if there are only optional values (there is also same questionable comment about that but no response), and NodeBuilder has essential fields (i.e. `context` and `name`), so, in this case, I cannot apply this pattern to NodeBuilder.

rclrs/src/node/builder.rs
Soya-Onishi: Thank you for letting me know about `domain_id`.
Please tell me where the description or documentation about `domain_id` is in context now.

rclrs/src/node/builder.rs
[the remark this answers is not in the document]
nnmm: Sorry, I guess that link is a bit confusing without context. Let me try to clarify
- I agree, the context should definitely be private.
- The "trivial setters" matklad calls "smelly" are those in the parent comment: `shape.set_position(Vec3(0, 9, 2));` etc. – they have a signature like `fn set_position(&mut self, position: Vec3)`.
  - This signature is different from the chainable setters in `NodeBuilder`, which look more like `fn position(self, position: Vec3) -> Self`. Chainable setters are useful and I think no one sees them as a code smell. So I think there's nothing wrong at all with the current implementation of `NodeBuilder`.
- So, the point of linking to that comment was just that public fields are not always bad. The "builder lite" blog post it's attached to isn't really relevant here. Sorry for the confusion.
- Further down in the reddit thread, people agree that public fields are fine unless (1) there are some invariants that need to be protected (we don't have that, since validation happens later) or (2) you need to be careful about making breaking changes to the struct (which is also not a concern, since we can make breaking changes, and since the private context field will make the changes non-breaking anyway).
- Yep, https://matklad.github.io/2022/05/29/binary-privacy.html recommends categorizing everything into data or abstract data types. Not sure which one a builder is supposed to be. I think it's closer to data, since we could also remove the context field and only pass it in in the `build()` method, then everything could be public. But in the end it's only a rule of thumb, just like "avoid trivial setters".

It's a really small change we're discussing here, so I'll leave it up to you :) I just find it nice to see the fields in the docs and see that there is nothing weird going on, and to be able to also use `builder.enable_rosout = true;` syntax.

rclrs/src/node/builder.rs
nnmm: In newer distros it's part of the `init_options` that are passed to `rcl_init()` (called in the `Context::new` method).

nnmm: Probably it's best to leave the `domain_id` for a separate PR.

nnmm:
@Soya-Onishi Great work! Could you also add `Drop` instances for `rcl_node_options_t` and `rcl_arguments_t` that call their `fini` functions?

rclrs/src/node/builder.rs
```diff
@@ -187,4 +266,37 @@ impl NodeBuilder {
...
+        let cstring_args = self
+            .arguments
+            .iter()
+            .map(|s| CString::new(s.as_str()).unwrap())
+            .collect::<Vec<_>>();
+        let cstring_arg_ptrs = cstring_args.iter().map(|s| s.as_ptr()).collect::<Vec<_>>();
+        // SAFETY: Getting a zero-initialized value is always safe.
+        let mut arguments = unsafe { rcl_get_zero_initialized_arguments() };
+        unsafe {
+            // SAFETY: This function does not store the ephemeral cstr_args_ptrs
```
nnmm, line 283:
```suggestion
            // SAFETY: This function does not store the ephemeral cstring_args_ptrs
```

rclrs/src/node/builder.rs
```diff
@@ -13,6 +13,10 @@ use std::sync::Arc;
 ///
 /// The default values for optional fields are:
 /// - `namespace: "/"`
+/// - `domain_id: ROS_DOMAIN_ID`
```
nnmm, line 16:
```suggestion
```

rclrs/src/node/builder.rs
```diff
@@ -13,6 +13,10 @@ use std::sync::Arc;
...
+/// - `use_global_arguments: true`
+/// - `arguments: []`
+/// - `enable_rosout`: true`
```
nnmm, line 19:
```suggestion
/// - `enable_rosout: true`
```

rclrs/src/node/builder.rs
```diff
@@ -142,28 +152,97 @@ impl NodeBuilder {
...
+    ///   .iter()
+    ///   .map(|s| s.to_string())
+    ///   .collect::<Vec<_>>();
+    /// let node = context
+    ///   .create_node_builder("my_node")
+    ///   .arguments(node_args)
+    ///   .build()?;
+    /// assert_eq!(node.name(), "your_node");
+    /// # Ok::<(), RclrsError>(())
+    /// ```
```
nnmm, line 210: I think it's good to just link to `Context::new()` (which refers to the official ROS 2 article), to have one source of truth for how command line arguments for ROS 2 work.

```suggestion
    /// Sets node-specific command line arguments.
    ///
    /// These arguments are parsed the same way as those for [`Context::new()`][1].
    /// However, the node-specific command line arguments have higher precedence than the arguments
    /// used in creating the context.
    ///    
    /// # Example
    /// ```
    /// # use rclrs::{Context, Node, NodeBuilder, RclrsError};
    /// // Usually, this would change the name of "my_node" to "context_args_node":
    /// let context_args = ["--ros-args", "--remap", "my_node:__node:=context_args_node"]
    ///   .map(String::from);
    /// let context = Context::new(context_args)?;
    /// // But the node arguments will change it to "node_args_node":
    /// let node_args = ["--ros-args", "--remap", "my_node:__node:=node_args_node"]
    ///   .map(String::from);
    /// let node = context
    ///   .create_node_builder("my_node")
    ///   .arguments(node_args)
    ///   .build()?;
    /// assert_eq!(node.name(), "node_args_node");
    /// # Ok::<(), RclrsError>(())
    /// ```
    ///
    /// [1]: crate::Context::new
```

rclrs/src/node/builder.rs
```diff
@@ -142,28 +152,97 @@ impl NodeBuilder {
...
+    pub fn arguments(mut self, arguments: impl IntoIterator<Item = String>) -> Self {
+        self.arguments = arguments.into_iter().collect();
+        self
+    }
+
+    /// Enables rosout.
+    /// If false, rosout logging is disabled.
+    pub fn enable_rosout(mut self, enable: bool) -> Self {
```
nnmm, line 218: Do you know if this actually affects logs emitted by the `rcl` layer? Or does it currently not have any effect at all?

rclrs/src/node/builder.rs
```diff
@@ -142,28 +152,97 @@ impl NodeBuilder {
...
+    ///   .build()?;
+    /// assert_eq!(node.name(), "your_node");
+    ///
+    /// // Not use global arguments
+    /// let node = context
+    ///   .create_node_builder("my_node")
+    ///   .use_global_arguments(false)
+    ///   .build()?;
+    /// assert_eq!(node.name(), "my_node");
+    /// # Ok::<(), RclrsError>(())
```
nnmm, line 180:
```suggestion
    /// Enables or disables using global arguments.
    ///
    /// The "global" arguments are those used in [creating the context][1].
    ///
    /// # Example
    /// ```
    /// # use rclrs::{Context, Node, NodeBuilder, RclrsError};
    /// let context_args = ["--ros-args", "--remap", "__node:=your_node"]
    ///   .map(String::from);
    /// let context = Context::new(context_args)?;
    /// // Ignore the global arguments:
    /// let node_without_global_args = context
    ///   .create_node_builder("my_node")
    ///   .use_global_arguments(false)
    ///   .build()?;
    /// assert_eq!(node_without_global_args.name(), "my_node");
    /// // Do not ignore the global arguments:
    /// let node_with_global_args = context
    ///   .create_node_builder("my_other_node")
    ///   .use_global_arguments(true)
    ///   .build()?;
    /// assert_eq!(node_with_global_args.name(), "your_node");
    /// # Ok::<(), RclrsError>(())
    /// ```
    ///
    /// [1]: crate::Context::new
```

nnmm:
I tested it out and got an error about `rcl_arguments_fini` being called twice (which makes sense, because `rcl_node_options_fini` also calls `rcl_arguments_fini`). So we should probably remove the `Drop` impl for `rcl_arguments_t` again, that would be easiest.

Apart from that, I only have one other comment (about enable_rosout), then it's ready to merge!

rclrs/src/node/builder.rs
```diff
@@ -142,28 +151,100 @@ impl NodeBuilder {
...
+    ///
+    /// [1]: crate::Context::new
+    /// [2]: https://design.ros2.org/articles/ros_command_line_arguments.html
+    pub fn arguments(mut self, arguments: impl IntoIterator<Item = String>) -> Self {
+        self.arguments = arguments.into_iter().collect();
+        self
+    }
+
+    /// Enables rosout.
+    /// If false, rosout logging is disabled.
```
nnmm, line 219:
```suggestion
    /// Enables or disables logging to rosout.
    ///
    /// When enabled, log messages are published to the `/rosout` topic in addition to
    /// standard output.
    ///
    /// This option is currently unused in `rclrs`.
```

rclrs/src/node/builder.rs
nnmm: I believe it currently has no effect.

rclrs/src/node/builder.rs
```diff
@@ -142,28 +151,100 @@ impl NodeBuilder {
...
     /// Node name and namespace validation is performed in this method.
     ///
     /// For example usage, see the [`NodeBuilder`][1] docs.
     ///
     /// # Panics
-    /// When the node name or namespace contain null bytes.
+    /// If any of below conditions are filled, panic occurs
+    ///  - node name contains null byte
+    ///  - namespace contains null byte
+    ///  - any of node argument strings contain null byte.
```
nnmm, line 235:
```suggestion
    /// If any of the below conditions are fulfilled, a panic occurs:
    ///  - The node name contains a null byte
    ///  - The namespace contains a null byte
    ///  - Any of the node arguments contain a null byte
```

rclrs/src/node/builder.rs
Soya-Onishi: As for now, there is no effect in ros2_rust.
However, as far as I know, this flag affects whether logging features publish to `rosout` topic or not.
For the future, this should retain and put comments to tell this is nothing to do for now, I think.

rclrs/src/node/builder.rs
Soya-Onishi: Thanks for adding more detailed comments!!
As I said above, this will work after implementing logging features in rclrs I think.

rclrs/src/node/builder.rs
```diff
@@ -187,4 +269,59 @@ impl NodeBuilder {
...
+        // SAFETY: No preconditions for this function.
+        let mut node_options = unsafe { rcl_node_get_default_options() };
+
+        let cstring_args = self
+            .arguments
+            .iter()
+            .map(|s| CString::new(s.as_str()).unwrap())
+            .collect::<Vec<_>>();
+        let cstring_arg_ptrs = cstring_args.iter().map(|s| s.as_ptr()).collect::<Vec<_>>();
+        // SAFETY: Getting a zero-initialized value is always safe.
```
nnmm, line 285: This line is outdated and can be removed.

rclrs/src/node/builder.rs
```diff
@@ -187,4 +269,59 @@ impl NodeBuilder {
...
+        node_options.use_global_arguments = self.use_global_arguments;
+        node_options.enable_rosout = self.enable_rosout;
+        // SAFETY: No preconditions for this function.
+        node_options.allocator = unsafe { rcutils_get_default_allocator() };
+
+        Ok(node_options)
+    }
+}
+
+impl Drop for rcl_arguments_t {
```
nnmm, line 307: I think it would be slightly cleaner to just remove this impl since it will always be freed as part of `rcl_node_options_fini`. But if you disagree, feel free to leave it in.

esteve: I can't think of an example where `rcl_arguments_t` would be used directly and not in combination with `rcl_node_options_t`, except for parsing command line arguments, but we'd use Rust's native facilities in that case. I can understand for the sake of completeness, it's up to you @Soya-Onishi if you want to keep this code, I'm leaning towards removing it.

esteve (changes requested):

nnmm:
@Soya-Onishi Could you adjust the CString operations to match https://github.com/ros2-rust/ros2_rust/pull/190, which was merged in the meantime?

rclrs/src/node/builder.rs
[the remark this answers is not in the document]
nnmm: Feel free to resolve this, it was just a suggestion and like I said I'll leave it up to you.

nnmm:
@Soya-Onishi I have another PR that conflicts with this one, https://github.com/ros2-rust/ros2_rust/pull/193. If you plan to make the couple of remaining fixes this week, I can wait with my PR until we merge this one, so you don't have extra work.

rclrs/src/node/builder.rs
Soya-Onishi: @esteve Thanks for your advice! I don't know whether it is related to what @nnmm says or not, but should I put statements about `rcl_arguments_t` is finished in `rcl_node_options_fini()`?

nnmm: Yes, we're both saying the same :) You could add a comment `// This also finalizes the rcl_arguments contained in the node options`  to the `Drop for rcl_node_options_t`, and then just delete the `Drop for rcl_arguments_t`.

rclrs/src/node/builder.rs
Soya-Onishi: Thanks for your pick! I'll fix it.

rclrs/src/node/builder.rs
```diff
@@ -192,4 +271,47 @@ impl NodeBuilder {
...
+
+        Ok(node_options)
+    }
+}
+
+impl Drop for rcl_node_options_t {
+    fn drop(&mut self) {
+        // SAFETY: Do not finish this struct except here.
+        unsafe {
+            // This also finalizes the `rcl_arguments_t` contained in `rcl_node_options_t`,
```
nnmm, line 313:
```suggestion
            // This also finalizes the `rcl_arguments_t` contained in `rcl_node_options_t`.
```

Super nitpicky, I know :)

rclrs/src/node/builder.rs
```diff
@@ -192,4 +271,47 @@ impl NodeBuilder {
             subscriptions: std::vec![],
         })
     }
+
+    /// Creates node options.
+    /// options are overrided by field values of builder.
```
nnmm, line 276:
```suggestion
    ///
    /// Any fields not present in the builder will have their default value.
```

esteve (dismissed):
@Soya-Onishi thanks for all the hard work and for the patience iterating with us! :slightly_smiling_face:

rclrs/src/node/builder.rs
```diff
@@ -273,15 +273,31 @@ impl NodeBuilder {
...
-            .map(|s| CString::new(s.as_str()).unwrap())
+            .map(|s| match CString::new(s.as_str()) {
+                Ok(cstr) => Ok(cstr),
+                Err(err) => Err(RclrsError::StringContainsNul { s: s.clone(), err }),
+            })
+            .partition(Result::is_ok);
+
+        if let Some(err) = parse_err_results.pop() {
+            return Err(err.unwrap_err());
+        }
```
nnmm, line 296: Did you know you can collect into a `Result<Vec<_>>`? That would be better here, I think.

Soya-Onishi: Thanks for your notice!

rclrs/src/node/builder.rs
nnmm: I just removed it since it was the only remaining fix, hope that's okay!

nnmm (approved):

nnmm:
Wonderful! Thanks @Soya-Onishi!

Soya-Onishi:
@nnmm @esteve 
Thanks for your many advices and picks!

nnmm:
Feel free to join the Matrix chat by the way.

PR #187, opened 2022-06-03
feat: add example launch

wep21:
I added minimal_pub_sub.launch.xml into example package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ ros2 run examples_rclrs_minimal_pub_sub minimal_subscriber
 ```
 or
 
-```
+```shell
 # In a new terminal (or tmux window)
 . ./install/setup.sh
 ros2 launch examples_rclrs_minimal_pub_sub minimal_pub_sub.launch.xml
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -56,6 +56,13 @@ ros2 run examples_rclrs_minimal_pub_sub minimal_publisher
 . ./install/setup.sh
 ros2 run examples_rclrs_minimal_pub_sub minimal_subscriber
 ```
+or
+
+```
```
nnmm, line 61:
```suggestion
```shell
```

nnmm:
Nice! Just one nitpick :)

nnmm:
Although – don't most people use Python launch files nowadays? Do you know if they are always preferred @esteve @tylerjw @jhdcs?

tylerjw:
We use only python launch files but it isn't because we think we need more imperative launch files. We all much prefer the xml launch system from ros1, I just don't think we've learned to write ros2 xml files.

nnmm (approved):
Yep, I agree, the XML version is nice and short. We could always additionally provide Python or YAML launch files if we want to.

I'll wait a couple of hours until merging though, to give others the chance to leave a comment if they want to.

esteve:
In general I prefer the Python, but for a simple case like this, XML is much cleaner. Thanks @wep21 !

esteve (approved):

PR #190, opened 2022-06-04
Make rclrs panic-free as far as possible

nnmm:
This is achieved by making CString errors a part of `RclrsError`. This was made possible by #183, which made the error struct an enum. 

Example error output:
 ```
Error: Could not convert string 'minimal_subscriber' to CString

Caused by:
    nul byte found in provided data at position: 13
```

The nul byte is not rendered in the output, as you can see.

### DIFF
--- a/rclrs/src/context.rs
+++ b/rclrs/src/context.rs
@@ -59,16 +59,18 @@ impl Context {
     /// let invalid_remapping = ["--ros-args", "-r", ":=:*/]"].map(String::from);
     /// assert!(Context::new(invalid_remapping).is_err());
     /// ```
-    ///
-    /// # Panics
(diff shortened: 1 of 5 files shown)


nnmm:
Note for reviewers: This does not increase the size of the `RclrsError` struct, since the other variants are larger.

The `map_err()` is of course a little verbose, but users should never need to do `CString` conversions anyway, so the API is clean. But still, if it's too ugly, we could remove the string member from this enum variant and `impl From<NulError> for RclrsError`.

jhdcs (approved):
These look to be fairly simple and reasonable changes to me.

nnmm:
Thanks @jhdcs!

Issue #148, opened 2022-05-02
Add functions to get topic name to publisher and subscription


nnmm:
The `rcl_subscription_get_topic_name` and `rcl_publisher_get_topic_name` functions should be wrapped and added to the `Subscription` and `Publisher`, analogously to https://github.com/ros2-rust/ros2_rust/pull/143.

surfertas:
Seems like this was closed by #194 . Am looking for issues to work on.

nnmm:
That's true. How about https://github.com/ros2-rust/ros2_rust/issues/147?

PR #194, opened 2022-06-09
Add functions to get topic name to publisher and subscriber

DS3a:
A new variable was added to the struct to store the topic name in a boxed string.

### DIFF
--- a/rclrs/src/node/publisher.rs
+++ b/rclrs/src/node/publisher.rs
@@ -5,7 +5,6 @@ use crate::Node;
 
 use std::borrow::Cow;
 use std::ffi::CString;
-use std::boxed::Box;
 use std::marker::PhantomData;
 use std::sync::Arc;
 
@@ -55,7 +54,6 @@ where
 {
     pub(crate) handle: Arc<PublisherHandle>,
     message: PhantomData<T>,
-    topic_name: Box<String>,
 }
 
 impl<T> Publisher<T>
@@ -106,13 +104,14 @@ where
         Ok(Self {
             handle,
             message: PhantomData,
-            topic_name: Box::new(topic.to_string())
         })
     }
 
     /// Returns the topic that the publisher is publishing data to
-    pub fn rcl_publisher_get_topic_name(&self) -> Result<String, RclrsError> {
-        return Ok(*self.topic_name.clone());
+    pub fn get_topic(&self) {
+        // SAFETY: No preconditions for this function.
+        let mut raw_topic = unsafe{ rcl_publisher_get_topic_name((&mut *self.handle.lock()) as *mut _) };
+        println!("The topic name is : {:?}", &raw_topic);
     }
 
     /// Publishes a message.
--- a/rclrs/src/node/publisher.rs
+++ b/rclrs/src/node/publisher.rs
@@ -4,8 +4,8 @@ use crate::rcl_bindings::*;
 use crate::Node;
 
 use std::borrow::Cow;
-use std::ffi::CString;
 use std::ffi::CStr;
+use std::ffi::CString;
 use std::marker::PhantomData;
 use std::sync::Arc;
 
@@ -108,28 +108,19 @@ where
         })
     }
 
-    /// Returns the topic that the publisher is publishing data to
-    pub fn get_topic(&self) -> Result<String, RclrsError> {
-
-        let mut _topic_name: String = String::default();
+    /// Returns the topic name of the publisher.
+    ///
+    /// This returns the topic name after remapping, so it is not necessarily the
+    /// topic name which was used when creating the publisher
+    pub fn get_topic(&self) -> Result<Cow<str>, RclrsError> {
         // SAFETY: No preconditions for the functions called.
-        // Edge cases for all unsafe variables created have been handled
-        // The unsafe variables created get dropped as soon as they're used
+        // The unsafe variables created get converted to safe types before being returned
         unsafe {
-            let raw_topic_pointer = rcl_publisher_get_topic_name((&mut *self.handle.lock()) as *mut _);
-            match CStr::from_ptr(raw_topic_pointer).to_str() {
-                Ok(name) => { _topic_name = name.to_string(); },
-                Err(_) => {
-                    return Err(RclrsError::RclError{
-                        code: crate::error::RclReturnCode::TopicNameInvalid,
-                        msg: None,
-                        // unable to return an error message as the String field in RclErrorMsg is private
-                    });
-                },
-            }
+            let raw_topic_pointer = rcl_publisher_get_topic_name(&*self.handle.lock());
+            return Ok(CStr::from_ptr(raw_topic_pointer)
+                .to_string_lossy()
+                .to_owned());
         }
-
-        return Ok(_topic_name);
     }
 
     /// Publishes a message.
--- a/rclrs/src/node/publisher.rs
+++ b/rclrs/src/node/publisher.rs
@@ -5,6 +5,7 @@ use crate::Node;
 
 use std::borrow::Cow;
 use std::ffi::CString;
+use std::boxed::Box;
 use std::marker::PhantomData;
 use std::sync::Arc;
 
@@ -54,6 +55,7 @@ where
 {
     pub(crate) handle: Arc<PublisherHandle>,
     message: PhantomData<T>,
+    topic_name: Box<String>,
 }
 
 impl<T> Publisher<T>
@@ -104,7 +106,13 @@ where
         Ok(Self {
             handle,
             message: PhantomData,
+            topic_name: Box::new(topic.to_string())
         })
+    }
+
+    /// Returns the topic that the publisher is publishing data to
+    pub fn rcl_publisher_get_topic_name(&self) -> Result<String, RclrsError> {
+        return Ok(*self.topic_name.clone());
     }
 
     /// Publishes a message.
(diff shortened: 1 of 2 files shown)


nnmm:
Welcome to the ros2-rust project @DS3a!

This implementation is probably based on a misunderstanding of https://github.com/ros2-rust/ros2_rust/issues/148: The functions `rcl_subscription_get_topic_name` and `rcl_publisher_get_topic_name` already exist, in the `rcl_bindings` module (they come from the `rcl` package). You should call them to fetch the string. The reason is that these functions return the topic name _after remapping_, and not just the topic name that was passed in. See https://github.com/ros2-rust/ros2_rust/pull/143 for more context, and for an example of how to deal with C strings.

nnmm:
Is it a requirement @jhdcs?

jhdcs:
Now that I look at it again, it does not appear to be for us... However, I believe that other ROS 2 projects (such as the client libraries) require it as part of the DCO check.

DS3a:
> Additionally, it looks like your commits were not able to be verified. Would you be able to look into that? That will need to be fixed before we can merge.

Aight... I'll look into that.

DS3a:
> Welcome to the ros2-rust project @DS3a!
> 
> This implementation is probably based on a misunderstanding of https://github.com/ros2-rust/ros2_rust/issues/148: The functions `rcl_subscription_get_topic_name` and `rcl_publisher_get_topic_name` already exist, in the `rcl_bindings` module (they come from the `rcl` package). You should call them to fetch the string. The reason is that these functions return the topic name _after remapping_, and not just the topic name that was passed in. See https://github.com/ros2-rust/ros2_rust/pull/143 for more context, and for an example of how to deal with C strings.

Ahhh... My bad... I'll change it accordingly

rclrs/src/node/subscription.rs
```diff
@@ -125,6 +126,28 @@ where
...
+    /// Returns the name of the topic which is being subscribed
+    pub fn get_topic(&self) -> Result<String, RclrsError> {
+
+        let mut _topic_name: String = String::default();
+        // SAFETY: No preconditions for the function used
+        // Edge cases for all unsafe variables have been handled
+        // The unsafe variables created get dropped within the unsafe scope as soon as they're used
+        unsafe {
+            let raw_topic_pointer = rcl_subscription_get_topic_name((&mut *self.handle.lock()) as *mut _);
+            if let Ok(name) =  CStr::from_ptr(raw_topic_pointer).to_str() {
```
esteve, line 138: Mostly a style preference, but I like the match syntax more than if..else.. for cases like this:

```rust
match CStr::from_ptr(raw_topic_pointer).to_str() {
  Ok(name) => name.to_string(),
  Err(_) => Err(RclsError::RclError{code: crate::error::RclReturnCode::TopicNameInvalid,
    msg: None,
    // Unable to give an error message as the String field in RclErrorMsg is private
  },
}
```

esteve (changes requested):

DS3a (commented):
A self review has been done... This is my first PR, forgive me for having to be reminded of the protocol

rclrs/src/node/publisher.rs
```diff
@@ -107,6 +108,30 @@ where
...
+    /// Returns the topic that the publisher is publishing data to
+    pub fn get_topic(&self) -> Result<String, RclrsError> {
+
+        let mut _topic_name: String = String::default();
+        // SAFETY: No preconditions for the functions called.
+        // Edge cases for all unsafe variables created have been handled
+        // The unsafe variables created get dropped as soon as they're used
+        unsafe {
+            let raw_topic_pointer = rcl_publisher_get_topic_name((&mut *self.handle.lock()) as *mut _);
+            match CStr::from_ptr(raw_topic_pointer).to_str() {
```
nnmm, line 120: In other places I've just used `to_string_lossy().into_owned()`, which also has the advantage that the function can just return a `String`. Would be nice to do that here too for consistency.

rclrs/src/node/publisher.rs
```diff
@@ -107,6 +108,30 @@ where
...
 
+    /// Returns the topic that the publisher is publishing data to
+    pub fn get_topic(&self) -> Result<String, RclrsError> {
+
+        let mut _topic_name: String = String::default();
+        // SAFETY: No preconditions for the functions called.
+        // Edge cases for all unsafe variables created have been handled
+        // The unsafe variables created get dropped as soon as they're used
+        unsafe {
+            let raw_topic_pointer = rcl_publisher_get_topic_name((&mut *self.handle.lock()) as *mut _);
```
nnmm, line 119: The `as *mut _` can be omitted, see https://github.com/ros2-rust/ros2_rust/pull/134.

rclrs/src/node/publisher.rs
```diff
@@ -107,6 +108,30 @@ where
         })
     }
 
+    /// Returns the topic that the publisher is publishing data to
```
nnmm, line 111: Could you add documentation in the same style as [the one for `Node::name` etc](https://github.com/ros2-rust/ros2_rust/blob/30b193152e27c09cf9101265d94cc10358edd58e/rclrs/src/node.rs#L122-L143)? E.g.

```suggestion
    /// Returns the topic name of the publisher.
    ///
    /// This returns the topic name after remapping, so it is not necessarily the same as the
    /// topic name that was used when creating the publisher.
```

It's okay to not have example code, though (because we'd need an extra dependency on a message package, which we don't want).

nnmm (changes requested):
Thanks for adjusting, the PR looks good in general now, though I have a couple of comments.

As you can see, our CI checks formatting and clippy, so make sure to run `cargo fmt` and `cargo clippy` as well.

rclrs/src/node/publisher.rs
nnmm: These comments of course also apply to the subscription.

rclrs/src/node/publisher.rs
DS3a: noted, these changes have been made to add consistency

rclrs/src/node/publisher.rs
DS3a: ohh, okay, the declaration had this type so I just assumed it was required... I've made the necessary changes

rclrs/src/node/publisher.rs
DS3a: The comments have been formatted as requested... Please do let me know if there are any further changes required

jhdcs:
> A self review has been done... This is my first PR, forgive me for having to be reminded of the protocol

No worries! We all start somewhere! Thank _you_ for taking the time to try and make `ros2_rust` a little better! We really appreciate it!

rclrs/src/node/subscription.rs
```diff
@@ -125,6 +126,21 @@ where
...
+    pub fn get_topic(&self) -> Result<Cow<str>, RclrsError> {
+        // SAFETY: No preconditions for the function used
+        // The unsafe variables get converted to safe types before being returned
+        unsafe {
+            let raw_topic_pointer = rcl_subscription_get_topic_name(&*self.handle.lock());
+            return Ok(CStr::from_ptr(raw_topic_pointer)
+                .to_string_lossy()
+                .to_owned());
+        }
+    }
```
nnmm, line 142:
```suggestion
    pub fn topic_name(&self) -> String {
        // SAFETY: No preconditions for the function used
        // The unsafe variables get converted to safe types before being returned
        unsafe {
            let raw_topic_pointer = rcl_subscription_get_topic_name(&*self.handle.lock());
            CStr::from_ptr(raw_topic_pointer)
                .to_string_lossy()
                .into_owned()
        }
    }
```

nnmm: Same for the publisher. Rust doesn't need `return` at the end of the function, and `into_owned()` returns a String.

DS3a: I've removed the return statement... But The compiler is asking me to add `to_string()` after `into_owned()`... So I've added that in as well... Please let me know if I should have done it differently

nnmm:
@DS3a I noticed you did something different than I suggested. Did the suggestion not work?

nnmm:
@DS3a Are you sure you tried with `into_owned()` instead of `to_owned()`? They are different methods.

rclrs/src/node/subscription.rs
nnmm: Only one thing is missing: please remove the `Result` and `Ok(…)`, since they're not needed.

DS3a:
okay... I've made changes to return `String` instead of `Result<String, Err>`

rclrs/src/node/subscription.rs
```diff
@@ -125,6 +126,21 @@ where
         })
     }
 
+    /// Returns the topic name of the subscription.
+    ///
+    /// This returns the topic name after remapping, so it is not necessarily the
+    /// topic name which was used when creating the subscription
```
nnmm, line 132:
```suggestion
    /// topic name which was used when creating the subscription.
```

rclrs/src/node/publisher.rs
```diff
@@ -107,6 +108,21 @@ where
         })
     }
 
+    /// Returns the topic name of the publisher.
+    ///
+    /// This returns the topic name after remapping, so it is not necessarily the
+    /// topic name which was used when creating the publisher
```
nnmm, line 114:
```suggestion
    /// topic name which was used when creating the publisher.
```

nnmm (approved):
Looks good to me. Thanks for your patience @DS3a!

DS3a:
> Looks good to me. Thanks for your patience @DS3a!

Thanks... I look forward to contributing more to ros2_rust!!

esteve (approved):

esteve:
@DS3a thanks for you contribution! :slightly_smiling_face:

PR #200, opened 2022-06-13
Move create_node and create_node_builder back to the rclrs module

### LINKED ISSUE #161: Remove `create_node` from `Context`
Neither the two official client libraries (`rclpy` and `rclcpp`) have a `create_node` method in their `Context` class. `rclcpy` has a `create_node` function at the module level, and `rclcpp` got rid of such function a while ago.

With both libraries when creating a node, they create a default context that gets assigned to the node if none is passed. With a builder pattern API we can achieve the same if needed. The goal of this ticket is to either move `create_node` back to the `rclrs` module or remove it altogether.

esteve:
Closes #161

### DIFF
--- a/rclrs/src/node.rs
+++ b/rclrs/src/node.rs
@@ -290,7 +290,7 @@ unsafe fn call_string_getter_with_handle(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{Context, Node, QOS_PROFILE_DEFAULT, create_node};
+    use crate::{create_node, Context, Node, QOS_PROFILE_DEFAULT};
 
     #[test]
     fn test_new() -> Result<(), RclrsError> {
--- a/rclrs/src/node/publisher.rs
+++ b/rclrs/src/node/publisher.rs
@@ -167,13 +167,13 @@ impl<'a, T: Message> MessageCow<'a, T> for &'a T {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{Context, Publisher, QOS_PROFILE_DEFAULT, create_node};
+    use crate::{create_node, Context, Publisher, QOS_PROFILE_DEFAULT};
 
     fn create_fixture(name: &str) -> (Context, Node) {
         let context =
             Context::new(vec![]).expect("Context instantiation is expected to be a success");
-        let node = create_node(&context, name)
-            .expect("Node instantiation is expected to be a success");
+        let node =
+            create_node(&context, name).expect("Node instantiation is expected to be a success");
 
         (context, node)
     }
--- a/rclrs/src/node/subscription.rs
+++ b/rclrs/src/node/subscription.rs
@@ -210,7 +210,7 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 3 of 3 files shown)


jhdcs (dismissed):

nnmm (dismissed):
Why do we not want it as a method of context?

nnmm:
Yes, I almost never use "changes requested" for this reason, but in this case the change doesn't seem beneficial to me, so I intentionally wanted to block it until we've had a chance to reach an agreement.

jhdcs:
I thought this had been discussed earlier? If I recall correctly, neither `rclcpp` nor `rclpy` have a `context::create_node` function anymore. This change was to bring us to a similar state. Right?

esteve:
> Yes, I almost never use "changes requested" for this reason, but in this case the change doesn't seem beneficial to me, so I intentionally wanted to block it until we've had a chance to reach an agreement.

We did discuss this, as @jhdcs pointed out. And there's an agreement, two of us agree on this change. We've had instances were two people agreed on something and one was not fully on board (e.g. I was the one who didn't agree), but we as a project went with the change, because there's really no other way than going with what the majority decides.

> I'm not opposed to having both, by the way.

Why both? There should be one clear API to create nodes. And it should be the one that resembles the other client libraries the most, because that's what people learn.

>Although I don't understand why `rclpy` or `rclrs` should need the `create_node` function when they have `Node` classes/structs they could just instantiate.

`rclcpp` does not have a `create_node` function any more, the way of creating a `Node` is via inheritance. There are other more intricate ways of creating a `Node` in `rclcpp`, but for the sake of simplicity, let's say that inheritance is the only way.

> Seems like you added it @esteve, so maybe you know :) [ros2/rclpy@c311569](https://github.com/ros2/rclpy/commit/c3115697a1de2e279c1c2d8cae2a9f4cba3d8903)

Yes, I did. At the time it wasn't clear how we'd instantiate `Node`s, so I mimicked the `create_node` API from `rclcpp`. And there was no context back then.

> Another comment: `rclpy.create_node` accepts many more arguments than `rclrs::create_node`.

What does that have anything to do with this discussion? I don't understand.

nnmm (commented):
@esteve I'll reply to the meta/process issues elsewhere. But on the topic itself:

I still find my points convincing, and prefer not to follow `rclpy` here. The argument from the composition API is a bit vague imo, I'm not even sure we ever need a `Node` trait. But at this point I don't feel like anyone has incomplete information, and I'll remove the blocker from my review.

I'd just like to hear from @jhdcs whether his opinion has stayed the same after hearing mine.

jhdcs:
However, I will note there appear to be build errors.

nnmm:
Yeah, agree that `rclcpp` is a closer neighbor than `rclpy`.

nnmm:
@esteve Don't forget to also update the [tutorial](https://github.com/ros2-rust/ros2_rust/blob/main/docs/writing-your-first-rclrs-node.md).

esteve:
@nnmm there doesn't seem to be any reference to `create_node` in the tutorial, unless I've missed something.

jhdcs:
@esteve @nnmm I don't see any references either. Everything seems to use `Node::new()`

nnmm:
Oh, that's fine then.

PR #202, opened 2022-06-13
Fix publisher handle variable name

esteve:
This fixes `main` not building due to #194


jhdcs (approved):

nnmm (approved):

jhdcs:
The tests appear to be failing from an internal server error where the repositories are stored. Should we wait for that to get 
resolved before merging?

esteve:
@jhdcs @nnmm thanks for the prompt reviews, I'll wait until CI passes, but it looks like there's an issue with docker.io being down.

Issue #167, opened 2022-05-16
Warn or error on unknown ROS arguments


nnmm:
Currently, there is no warning for unknown ROS arguments, for instance:
```
cargo run --bin minimal_subscriber -- --ros-args --params-file=lol.yaml
```

This does _not_ load the parameter file lol.yaml, since ROS 2 doesn't support the `=` syntax here (it should be a space instead).

Nizerlak:
Depending on which approach would be chosen (warn or error) there should be logging (#184) implemented before or not. In case of warn prompting, logging mechanism is necessary, since error won't be returned from [Context::new()](https://github.com/ros2-rust/ros2_rust/blob/main/rclrs/src/context.rs#L62) and log call should happen there. Error raising variant would be simpler, since the only thing to do would be to add new enum to [RclrsError](https://github.com/ros2-rust/ros2_rust/blob/main/rclrs/src/error.rs#L8) (with Display implementation) and handle unparsed arguments similarly to non-ROS ones as in #147. 

I think that warn is more appropriate and logging mechanism should be implemented anyway, so in this case this issue will be blocked by #184. To make work more parallel maybe introduce bare logging interface without implementation, so call to it could be performed and this issue will be independent of #184 progress.

nnmm:
Yes, I agree logging is probably the way to go. Theoretically we could also make `Context::new()` return something like

```rust
struct ContextWithExtraInfo {
    context: Context,
    unparsed_ros_args: Vec<String>
}
```

but that's cumbersome to use.

Lucas-Rodrigues2:
Proposed approach for issue #167

Thanks for keeping this issue open. I would like to work on it.

The current code already has the internal `arguments::get_rcl_arguments` helper, and logging is now implemented. My proposed implementation is:

1. after `rcl_init` succeeds, query `rcl_context.global_arguments` with `rcl_arguments_get_count_unparsed_ros` and `rcl_arguments_get_unparsed_ros`;
2. after `LoggingLifecycle::configure` succeeds, emit one warning from the `rclrs.context` logger containing the unparsed ROS arguments in their original order;
3. keep non-ROS arguments unaffected;
4. add regression coverage for an unknown ROS argument such as `--params-file=missing.yaml`, plus a control case for a valid spelling using a separate argument.

I would prefer a warning rather than a new `RclrsError` variant because that matches the direction discussed above and avoids making context creation fail for arguments that `rcl` currently tolerates.

[...]

PR #210, opened 2022-06-27
 Add function to get non-ROS arguments

Nizerlak:
Hi guys,
This is my first PR in this repo.
I faced an issue during creating this PR. In order to obtain non-ROS args, the input args (passed from console on node startup) has to be known, which is possible only during `Context` initialization (`new` function). My solution was to get those non-ROS arguments and store in new field of `Context` - `non_ros_arguments` - which could be accessed by getter function. Unfortunately it generated conflict with [lib.rs](https://github.com/ros2-rust/ros2_rust/pull/210/files#diff-eac337ff956243a705728ddf8a51070fa83f01b62bcf6949de01133259876ca5R39). I made a temporary fix, but I think that, a design of Node and/or Context should be revised to resolve this conflict.


Nizerlak:
`Clone` implementation for `Context` would resolve described problem. Also it could make mentioned lib.rs more elegant (call for `clone` instead of building struct with calls to fields clone anyway).

rclrs/src/context.rs
```diff
@@ -118,6 +127,47 @@ impl Context {
...
 
+/// Returns non-ROS arguments held by context.
+///
+/// This function should be called after rcl_init. `rcl_arguments` should be `global_arguments`
+/// field of initialized context. `args` is array of input arguments passed to node.
+fn get_non_ros_arguments(
+    rcl_arguments: *const rcl_arguments_t,
+    args: &[*const c_char],
+) -> Result<Vec<String>, RclrsError> {
+    // SAFETY: In case of error -1 will be returned and handled a line below.
```
jhdcs, line 138: The purpose of a safety comment (in my mind) is to indicate what things need to be done in order to prevent undefined behavior (UB) from occurring from the result of running an `unsafe` function, or from using its output. In other words, what needs to be done to keep the `unsafe` function from causing UB, or what needs to be done to safely use its output.

Since returning a `-1` will not cause UB, and because using/accessing a -1 is safe, I believe that this should be:

```suggestion
    // SAFETY: No preconditions for this function.
```

rclrs/src/context.rs
```diff
@@ -118,6 +127,47 @@ impl Context {
...
+    let args_status = unsafe { rcl_arguments_get_count_unparsed(rcl_arguments) };
+    if args_status == -1 {
+        return Err(RclrsError::RclError {
+            code: InvalidArgument,
+            msg: None,
+        });
+    }
```
jhdcs, line 145:
```suggestion
    let args_status = unsafe { rcl_arguments_get_count_unparsed(rcl_arguments).ok()? };
```

rclrs/src/context.rs
```diff
@@ -118,6 +127,47 @@ impl Context {
...
+    // SAFETY: All possible negative args_status values handled above.
+    let args_num = args_status as usize;
+    if args_num == 0 {
+        return Ok(Vec::new());
+    }
```
jhdcs, line 150: The `as` keyword signifies a safe cast, so this cannot cause undefined behavior. As such, a safety comment isn't needed here (those should be for dealing with `unsafe` and UB.

HOWEVER, it's good to document why the cast isn't going to cause a logic error.
```suggestion
    // All possible negative args_status values were handled above.
    let args_num = args_status as usize;
    if args_num == 0 {
        return Ok(Vec::new());
    }
```

rclrs/src/context.rs
```diff
@@ -118,6 +127,47 @@ impl Context {
...
+    let mut out_ptr: *mut i32 = null_mut();
+    unsafe {
+        // SAFETY: No preconditions for this function.
+        let allocator = rcutils_get_default_allocator();
+        // SAFETY: Error will be returned in case of broken rcl_arguments, no precondition for the rest.
+        rcl_arguments_get_unparsed(rcl_arguments, allocator, &mut out_ptr).ok()?;
```
jhdcs, line 156: This safety comment needs to be expanded as to why `out_ptr` can be safely used. Specifically, in the event of no arguments, `out_ptr` will be null, though an error will not be raised.

rclrs/src/context.rs
```diff
@@ -118,6 +127,47 @@ impl Context {
...
+    }
+    // SAFETY: out_ptr won't be allocated in case of error and this code won't be reached.
+    // Indices of non-ROS args in input args array
+    let indices: Vec<i32> = unsafe { Vec::from_raw_parts(out_ptr, args_num, args_num) };
+    let mut non_ros_args: Vec<String> = Vec::with_capacity(args_num);
+
+    for i in indices.iter() {
+        // SAFETY: args are expected to have been transformed from String. Indices are expected to be positive.
```
jhdcs, line 164: I don't think this is necessarily a safe assumption - I believe most arguments provided to the program as OSString, though I may be mistaken on that front.

rclrs/src/context.rs
```diff
@@ -118,6 +127,47 @@ impl Context {
...
+    let mut out_ptr: *mut i32 = null_mut();
+    unsafe {
+        // SAFETY: No preconditions for this function.
+        let allocator = rcutils_get_default_allocator();
+        // SAFETY: Error will be returned in case of broken rcl_arguments, no precondition for the rest.
+        rcl_arguments_get_unparsed(rcl_arguments, allocator, &mut out_ptr).ok()?;
+    }
+    // SAFETY: out_ptr won't be allocated in case of error and this code won't be reached.
+    // Indices of non-ROS args in input args array
+    let indices: Vec<i32> = unsafe { Vec::from_raw_parts(out_ptr, args_num, args_num) };
```
jhdcs, line 160: According to the official [documentation](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts), this function is highly unsafe, and I'm not sure that all the invariants are accounted for...

For example, `out_ptr` needs to have been previously allocated via `String` or `Vec`, whereas our `out_ptr` was populated via a C string array, which isn't quite the same thing.

Is it possible to do this via any other method?

jhdcs (changes requested):
I have a few safety concerns, especially the use of `Vec::from_raw_parts`... Could you please take a look at the suggested changes and address them?

Otherwise, this looks like a good start! Hopefully I haven't missed anything. And thank you for your contribution!!

nnmm:
@Nizerlak Thanks for the PR, and you're right about the issue in lib.rs. That is a bit hacky and needs to be refactored.

@jhdcs @esteve I think the cleanest solution is to add a `&Context` argument to the spin functions. Are you fine with that?

esteve:
> @jhdcs @esteve I think the cleanest solution is to add a &Context argument to the spin functions. Are you fine with that?

I'd void that as much a possible, after all, `Node`s have a reference to `Context`, so adding a `Context` is redundant. How does `rclcpp` tackle this?

nnmm:
> I'd void that as much a possible, after all, `Node`s have a reference to `Context`, so adding a `Context` is redundant. How does `rclcpp` tackle this?

`rclcpp` uses the implicit global context.

Nodes don't have a reference to `Context` but only to the `rcl context_t`. If we want them to have a full `Context`, we need to make `Context::new` return an `Arc` or possibly `Arc<Mutex<_>>` – that's also possible but a little ugly/inefficient.

A third way is to have a private version of `WaitSet::new()` that takes an `rcl_context_t` instead of a `&Context`. Then we don't need a full `Context` inside of lib.rs.

Nizerlak:
> > @jhdcs @esteve I think the cleanest solution is to add a &Context argument to the spin functions. Are you fine with that?
> 
> I'd void that as much a possible, after all, `Node`s have a reference to `Context`, so adding a `Context` is redundant. How does `rclcpp` tackle this?

So maybe refactor `WaitSet` to receive in `new` rcl context mutex? [Here](https://github.com/ros2-rust/ros2_rust/blob/main/rclrs/src/wait.rs#L74) it is unpacking inner mutex from context anyway so [here](https://github.com/ros2-rust/ros2_rust/blob/main/rclrs/src/lib.rs#L41) it could be passed directly without wrapping with `Context`.

In [rclcpp](https://github.com/ros2/rclcpp/blob/33dae5d679751b603205008fcb31755986bcee1c/rclcpp/include/rclcpp/wait_set_template.hpp#L80) i can see that wrapped `Context` is passed, but it doesn't hold information about unparsed argumets.

The only way they're dealing with unprased arguments is throwing an exception [here](https://github.com/ros2/rclcpp/blob/33dae5d679751b603205008fcb31755986bcee1c/rclcpp/src/rclcpp/context.cpp#L237) in case of bad ROS arguments (equivalent of ours #167). In order to get non-ROS arguments they retrieve arguments structure from `Node` and call rcl function directly [here](https://github.com/ros2/rclcpp/blob/33dae5d679751b603205008fcb31755986bcee1c/rclcpp/test/rclcpp/test_node_options.cpp#L87). Maybe we should also utilize this approach and deal with non-ROS arguments independently.

rclrs/src/context.rs
Nizerlak: My bad, didn't even check if C pointer can be freely deallocated by `Vec`.

Another approach would be parsing through memory allocated to `out_ptr` and call `deallocate` on `allocator` like this:

```rs
let mut indices = Vec::with_capacity(args_num);

for i in [0..args_num] {
    indices.push_back(*(out_ptr + i));
}

allocator.deallocate(out_ptr, null());
```
Note that it is safe to pass `NULL` to deallocate, since it's not used by [rcl_allocator_t implementation](https://github.com/ros2/rcutils/blob/master/src/allocator.c#L43).

The only safety precaution is that memory allocated during `rcl_arguments_get_unparsed` call has the exact same size as `rcl_arguments_get_count_unparsed` returns.

rclrs/src/context.rs
Nizerlak: Actually there shouldn't be a situation when there's no arguments as stated [here](https://github.com/ros2/rcl/blob/foxy/rcl/include/rcl/arguments.h#L150), but yeah in this code such situation should be considered.

rclrs/src/context.rs
Nizerlak: `rcl_arguments_get_count_unparsed` does not return `Result` but  `::std::os::raw::c_int` as I can see in generated bindings:
```rs
pub fn rcl_arguments_get_count_unparsed(args: *const rcl_arguments_t) -> ::std::os::raw::c_int;
```
Could you please explain this suggestion?

Also, suggested change does not compile:
```
error[E0308]: mismatched types
   --> src/context.rs:140:23
    |
140 |     if args_status == -1 {
    |                       ^^ expected `()`, found integer

error[E0605]: non-primitive cast: `()` as `usize`
   --> src/context.rs:147:20
    |
147 |     let args_num = args_status as usize;
    |                    ^^^^^^^^^^^^^^^^^^^^ an `as` expression can only be used to convert between primitive types or to coerce to a specific trait object
```

rclrs/src/context.rs
Nizerlak: Oh right, seems that I misunderstood the concept of `SAFETY` comments. Indeed `rcl_arguments_get_count_unparsed` needs no precondition, since it's so foolproof, that it can deal with `NULL` also.

Nizerlak:
Also, I have a question about workflow in this repo. While applying fixes to this PR shall I 
1. make fix commits 
              or 
2. rebase with corresponding commits edition and force push?

nnmm:
> Also, I have a question about workflow in this repo. While applying fixes to this PR shall I
> 
>     1. make fix commits
>        or
> 
>     2. rebase with corresponding commits edition and force push?

Good question. I think it's fine to force-push in the very early stages (e.g. you noticed that CI isn't passing), but generally fix commits are a little easier for reviewers. We squash commits when merging anyway.

nnmm:
> So maybe refactor `WaitSet` to receive in `new` rcl context mutex? [Here](https://github.com/ros2-rust/ros2_rust/blob/main/rclrs/src/wait.rs#L74) it is unpacking inner mutex from context anyway so [here](https://github.com/ros2-rust/ros2_rust/blob/main/rclrs/src/lib.rs#L41) it could be passed directly without wrapping with `Context`.

That is basically what I'm suggesting with my comment "A third way is to have a private version of `WaitSet::new()` that takes an `rcl_context_t` instead of a `&Context`." We will need two versions: `pub new()` like it is now, and `new_from_rcl_context()` (not `pub`) that  takes `rcl_context_t`. Public APIs should not have the `rcl` types in it. `new()` can then unpack the `rcl_context` and call `new_from_rcl_context()`.

rclrs/src/context.rs
nnmm: I think this should be

```
let args_count = unsafe { rcl_arguments_get_count_unparsed(rcl_arguments) };
debug_assert!(args_count != -1);
```

See also https://github.com/ros2-rust/ros2_rust/blob/main/docs/CONTRIBUTING.md#return-type-to-use-when-wrapping-rcl-functions.

Nizerlak:
I've implemented another approach. `non_ros_arguments` is member function of `Context` and should be called with program input arguments. Possible problem may occur, when user calls this function with wrong `input_args` for some reason. However, problem with holding arguments as `Context` member is resolved.

Nizerlak:
Also, I've generalized getting rcl_arguments, which may be useful while resolving #167, since identical steps to obtain ROS unparsed arguments would be performed as for non-ROS ones.

rclrs/src/rcl_utils.rs
```diff
@@ -0,0 +1,101 @@
...
+pub(crate) trait RclGetArg {
+    /// Get count function from rcl library (e.g. rcl_arguments_get_count_unparsed)
+    unsafe fn get_count(rcl_args: *const rcl_arguments_t) -> ::std::os::raw::c_int;
+
+    /// Get function from rcl library (e.g. rcl_arguments_get_unparsed)
+    unsafe fn get_indices(
+        rcl_args: *const rcl_arguments_t,
+        allocator: rcl_allocator_t,
+        out_ptr: *mut *mut ::std::os::raw::c_int,
+    ) -> rcl_ret_t;
```
jhdcs, line 17: Is it possible/desirable to have this return a `Result` instead? Akin to how we use `.ok()?` for most of our calls to `rcl`?

rclrs/src/rcl_utils.rs
```diff
@@ -0,0 +1,101 @@
...
+        return Ok(Vec::new());
+    }
+    let mut non_ros_args: Vec<String> = Vec::with_capacity(args_count);
+    unsafe {
+        let mut indices_ptr: *mut i32 = null_mut();
+        // SAFETY: No preconditions for next 2 functions.
+        let allocator = rcutils_get_default_allocator();
+        // Indices will be set to NULL in case of no arguments handled, but then this code won't be
+        // reached because of if statement above.
+        ArgGetter::get_indices(rcl_arguments, allocator, &mut indices_ptr).ok()?;
```
jhdcs, line 87:
```suggestion
        // SAFETY: No preconditions for this function.
        let allocator = rcutils_get_default_allocator();
        // Indices will be set to NULL in case of no arguments handled, but then this code won't be
        // reached because of if statement above.
        //
        // SAFETY: No preconditions for this function.
        ArgGetter::get_indices(rcl_arguments, allocator, &mut indices_ptr).ok()?;
```

rclrs/src/rcl_utils.rs
```diff
@@ -0,0 +1,101 @@
...
+        for i in 0..args_count {
+            // If get_indices finishes with success, indices_ptr is valid
+            // and is allocated with size equal to one returned by rcl_arguments_get_count_unparsed.
+            // SAFETY: No preconditions for this function.
+            let index = *(indices_ptr.add(i));
+            non_ros_args.push(args.get(index as usize).unwrap().clone());
+        }
+        // SAFETY: No preconditions for next 2 functions.
+        let allocator = rcutils_get_default_allocator();
+        allocator.deallocate.unwrap()(indices_ptr as *mut c_void, null_mut());
```
jhdcs, line 98:
```suggestion
        // SAFETY: No preconditions for this function.
        let allocator = rcutils_get_default_allocator();
        // SAFETY: No preconditions for this function.
        allocator.deallocate.unwrap()(indices_ptr as *mut c_void, null_mut());
```

Additionally, is there a way to re-write this to avoid the `unwrap()`?

rclrs/src/rcl_utils.rs
```diff
@@ -0,0 +1,101 @@
...
+        // Indices will be set to NULL in case of no arguments handled, but then this code won't be
+        // reached because of if statement above.
+        ArgGetter::get_indices(rcl_arguments, allocator, &mut indices_ptr).ok()?;
+
+        for i in 0..args_count {
+            // If get_indices finishes with success, indices_ptr is valid
+            // and is allocated with size equal to one returned by rcl_arguments_get_count_unparsed.
+            // SAFETY: No preconditions for this function.
+            let index = *(indices_ptr.add(i));
+            non_ros_args.push(args.get(index as usize).unwrap().clone());
```
jhdcs, line 94: Is there a way to re-write this to avoid the `unwrap()`?

rclrs/src/rcl_utils.rs
```diff
@@ -0,0 +1,101 @@
...
+///
+/// Tag designated to be used as generic argument of [`get_rcl_arguments`].
+pub(crate) struct UnparsedNonRos;
+
+/// Macro to avoid boilerplate code for above tags.
+///
+/// tag - one of above tags
+/// get_count_function - function for getting count of given arguments type
+/// get_indices_function - function for getting indices of given arguments type
+macro_rules! impl_get_arg {
```
jhdcs, line 35: Is the macro here really necessary? Will we be using this for more cases than just the above two tags?

Additionally, it creates two `unsafe` functions that won't have (and possibly can't have) safety comments, which makes me a little nervous.

@nnmm, @esteve, what are your thoughts?

jhdcs (changes requested):
Sorry for the delay - it was a long weekend over here.

nnmm:
> I've implemented another approach. non_ros_arguments is member function of Context and should be called with program input arguments. Possible problem may occur, when user calls this function with wrong input_args for some reason. However, problem with holding arguments as Context member is resolved.

That's also a possibility, but I actually think our idea of constructing the wait set with an `rcl_context_t` is better, since (1) the end result is more user-friendly and (2) we'll probably run into the same problem of wanting to store something in `Context` soon again, so why not fix it properly now. @esteve @jhdcs let me know if you don't agree.

rclrs/src/rcl_utils.rs
nnmm: Well, I've only been focusing on the high-level design so far, see https://github.com/ros2-rust/ros2_rust/pull/210#issuecomment-1172905936 and https://github.com/ros2-rust/ros2_rust/pull/210#issuecomment-1175018118. Until that's settled maybe it's a bit early to concern ourselves with implementation details.

esteve: Doesn't seem to me that it'll be used much more than that, I'd rewrite this as a function

esteve: @nnmm I don't think it's never too early, it's fair to provide any feedback while we review this PR

esteve:
It seems to be that we're running in circles just to avoid what `rclcpp` did (an implicit context), it's not the cleanest solution, that's true, but all this going back and forth just in a way shows that perhaps `rclcpp`'s solution is worth considering after all.

jhdcs:
> It seems to be that we're running in circles just to avoid what `rclcpp` did (an implicit context), it's not the cleanest solution, that's true, but all this going back and forth just in a way shows that perhaps `rclcpp`'s solution is worth considering after all.

Perhaps we should discuss this in a separate issue? I foresee the potential for a very long discussion on the merits, and I don't want to hold up @Nizerlak's PR if it can be avoided. Especially if we're fundamentally restructuring `rclrs` underneath him from that discussion.

The work coming from that discussion should be a separate PR.

rclrs/src/rcl_utils.rs
Nizerlak: So it's ok to leave this rcl_utils.rs file and rewrite only that macro part, is it?

nnmm:
> It seems to be that we're running in circles just to avoid what rclcpp did (an implicit context), it's not the cleanest solution, that's true, but all this going back and forth just in a way shows that perhaps rclcpp's solution is worth considering after all.

I'm not against considering it in principle, but (1) we'd have the same problem here if we had a global context (since we'd still want the context instance associated with the node to be used in the wait set – it's the same in rclcpp) and (2) it's quite easy to fix.

rclrs/src/rcl_utils.rs
Nizerlak: It was intended to be just a wrapper around corresponding `rcl_argument_get...` and `rcl_argument_get_count...`, so i left the interface repeated.

rclrs/src/rcl_utils.rs
Nizerlak: I don't see a reason why `unwrap` could return `None` here, since `allocator` is retrieved by `rcutils_get_default_allocator()` just above. I could add another error type to `RclrsError` like I suggested [here](https://github.com/ros2-rust/ros2_rust/pull/210#discussion_r914144900), but still I don't see a reason for that.

esteve:
@Nizerlak thank you so much for iterating with us. Here's an idea that we've discussed with @jhdcs and @nnmm, let us know what you think of it. How about a free function that just extracts the arguments from `env::args()`? Its usage would be like the following:

```rust

fn main() {
    let non_ros_args = rclrs::extract_non_ros_args(env::args());
    // do rclrs stuff, create Context, create Node, etc.
}
```

Most of the time, once the command line arguments have been parsed, there's no need to access them again. This has the advantage that we're not storing anything for the entire lifetime of a `Node` or a `Context`, just get the arguments we're interested in, and then execute the rest of the logic of our program. With a solution like this, we could even parse arguments and only initialize `rclrs` constructs after all has been processed by `clap`, and not before (e.g. ensure that all arguments are correct before doing anything). What do you think?

Nizerlak:
I like that idea @esteve. With this approach there's no problem when user passes different arguments while calling `Context::new` and `Context::non_ros_arguments` and detaches non-ROS things (which are non-ROS args obviously) from ROS things - `Context`. I'll apply this suggestion as soon as possible.

rclrs/src/lib.rs
```diff
@@ -119,6 +124,81 @@ pub fn create_node_builder(context: &Context, node_name: &str) -> NodeBuilder {
...
+        .into_iter()
+        .map(|arg| {
+            CString::new(arg.as_str()).map_err(|err| RclrsError::StringContainsNul {
+                err,
+                s: arg.clone(),
+            })
+        })
+        .collect::<Result<_, _>>()?;
+    // Vector of pointers into cstring_args
+    let c_args: Vec<*const c_char> = cstring_args.iter().map(|arg| arg.as_ptr()).collect();
```
Nizerlak, line 159: Here is a code repetition from [context.rs](https://github.com/ros2-rust/ros2_rust/blob/main/rclrs/src/context.rs#L65). Initially I've isolated it to a function, but that's lead to double free error during UTs for some reason. Are you ok with this repetition? If not, could you please give me a hint how can I isolate this fragment?

nnmm: Yes, that's okay.

nnmm: To expand on that, if you try to make a function like
```rust
fn to_c_char_ptrs(args: &[String]) -> Result<Vec<*const c_char>, RclrsError> {
    let cstring_args: Vec<CString> = args
        .into_iter()
        .map(|arg| {
            CString::new(arg.as_str()).map_err(|err| RclrsError::StringContainsNul {
                err,
                s: arg.clone(),
            })
        })
        .collect::<Result<_, _>>()?;
    // Vector of pointers into cstring_args
    Ok(cstring_args.iter().map(|arg| arg.as_ptr()).collect())
}
```
then the returned pointers will be dangling (the `cstring_args` they're pointing into will be gone after the function returns). 

We could either do something clever where we return the `cstring_args` along with the `c_args` in one struct, or we could make it a macro instead of a function. But leaving it like it is is fine for now, we can do that in a separate PR.

nnmm:
@Nizerlak Is this PR ready for a review?

Nizerlak:
Yes, I forgot to mention. Sorry for that.

rclrs/src/lib.rs
```diff
@@ -119,6 +124,81 @@ pub fn create_node_builder(context: &Context, node_name: &str) -> NodeBuilder {
...
+            allocator,
+            &mut rcl_arguments,
+        )
+        .ok()
+    };
+
+    if let Err(err) = ret {
+        // SAFETY: No preconditions for this function
+        unsafe {
+            rcl_arguments_fini(&mut rcl_arguments).ok()?;
```
nnmm, line 179: I think it's safe to rely on this being done by `rcl` if `rcl_parse_arguments()` fails: https://github.com/ros2/rcl/blob/4eccc3cbe65f5b58981f22efaf612a65731cb2f0/rcl/src/rcl/arguments.c#L670

rclrs/src/rcl_utils.rs
```diff
@@ -0,0 +1,104 @@
+use crate::rcl_bindings::*;
+use crate::{RclrsError, ToResult};
+
+use crate::RclrsError::IndexOutOfRange;
+use libc::c_void;
+use std::ptr::null_mut;
+
+/// Trait used for obtaining information from rcl_arguments_t.
+pub(crate) trait RclGetArg {
```
nnmm, line 9: A trait seems to be a little much abstraction. With two function pointers as arguments to `get_rcl_arguments()`, the code would be shorter and in my opinion simpler.

rclrs/src/rcl_utils.rs
```diff
@@ -0,0 +1,104 @@
...
+pub(crate) struct UnparsedNonRos;
+
+/// Get ROS unparsed arguments tag.
+///
+/// Tag designated to be used as generic argument of [`get_rcl_arguments`].
+pub(crate) struct UnparsedRos;
+
+/// Implementation is safe, since both functions corresponds to each other as rcl documentation states.
+impl RclGetArg for UnparsedNonRos {
+    unsafe fn get_count(rcl_args: *const rcl_arguments_t) -> ::std::os::raw::c_int {
```
jhdcs, line 33: This should have a `// SAFETY:` comment.

rclrs/src/rcl_utils.rs
```diff
@@ -0,0 +1,104 @@
...
+        rcl_arguments_get_count_unparsed(rcl_args)
+    }
+
+    unsafe fn get_indices(
```
jhdcs, line 37: This should have a `// SAFETY:` comment.

jhdcs (changes requested):
Mostly safety comments and style critiques. But I feel this is getting closer!

rclrs/src/rcl_utils.rs
Nizerlak: I wanted to provide some kind of ready sets of functions to use those `rcl_argument_` API. This way seemed to be the only one. I know it may be an overkill, but I'm still learning this language and treated this like a kind of exercise.

rclrs/src/error.rs
```diff
@@ -34,6 +34,8 @@ pub enum RclrsError {
         /// Index provided
         wrong_index: usize,
     },
+    /// It was attempted to add a waitable to a wait set twice.
```
jhdcs, line 37:
```suggestion
    /// Given waitable was already added to a wait set
```

rclrs/src/lib.rs
```diff
@@ -302,7 +297,11 @@ mod tests {
                     return Err(msg);
                 }
             }
-            Ok(())
         }
+
+        let empty_non_ros_args = extract_non_ros_args(vec![]).unwrap();
```
jhdcs, line 302: I feel like this should actually be spun into a separate test function - the other two cases are both parsing the same dataset for testing, while this case is parsing a new (empty) dataset.

jhdcs (changes requested):
A couple more changes, otherwise looks ok to me, I think...

rclrs/src/error.rs
Nizerlak: It was not part of this PR. It appeared there after mercedes.

rclrs/src/lib.rs
```diff
@@ -148,6 +153,82 @@ pub fn create_node_builder(context: &Context, node_name: &str) -> NodeBuilder {
     Node::builder(context, node_name)
 }
 
+/// Extract non-ROS arguments from program's input arguments.
+///
+/// `args` is expected to be input arguments of the program (passed via [`env::args()`])
+/// According to rcl documentation ROS arguments are between `--ros-args` and `--` arguments.
+/// Everything beside is considered as non-ROS and left unparsed. Extracted non-ROS args are
+/// returned in order as they appear (see example below).
```
jhdcs, line 161:
```suggestion
/// `args` is expected to be the input arguments of the program (passed via [`env::args()`]).
/// According to rcl documentation, ROS arguments are between `--ros-args` and `--` arguments.
/// Everything else is considered as non-ROS and left unparsed. Extracted non-ROS args are
/// returned in the order that they appear (see example below).
```

Minor change, looks a bit more natural.

jhdcs (approved):
Only one small suggested comment wording change. However, since it's so minor, I will mark this PR as approved.

Thank you _**very much**_ for your contribution, and your patience! Very nice work!

nnmm:
@jhdcs Could you wait with merging until I approve too? Thanks.

jhdcs (dismissed):
Just a question on how things are organized.

rclrs/src/rcl_utils.rs
```diff
@@ -0,0 +1,61 @@
...
+/// Returns arguments type held by `rcl_arguments` basing on `rcl_get_count` and `rcl_get_indices` function pointers.
+///
+/// This function must be called after `rcl_arguments` was initialized. `args` must be array of input arguments passed to node/program.
+///
+/// SAFETY: `rcl_get_count` and `rcl_get_indices` has to be corresponding rcl API functions, e.g.:
+/// `rcl_arguments_get_count_unparsed` -> `rcl_arguments_get_unparsed`
+/// `rcl_arguments_get_count_unparsed_ros` -> `rcl_arguments_get_count_ros`
+/// ...
+pub(crate) fn get_rcl_arguments(
+    rcl_get_count: unsafe extern "C" fn(*const rcl_arguments_t) -> std::os::raw::c_int,
```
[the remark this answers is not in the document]
Nizerlak: Honestly I don't see how to wrap this common portions to separate functions. They're just too small and linked tightly to te rest. Maybe wrapping it into macro woud solve this, but it's a bit nasty and seems like overkill.
Current approach is on @nnmm request, so if there's still something wrong with it please agree on one version, since I'd like to start resolving other issues not being attached to this one for so long :/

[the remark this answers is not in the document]
jhdcs: Ah, I had missed that. My apologies.

rclrs/src/lib.rs
```diff
@@ -147,3 +152,136 @@ pub fn create_node(context: &Context, node_name: &str) -> Result<Node, RclrsErro
...
+/// Extract non-ROS arguments from program's input arguments.
+///
+/// `args` is expected to be the input arguments of the program (passed via [`env::args()`]).
+/// According to rcl documentation, ROS arguments are between `--ros-args` and `--` arguments.
+/// Everything else is considered as non-ROS and left unparsed. Extracted non-ROS args are
+/// returned in the order that they appear (see example below).
+///
+/// # Example
+/// ```
+/// let input_args: Vec<String> = vec![
```
nnmm, line 165: In the other examples, I've used arrays since they're shorter and nowadays they also implement `IntoIterator`. But up to you.

rclrs/src/lib.rs
```diff
@@ -147,3 +152,136 @@ pub fn create_node(context: &Context, node_name: &str) -> Result<Node, RclrsErro
...
+///         .collect();
+/// let non_ros_args = rclrs::extract_non_ros_args(input_args).unwrap();
+/// assert_eq!(non_ros_args.len(), 2);
+/// assert_eq!(non_ros_args[0], "arg1");
+/// assert_eq!(non_ros_args[1], "arg2");
+/// ```
+pub fn extract_non_ros_args(
+    args: impl IntoIterator<Item = String>,
+) -> Result<Vec<String>, RclrsError> {
+    // SAFETY: No preconditions for this function
```
nnmm, line 179: Nitpick: We've used this phrasing elsewhere, would be ideal to use it here too for consistency:

```suggestion
    // SAFETY: Getting a zero-initialized value is always safe.
```

rclrs/src/rcl_utils.rs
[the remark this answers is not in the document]
nnmm: Yes, that PR has gone on quite long, and I agree with @Nizerlak that no more common functions can be factored out.

nnmm:
I'm really happy with the code quality in this PR overall, most of my comments relate to the `SAFETY` comments which seem like they maybe seemed a bit pointless to you :) I encourage you to just be a bit creative with these, and write down notable comments about usage of this function even if there is nothing strictly safety-related to be said about it.

One request: Could you move everything into an `arguments.rs` module that is `pub use`d by `lib.rs`?

And sorry about the long review time – the next one will be faster.

rclrs/src/arguments.rs
```diff
@@ -0,0 +1,122 @@
...
+/// # Example
+/// ```
+/// let input_args : [String;6] = [
+///             "arg1", "--ros-args", "some", "args", "--", "arg2"
+///         ].map(|x| x.to_string());
+/// let non_ros_args = rclrs::extract_non_ros_args(input_args).unwrap();
+/// assert_eq!(non_ros_args.len(), 2);
+/// assert_eq!(non_ros_args[0], "arg1");
+/// assert_eq!(non_ros_args[1], "arg2");
+/// ```
```
nnmm, line 24: Just some minor suggestions for improvements (using ? in example, formatting, mentioning that final `--` is optional, etc)

```suggestion
/// Extract non-ROS arguments from program's input arguments.
///
/// `args` is expected to be the input arguments of the program (e.g. [`std::env::args()`]),
/// which are expected to contain at least one element - the executable name.
/// 
/// ROS arguments are arguments between `--ros-args` and `--`, with the final `--` being optional.
/// Everything else is considered as non-ROS arguments and will be left unparsed by 
/// [`Context::new()`][1] etc.
/// Extracted non-ROS arguments are returned in the order that they appear by this function.
///
/// # Example
/// ```
/// # use rclrs::RclrsError;
/// let input_args = [
///     "arg1", "--ros-args", "some", "args", "--", "arg2"
/// ].map(|x| x.to_string());
/// let non_ros_args = rclrs::extract_non_ros_args(input_args)?;
/// assert_eq!(non_ros_args.len(), 2);
/// assert_eq!(non_ros_args[0], "arg1");
/// assert_eq!(non_ros_args[1], "arg2");
/// # Ok::<(), RclrsError>(())
/// ```
/// 
/// [1]: crate::Context::new
```

nnmm:
Thanks for making all the changes! Could you just also move the code from `rcl_utils.rs` into `arguments.rs`?

nnmm (approved):
I think we're just one `cargo fmt` away from merging it.

nnmm (approved):
LGTM, thank you. I'm looking forward to what you'll tackle next :-)

PR #212, opened 2022-07-02
Add support for loaned messages

### LINKED ISSUE #136: Add support for zero-copy pub/sub
This could be a distinguishing feature of `ros2_rust` compared to other Rust client libraries.

The loaned message mechanism might be a good fit for Rust's reference types (or maybe not).

nnmm:
I tested it with CycloneDDS and it worked.

The only interesting things are
- I stored a reference to the publisher/subscription in the message, so that the loan can be directly returned when the message is dropped (and it is guaranteed that the publisher/subscription still exists)
- I made `publish()` a method on the `LoanedMessage` that takes no argument, since it can only be used with the publisher that was used to create it, and that publisher is stored inside the message. So this design statically guarantees that a loaned message cannot be published to the wrong publisher.

Closes #136

### DIFF
--- a/rclrs/src/node/subscription/readonly_loaned_message.rs
+++ b/rclrs/src/node/subscription/readonly_loaned_message.rs
@@ -1,0 +1,47 @@
+use crate::rcl_bindings::*;
+use crate::{Subscription, ToResult};
+
+use rosidl_runtime_rs::RmwMessage;
+
+use std::ops::Deref;
+
+/// A message that is owned by the middleware, loaned out for reading.
+///
+/// It dereferences to a `&T`.
+///
+/// This type is returned by [`Subscription::take_loaned_message()`].
+///
+/// The loan is returned by dropping the `ReadOnlyLoanedMessage`.
+pub struct ReadOnlyLoanedMessage<'a, T>
+where
+    T: RmwMessage,
+{
+    pub(super) msg_ptr: *const T,
+    pub(super) subscription: &'a Subscription<T>,
+}
+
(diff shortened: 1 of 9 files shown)


rclrs/src/node/subscription/readonly_loaned_message.rs
```diff
@@ -0,0 +1,47 @@
...
+    }
+}
+
+impl<'a, T> Drop for ReadOnlyLoanedMessage<'a, T>
+where
+    T: RmwMessage,
+{
+    fn drop(&mut self) {
+        unsafe {
+            rcl_return_loaned_message_from_subscription(
```
jhdcs, line 39: What happens if a user `.clone()`s a loaned message? Will that affect this function?

jhdcs (changes requested):
One small comment tweak. However, I'm curious about how cloning might affect this... Is it possible to get a double-free error if you clone a loaned message, then drop both? Or am I just being paranoid?

rclrs/src/node/subscription/readonly_loaned_message.rs
nnmm: Yep, that would be bad, that's why loaned messages don't implement `Clone`.

jhdcs (dismissed):
My biggest concern was addressed in a comment, and the documentation change isn't worth holding up the PR.

jhdcs (dismissed):

nnmm:
@esteve Do you also want to have a look at this?

nnmm:
@esteve Please review (or let me know if Jacob's review is sufficient for you).

esteve (approved):
Great work!

PR #223, opened 2022-07-10
Domain id init

DS3a:
I've added the a `ContextBuilder` which I am using in `new()` in `context.rs` please review this so that I can proceed to adding domain_id options by overrloading these functions using `ros_distro` cfg attributes


esteve:
@DS3a thanks for your PR, sorry it took so long to get back to you. We now have an API for setting the domain id (https://github.com/ros2-rust/ros2_rust/blob/main/rclrs/src/context.rs#L260-L263)

PR #224, opened 2022-07-11
Ensure that a waitable is only added to one wait set at a time

### LINKED ISSUE #207: Track all waitables that are currently in a wait set
During some refactoring to remove `SubscriptionBase`, I noticed that the wait set currently violates the unsafe code principles: Safe code can cause undefined behavior by adding the same item to multiple wait sets in different threads and waiting on them. From the `rcl` docs for `rcl_wait()`:

>  This function is thread-safe for unique wait sets with unique contents.
This function cannot operate on the same wait set in multiple threads, and the wait sets may not share content.
For example, calling rcl_wait() in two threads on two different wait sets that both contain a single, shared guard condition is undefined behavior.

We need to either make all functions that end up calling `rcl_wait()` unsafe, like `rclrs::spin()`, or we need to guarantee that wait sets do not share contents. The latter could be done by having a global list of waitables that are currently in a wait set, and would be my favored approach.

nnmm:
Closes #207.

The use of atomic bools follows `rclcpp`'s approach.

### DIFF
--- a/rclrs/src/error.rs
+++ b/rclrs/src/error.rs
@@ -27,6 +27,8 @@ pub enum RclrsError {
         /// The error indicating the position of the nul byte.
         err: NulError,
     },
+    /// It was attempted to add a waitable to a wait set twice.
+    AlreadyAddedToWaitSet,
 }
 
 impl Display for RclrsError {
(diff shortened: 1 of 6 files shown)


esteve:
@nnmm looks good overall. Could you add a unit test for `ExclusivityGuard` and perhaps another for when a `Subscription`/`Client`/`Service` is added to a `WaitSet` twice and to two different `WaitSet`s? Thanks.

esteve (approved):

esteve:
@nnmm thanks!

PR #228, opened 2022-07-15
Fix path handling in rosidl_generator_rs on Windows

nnmm:
Paths on Windows can contain colons. With rsplit, the drive letter was grouped with the package name.


esteve (approved):
Might be worth using [pathlib](https://docs.python.org/3/library/pathlib.html) for this to ensure cross-platform compatibility, but it's ok for now.

jhdcs (approved):
I agree that a `pathlib`-based solution would be a good upgrade to do at some point. Doesn't have to be now, though.

Issue #225, opened 2022-07-13
colcon build fails with error "failed to resolve patches for `https://github.com/rust-lang/crates.io-index`"


VRichardJP:
I am folloowing the build documentation using the foxy docker image (https://github.com/ros2-rust/ros2_rust/blob/main/docs/building.md#building-with-colcon)

when I run  `colcon build` to build all the default packages, I get the following error:
```
root@6f6e313c6d9a:/workspace# colcon build
Starting >>> rosidl_runtime_rs
Starting >>> test_interface_files
Finished <<< rosidl_runtime_rs [0.09s]
Starting >>> rosidl_generator_rs
Finished <<< test_interface_files [0.10s]                                                            
Finished <<< rosidl_generator_rs [0.15s]                   
Starting >>> rosidl_default_runtime
Starting >>> rosidl_default_generators
Finished <<< rosidl_default_runtime [0.09s]                                   
Finished <<< rosidl_default_generators [0.09s]
Starting >>> builtin_interfaces
Starting >>> unique_identifier_msgs
Starting >>> rclrs_example_msgs
Starting >>> std_srvs
Starting >>> lifecycle_msgs
Finished <<< unique_identifier_msgs [0.48s]                                                                   
Finished <<< builtin_interfaces [0.52s]
Starting >>> std_msgs
Starting >>> action_msgs
Starting >>> rcl_interfaces
Starting >>> rosgraph_msgs
Starting >>> statistics_msgs
Finished <<< std_srvs [0.55s]                                                                          
Finished <<< rclrs_example_msgs [0.77s]                                                                      
Finished <<< lifecycle_msgs [0.98s]                                                               
Finished <<< rosgraph_msgs [0.53s]                                                                
Finished <<< action_msgs [0.72s]                                                                
Starting >>> example_interfaces
Starting >>> test_msgs
Finished <<< statistics_msgs [0.73s]                                                                 
Finished <<< rcl_interfaces [1.52s]                                                                    
Starting >>> composition_interfaces
Finished <<< std_msgs [1.66s]                                                                          
Starting >>> geometry_msgs
Starting >>> rclrs
Starting >>> actionlib_msgs
--- stderr: rclrs                                                                                          
error: failed to resolve patches for `https://github.com/rust-lang/crates.io-index`

Caused by:
  failed to load source for dependency `builtin_interfaces`

Caused by:
  Unable to update /workspace/install/builtin_interfaces/share/builtin_interfaces/rust

Caused by:
[...]

PR #229, opened 2022-07-18
use latest rust toolchain in docker

VRichardJP:
fixes https://github.com/ros2-rust/ros2_rust/issues/225


Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN apt-get update && apt-get install -y \
     && rm -rf /var/lib/apt/lists/*
 
 # Install Rust and the cargo-ament-build plugin
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain 1.59.0 -y
```
esteve, line 14: Instead of leaving out the version of the toolchain, I'd update it to the latest one so we know for sure which version is the one in the toolchain. Could you update it to 1.62.0? Thanks.

esteve (changes requested):

esteve:
@VRichardJP thanks for the PR! I only have one comment.

esteve (approved):
@VRichardJP excellent! Thanks :slightly_smiling_face:

PR #230, opened 2022-07-21
Declare fn new in waitables as pub(crate)

esteve:
This PR restricts the visibility of `fn new` in waitables (`Subscription`, `Client`, `Service`) so that only other code in the crate can access it. This prevents users from instantiating these structures and therefore we ensure that the only way of creating them is via the `Node::create_X` methods so that any internal logic is not bypassed.

### DIFF
--- a/rclrs/src/node.rs
+++ b/rclrs/src/node.rs
@@ -184,11 +184,17 @@ impl Node {
     ///
     /// [1]: crate::Client
     // TODO: make client's lifetime depend on node's lifetime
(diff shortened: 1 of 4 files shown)


jhdcs (approved):
I'm not seeing any issues.

nnmm (approved):
Looks good to me, personally I'd add a note to the documentation of each waitable that this entity is created by `Node::create_x`, and the rationale for making it `pub(crate)` to `X::new()`,  but I'll leave that up to you.

esteve:
@jhdcs @nnmm I've pushed fcae94a explaining the rationale. Thanks for your reviews.

Issue #123, opened 2022-04-22
Add support for graph


nnmm:
See https://github.com/ros2/rcl/blob/master/rcl/include/rcl/graph.h

Since this is a complex feature, it would be good to have some prior experience with the code base and/or seek early-stage feedback from other contributors.

nnmm:
Closed by #233.

PR #234, opened 2022-07-26
Add Graph Support

christopherjreid:
This PR adds support for the functionality implemented in rcl/graph.h,
as requested by #123

### DIFF
--- a/rclrs/src/node/graph.rs
+++ b/rclrs/src/node/graph.rs
@@ -1,0 +1,553 @@
+use std::collections::HashMap;
+use std::ffi::{CStr, CString};
+use std::slice;
+
+use crate::rcl_bindings::*;
+use crate::{Node, RclrsError, ToResult};
+
+impl Drop for rmw_names_and_types_t {
+    fn drop(&mut self) {
+        // SAFETY: No preconditions for this function.
+        unsafe {
+            rcl_names_and_types_fini(self).ok().unwrap();
+        }
+    }
+}
+
+impl Drop for rmw_topic_endpoint_info_array_t {
+    fn drop(&mut self) {
+        // SAFETY: No preconditions for this function.
+        unsafe {
+            rmw_topic_endpoint_info_array_fini(self, &mut rcutils_get_default_allocator())
+                .ok()
+                .unwrap();
+        }
+    }
+}
+
+impl Drop for rcutils_string_array_t {
+    fn drop(&mut self) {
+        // SAFETY: No preconditions for this function.
+        unsafe {
+            rcutils_string_array_fini(self);
+        }
+    }
+}
+
+/// Stores a list of types associated with each topic.
+pub type TopicNamesAndTypes = HashMap<String, Vec<String>>;
+
+/// Stores a node's name and namespace
+#[derive(Debug, PartialEq)]
+pub struct NodeNameInfo {
+    /// The name of the node
+    pub name: String,
+    /// The namespace of the node
+    pub namespace: String,
+}
+
+/// Contains topic endpoint information
+pub struct TopicEndpointInfo {
+    /// The name of the endpoint node
+    pub node_name: String,
+    /// The namespace of the endpoint node
+    pub node_namespace: String,
+    /// The type of the topic
+    pub topic_type: String,
+}
+
+impl Node {
+    /// Returns a list of topic names and types for publishers associated with a node.
+    pub fn get_publisher_names_and_types_by_node(
+        &self,
+        node: &str,
+        namespace: &str,
+    ) -> Result<TopicNamesAndTypes, RclrsError> {
+        // SAFETY: Forwarding arguments to the inner C function is safe
+        unsafe extern "C" fn wrapper(
+            node: *const rcl_node_t,
+            allocator: *mut rcl_allocator_t,
+            node_name: *const ::std::os::raw::c_char,
+            node_namespace: *const ::std::os::raw::c_char,
+            topic_names_and_types: *mut rcl_names_and_types_t,
+        ) -> rcl_ret_t {
+            rcl_get_publisher_names_and_types_by_node(
+                node,
+                allocator,
+                false,
+                node_name,
+                node_namespace,
+                topic_names_and_types,
+            )
+        }
+
+        self.get_names_and_types_by_node(node, namespace, wrapper)
+    }
+
+    /// Returns a list of topic names and types for subscriptions associated with a node.
+    pub fn get_subscription_names_and_types_by_node(
+        &self,
+        node: &str,
+        namespace: &str,
+    ) -> Result<TopicNamesAndTypes, RclrsError> {
+        // SAFETY: Forwarding arguments to the inner C function is safe
+        unsafe extern "C" fn wrapper(
+            node: *const rcl_node_t,
+            allocator: *mut rcl_allocator_t,
+            node_name: *const ::std::os::raw::c_char,
+            node_namespace: *const ::std::os::raw::c_char,
+            topic_names_and_types: *mut rcl_names_and_types_t,
+        ) -> rcl_ret_t {
+            rcl_get_subscriber_names_and_types_by_node(
+                node,
+                allocator,
+                false,
+                node_name,
+                node_namespace,
+                topic_names_and_types,
+            )
+        }
+
+        self.get_names_and_types_by_node(node, namespace, wrapper)
+    }
+
+    /// Returns a list of topic names and types for services associated with a node.
+    pub fn get_service_names_and_types_by_node(
+        &self,
+        node: &str,
+        namespace: &str,
+    ) -> Result<TopicNamesAndTypes, RclrsError> {
+        self.get_names_and_types_by_node(node, namespace, rcl_get_service_names_and_types_by_node)
+    }
+
+    /// Returns a list of topic names and types for clients associated with a node.
+    pub fn get_client_names_and_types_by_node(
+        &self,
+        node: &str,
+        namespace: &str,
+    ) -> Result<TopicNamesAndTypes, RclrsError> {
+        self.get_names_and_types_by_node(node, namespace, rcl_get_client_names_and_types_by_node)
+    }
+
+    /// Returns a list of all topic names and their types.
+    pub fn get_topic_names_and_types(&self) -> Result<TopicNamesAndTypes, RclrsError> {
+        // SAFETY: Getting a zero-initialized value is always safe
+        let mut rcl_names_and_types = unsafe { rmw_get_zero_initialized_names_and_types() };
+
+        // SAFETY: rcl_names_and_types is zero-initialized as expected by this call
+        unsafe {
+            rcl_get_topic_names_and_types(
+                &*self.rcl_node_mtx.lock().unwrap(),
+                &mut rcutils_get_default_allocator(),
+                false,
+                &mut rcl_names_and_types,
+            )
+            .ok()?
+        };
+
+        Ok(convert_names_and_types(rcl_names_and_types))
+    }
+
+    /// Returns a list of service names and types for this node.
+    pub fn get_service_names_and_types(&self) -> Result<TopicNamesAndTypes, RclrsError> {
+        self.get_service_names_and_types_by_node(&self.name(), &self.namespace())
+    }
+
+    /// Returns a list of all node names.
+    pub fn get_node_names(&self) -> Result<Vec<NodeNameInfo>, RclrsError> {
+        // SAFETY: Getting a zero-initialized value is always safe
+        let (mut rcl_names, mut rcl_namespaces) = unsafe {
+            (
+                rcutils_get_zero_initialized_string_array(),
+                rcutils_get_zero_initialized_string_array(),
+            )
+        };
+
+        // SAFETY: node_names and node_namespaces are zero-initialized as expected by this call.
+        unsafe {
+            rcl_get_node_names(
+                &*self.rcl_node_mtx.lock().unwrap(),
+                rcutils_get_default_allocator(),
+                &mut rcl_names,
+                &mut rcl_namespaces,
+            )
+            .ok()?;
+        };
+
+        // SAFETY: Because the previous function call successfully returned, the names and
+        // namespaces are populated with valid data
+        let (names_slice, namespaces_slice) = unsafe {
+            (
+                slice::from_raw_parts(rcl_names.data, rcl_names.size),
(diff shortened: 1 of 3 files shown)


nnmm:
Hi @christopherjreid, from a first glance this looks very good. However, I find the `Graph` trait a bit odd, why not just implement the methods on the node directly?

christopherjreid:
> Hi @christopherjreid, from a first glance this looks very good. However, I find the `Graph` trait a bit odd, why not just implement the methods on the node directly?

That's a good question, I was just thinking about it from the "don't use the trait if you don't need to" perspective.
Would you rather:

- remove graph.rs, pull everything in the trait impl directly into node.rs, or
- keep graph.rs, but remove the trait, and change the trait impl to just `impl Node`, or
- add a `Graph` struct, put the impl on that, and add that as a member to `Node`, with a `get_graph` function in the `Node` impl?

nnmm:
> keep graph.rs, but remove the trait, and change the trait impl to just `impl Node`

That would be my preferred way to do it.

christopherjreid:
Comments have been addressed, and all checks currently pass!

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,568 @@
...
+    ) -> Result<TopicNamesAndTypes, RclrsError> {
+        self.get_names_and_types_by_node_with_demangle_arg(
+            node,
+            namespace,
+            rcl_get_publisher_names_and_types_by_node,
+        )
+    }
+
+    /// Get a list of topic names and types for subscribers associated with a node
+    pub fn get_subscriber_names_and_types_by_node(
```
nnmm, line 67: Weird that this is called `subscribers` in rcl and not `subscriptions`. Could you check in `rclcpp` which one they use?

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,568 @@
...
+            .collect();
+
+        Ok(topic_endpoint_infos_vec)
+    }
+}
+
+/// Convert a rmw_names_and_types_t object to a HashMap
+fn convert_names_and_types(
+    rcl_names_and_types: rmw_names_and_types_t,
+) -> Result<HashMap<String, Vec<String>>, RclrsError> {
```
nnmm, line 387: Nit: this function doesn't need to return a `Result`, so I'd prefer to add `Ok()` to each caller.

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,568 @@
...
+impl Drop for rcutils_string_array_t {
+    fn drop(&mut self) {
+        // SAFETY: No preconditions for this function
+        unsafe {
+            rcutils_string_array_fini(self);
+        }
+    }
+}
+
+/// Topic names and types
```
nnmm, line 37: To keep the doc comment style consistent, could you make sure that all doc comments are in the [declarative mood and end with a period](https://github.com/ros2-rust/ros2_rust/blob/main/docs/CONTRIBUTING.md#documentation)?

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,568 @@
...
+
+        convert_names_and_types(rcl_names_and_types)
+    }
+
+    /// Get a list of service names and types for this node
+    pub fn get_service_names_and_types(&self) -> Result<TopicNamesAndTypes, RclrsError> {
+        self.get_service_names_and_types_by_node(&self.name(), &self.namespace())
+    }
+
+    /// Get a list of all node names
```
nnmm, line 128: Would be good to document what the return type means (namespaces + node names) – or create a new struct for it.

rclrs/src/node/graph.rs
nnmm: If you want to go the extra mile, you could also add examples and more detailed information. E.g. for this type, we could explain what the key and the value is, and why there might be multiple types per topic.

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,568 @@
...
+
+    /// Get topic subscriber info
+    pub fn get_subscribers_info_by_topic(
+        &self,
+        topic: &str,
+    ) -> Result<Vec<TopicEndpointInfo>, RclrsError> {
+        self.get_publisher_subscriber_info_by_topic(topic, rcl_get_subscriptions_info_by_topic)
+    }
+
+    /// Get an rcl names_and_types function, with a "demangle" argument
```
nnmm, line 263: Shouldn't this be "no_demangle"?

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,568 @@
...
+            *const rcl_node_t,
+            *mut rcl_allocator_t,
+            bool,
+            *const ::std::os::raw::c_char,
+            *const ::std::os::raw::c_char,
+            *mut rcl_names_and_types_t,
+        ) -> rcl_ret_t,
+    ) -> Result<TopicNamesAndTypes, RclrsError> {
+        let mut rcl_names_and_types = unsafe { rmw_get_zero_initialized_names_and_types() };
+        let node_name = CString::new(node).unwrap();
```
nnmm, line 278: For conversion to C string types, we're using this error handling approach: https://github.com/ros2-rust/ros2_rust/blob/3194b772762c5ba410e97d838930fbd690e7b0db/rclrs/src/node/builder.rs#L236

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,568 @@
...
+        let node_namespace = CString::new(namespace).unwrap();
+        unsafe {
+            getter(
+                &*self.rcl_node_mtx.lock() as *const _,
```
nnmm, line 312: I think you might need to rebase onto main, this now should require `.lock().unwrap()` since we switched to std lib mutexes.

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,568 @@
...
+            *const rcl_node_t,
+            *mut rcl_allocator_t,
+            *const ::std::os::raw::c_char,
+            bool,
+            *mut rcl_topic_endpoint_info_array_t,
+        ) -> rcl_ret_t,
+    ) -> Result<Vec<TopicEndpointInfo>, RclrsError> {
+        let topic = CString::new(topic).unwrap();
+        let mut rcl_publishers_info =
+            unsafe { rmw_get_zero_initialized_topic_endpoint_info_array() };
```
nnmm, line 337: Please add a `// SAFETY` comment to each `unsafe` block. E.g. for these `rmw_get_zero_initialized_*` calls, we've used

```
// SAFETY: Getting a zero-initialized value is always safe.
```
in other places.

nnmm (commented):
I did a detailed review – mostly style comments, looks good overall.

rclrs/src/node/graph.rs
christopherjreid: https://github.com/ros2/rcl/blob/287ccd9ed06ff5bdded4dfb1130920d592a71bb7/rcl/include/rcl/graph.h#L142 calls it "subscribers", 
rclcpp does not seem to provide a wrapper for this function. I guess that raises the question - should I focus on "wrapping" the rcl functions, or imitating the rclcpp implementation?

christopherjreid:
Comments are addressed, except for my question about the differences between the rcl API, and the rclcpp wrapper around that API, and which one this should follow

rclrs/src/node/graph.rs
nnmm: Yeah, I saw that it's called `subscribers` in that file, but that seems like a mistake to me, and I though maybe it seemed that way to the rclcpp authors too. But if they don't have that function, I'd say we can just call it `get_subscription_names_and_types_by_node()`.

We can provide functionality that rclcpp doesn't provide, so I'd say the focus is to cover the `rcl` API.

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,635 @@
...
+    /// The namespace of the endpoint node
+    pub node_namespace: String,
+    /// The type of the topic
+    pub topic_type: String,
+    /// The RMW endpoint type of the topic
+    pub endpoint_type: rmw_endpoint_type_t,
+}
+
+impl Node {
+    /// Get a list of topic names and types for publishers associated with a node.
```
nnmm, line 62: Could you replace "Get" by "Returns" everywhere, to avoid the imperative mood?

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,635 @@
...
+/// Contains topic endpoint information
+pub struct TopicEndpointInfo {
+    /// The name of the endpoint node
+    pub node_name: String,
+    /// The namespace of the endpoint node
+    pub node_namespace: String,
+    /// The type of the topic
+    pub topic_type: String,
+    /// The RMW endpoint type of the topic
+    pub endpoint_type: rmw_endpoint_type_t,
```
nnmm, line 58: Could we make our own small enum for this? This would have the benefit that we don't expose the "Invalid" variant to the user.

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,635 @@
...
+        self.get_names_and_types_by_node_without_demangle_arg(
+            node,
+            namespace,
+            rcl_get_client_names_and_types_by_node,
+        )
+    }
+
+    /// Get a list of all topic names and their types.
+    pub fn get_topic_names_and_types(&self) -> Result<TopicNamesAndTypes, RclrsError> {
+        let mut rcl_names_and_types = unsafe { rmw_get_zero_initialized_names_and_types() };
```
nnmm, line 116:
```suggestion
        // SAFETY: Getting a zero-initialized value is always safe.
        let mut rcl_names_and_types = unsafe { rmw_get_zero_initialized_names_and_types() };
```

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,635 @@
...
+
+        // SAFETY: rcl_names_and_types is zero-initialized as expected by this call
+        unsafe {
+            rcl_get_topic_names_and_types(
+                &*self.rcl_node_mtx.lock().unwrap() as *const _,
+                &mut rcutils_get_default_allocator() as *mut _,
+                false,
+                &mut rcl_names_and_types as *mut _,
```
nnmm, line 124: Casting references to raw pointers should not be needed, and we generally avoid it.

```suggestion
                &*self.rcl_node_mtx.lock().unwrap(),
                &mut rcutils_get_default_allocator(),
                false,
                &mut rcl_names_and_types,
```

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,635 @@
...
+    pub fn get_node_names(&self) -> Result<Vec<NodeNameInfo>, RclrsError> {
+        // SAFETY: Getting a zero-initialized value is always safe
+        let (mut rcl_names, mut rcl_namespaces) = unsafe {
+            (
+                rcutils_get_zero_initialized_string_array(),
+                rcutils_get_zero_initialized_string_array(),
+            )
+        };
+
+        // SAFETY: node_names and node_namespaces are zero-initialized as expected by this call.
```
nnmm, line 147:
```suggestion
        // SAFETY: rcl_names and rcl_namespaces are zero-initialized as expected by this call.
```

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,635 @@
...
+        };
+
+        // SAFETY: The node_names, namespaces, and enclaves are zero-initialized as expected by this call.
+        unsafe {
+            rcl_get_node_names_with_enclaves(
+                &*self.rcl_node_mtx.lock().unwrap() as *const _,
+                rcutils_get_default_allocator(),
+                &mut rcl_names as *mut _,
+                &mut rcl_namespaces as *mut _,
+                &mut rcl_enclaves as *mut _,
```
nnmm, line 203:
```suggestion
                &*self.rcl_node_mtx.lock().unwrap(),
                rcutils_get_default_allocator(),
                &mut rcl_names,
                &mut rcl_namespaces,
                &mut rcl_enclaves,
```

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,635 @@
...
+            unsafe { rmw_get_zero_initialized_topic_endpoint_info_array() };
+
+        // SAFETY: topic has been zero-initialized
+        unsafe {
+            getter(
+                &*self.rcl_node_mtx.lock().unwrap(),
+                &mut rcutils_get_default_allocator() as *mut _,
+                topic.as_ptr(),
+                true,
+                &mut rcl_publishers_info as *mut _,
```
nnmm, line 401:
```suggestion
                &*self.rcl_node_mtx.lock().unwrap(),
                &mut rcutils_get_default_allocator(),
                topic.as_ptr(),
                true,
                &mut rcl_publishers_info,
```

rclrs/src/node/graph.rs
```diff
@@ -0,0 +1,635 @@
...
+                node_namespace.as_ptr(),
+                &mut rcl_names_and_types,
+            )
+        };
+
+        Ok(convert_names_and_types(rcl_names_and_types))
+    }
+
+    /// Get an rcl names_and_types function, without a "no_demangle" argument.
+    fn get_names_and_types_by_node_without_demangle_arg(
```
nnmm, line 334: Do you think it would be worth unifying the version with a demangle arg with this function, à la https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b173ae6ee24b818a94bac23c6d9142b9?

nnmm (commented):
I think this was the final iteration :+1:

christopherjreid:
Updated!

rclrs/src/node/graph.rs
nnmm: I meant an enum that has no `Invalid` variant. I think it is an error for the `rmw` layer to return an enum with this value, so we could panic if such a variant is encountered or create a new error variant in `RclrsError` for this.

Actually, thinking about this a litte more, do we even need to report the endpoint type to the user? It should be clear to the caller anyway – if they called `get_publishers_info_by_topic()`, it will be publishers, obviously. So maybe the best solution is to just eliminate this field. Do you agree?

Sorry for the back-and-forth.

christopherjreid: Don't apologize, I'm glad this is getting refined!

I poked around the rcl and rmw source a bit - I think you're right, at our layer, we should never expect that field to be INVALID, and you'll know which of the remaining types it is based on the function you just called. 
So, I'll just remove it!

nnmm (approved):
Congrats!

Issue #185, opened 2022-05-31
Support Box<T: Message> argument type to subscription callbacks 


nnmm:
We should also allow sending and receiving `Box<T>` where `T: Message`, in addition to `T` (and, in the case of publishers, `&T`) – sometimes that's more convenient or efficient.

For publishers, it would be straightforward: Change the `Cow` to something that also has a variant for boxed values.

For subscriptions, I don't know if we can somehow derive the concrete type (`Box<T>` vs `T`) from the callback given. But one way that would work is to give subscriptions an additional generic parameter for the concrete type.

LaVieEstDure:
Currently `Subscription::new` takes `callback: F` where `F: FnMut(T) + 'static + Send` and `T: Message`. We can have `Box<MsgType>` still fit this if we `impl<T: Message> Message for Box<T>` and forward the behaviour. If no one else is working on this, I can have a crack at this

PR #235, opened 2022-07-26
Implement Message for Box<T: Message> 

LaVieEstDure:
Fix for #185. Implements the Message trait for Box<T: Message> to allow for Boxed publishers and subscribers. 
You should now be able to do 
```rust
    let publisher =
        node.create_publisher::<Box<std_msgs::msg::String>>("topic", rclrs::QOS_PROFILE_DEFAULT)?;
    let mut message = Box::new(std_msgs::msg::String::default());
    publisher.publish(&message)?;
```
and 
```rust
    let _subscription = node.create_subscription::<Box<std_msgs::msg::String>, _>(
        "topic",
        rclrs::QOS_PROFILE_DEFAULT,
        move |msg: Box<std_msgs::msg::String>| {
            num_messages += 1;
            println!("I heard boxed: '{}'", (*msg).data);
            println!("(Got {} messages so far)", num_messages);
        },
    )?;
```
It's been 3 years since I last properly wrote Rust so I'm not sure my Cow handling is idiomatic. Also not sure if the trait implementation is in an appropriate location.

### DIFF
--- a/rosidl_runtime_rs/src/traits.rs
+++ b/rosidl_runtime_rs/src/traits.rs
@@ -144,6 +144,19 @@ pub trait Message: Clone + Debug + Default + 'static + Send + Sync {
     fn from_rmw_message(msg: Self::RmwMsg) -> Self;
 }
 
+/// Allows boxed values to be used by users in Publishers and Subscribers
+impl<T: Message> Message for Box<T> {
+    type RmwMsg = T::RmwMsg;
+
+    fn into_rmw_message(msg_cow: Cow<'_, Self>) -> Cow<'_, Self::RmwMsg> {
+        T::into_rmw_message(Cow::Owned(*msg_cow.into_owned()))
+    }
+
+    fn from_rmw_message(msg: Self::RmwMsg) -> Self {
+        Box::new(T::from_rmw_message(msg))
+    }
+}
+
 /// Trait for services.
 ///
 /// User code never needs to call this trait's method, much less implement this trait.
--- a/rosidl_runtime_rs/src/traits.rs
+++ b/rosidl_runtime_rs/src/traits.rs
@@ -144,7 +144,8 @@ pub trait Message: Clone + Debug + Default + 'static + Send + Sync {
     fn from_rmw_message(msg: Self::RmwMsg) -> Self;
 }
 
-/// Allows boxed values to be used by users in Publishers and Subscribers
+/// Allows boxed values to be used by users in Publishers and Subscriptions.
+// See https://github.com/ros2-rust/ros2_rust/pull/235 for a discussion of possible alternatives.
 impl<T: Message> Message for Box<T> {
     type RmwMsg = T::RmwMsg;
 


nnmm:
Hi @LaVieEstDure, this works, let me just think aloud a little bit whether this is the best way.

The nicest solution imo would be one where the user can simply pass in either a `T` callback or a `Box<T>` callback when constructing a `Subscription<T>`. But I don't know how that could be made to work, you [can't write a trait that abstracts over both forms](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dcd472bdccbca34905e33d3b3e81479f). Can you think of a way to do this?

With this impl, you can now also construct `Publisher<Box<T>>`, but for publishers I'd prefer to integrate the `Box` variant into the `MessageCow` mechanism. With the `MessageCow` mechanism, the user can decide for each individual message whether it should be published by reference or owned (or as a `Box`, if we add that), but with the `Message` mechanism, you have to decide at the beginning, when creating the publisher, whether you're ever going to publish `Box`es or not.

Another possibility would be always taking a `Box<T>` in callbacks. That has the drawback of being a bit more verbose/unergonomic and costing a heap allocation each time (not sure if that matters though).

What I like about the solution in this PR is that it keeps things simple – no extra traits.

So I'm inclined to say we go forward with this change, but I'd like to see if you or @esteve and @jhdcs see a way to get an optimal solution. BTW, your use of `Cow` and the location is fine :)

LaVieEstDure:
> But I don't know how that could be made to work, you [can't write a trait that abstracts over both forms](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dcd472bdccbca34905e33d3b3e81479f). Can you think of a way to do this?

I think the demo there is the wrong approach. You're trying to specialise a Fn trait which Rust doesnt support. Would it work if instead of having `callback: CallbackTrait` which we have to specialise, we had `callback: Fn(CallbackArgTrait)` and we implemented CallbackArgTrait for all generated Message types which forwarded Message function implementations, and manually implemented the CallbackArgTrait for Box where the associated type of the implementation was the type of what the box was holding

jhdcs:
If I remember correctly from ROS 1, you had to specify whether a subscription was for a message or a constant pointer to said message, right? If that's the case, then this PR is bringing back functionality from ROS 1 that could be useful for when a `Box` is needed.

I feel like we can probably look at reviewing this PR as-is, and revisiting it later if/when we discover that optimizations are needed.

> Another possibility would be always taking a `Box<T>` in callbacks. That has the drawback of being a bit more verbose/unergonomic and costing a heap allocation each time (not sure if that matters though).

Personally, I would vote against this approach. I feel like it adds a "hidden tax" on the user by forcing them to make more heap allocations. This may reduce performance, and prevent `rclrs` from being used in the sort of high-speed high-throughput applications that Rust is normally good at.

LaVieEstDure:
One option, that is a bit more complicated than just simply implementing Message for Box is adding a second generic parameter over Subscription and constraining them by their associated type
```rust
impl<T, Arg> Subscription<T, Arg = T>
where
    T: Message,
    Arg: Message<RmwMsg = T::RmwMsg>
{
    /// Creates a new subscription.
    pub(crate) fn new<F>(
        node: &Node,
        topic: &str,
        qos: QoSProfile,
        callback: F,
    ) -> Result<Self, RclrsError>
    // This uses pub(crate) visibility to avoid instantiating this struct outside
    // [`Node::create_subscription`], see the struct's documentation for the rationale
    where
        T: Message,
        Arg: Message<RmwMsg = T::RmwMsg>,
        F: FnMut(Arg) + 'static + Send,
```
Which I think _should_ work and I _think_ you'll be able to change node to 
```rust
    pub fn create_subscription<T, Arg, F>(
        &mut self,
        topic: &str,
        qos: QoSProfile,
        callback: F,
    ) -> Result<Arc<Subscription<T>>, RclrsError>
    where
        T: Message,
        Arg: Message<RmwMsg = T::RmwMsg>,
        F: FnMut(Arg) + 'static + Send,
    {
        let subscription = Arc::new(Subscription::<T, Arg>::new(self, topic, qos, callback)?);
        self.subscriptions
            .push(Arc::downgrade(&subscription) as Weak<dyn SubscriptionBase>);
        Ok(subscription)
    }
```
Or if we're trying real hard, we might be able to extend the Message trait to implement an UnderlyingMessage associated type, which is set to the _same_ type for Message types and to the contained type for containers, and do
```rust
        let subscription = Arc::new(Subscription::<Arg::UnderlyingMsg, Arg>::new(self, topic, qos, callback)?);
```
But now we're getting into uglier territory. Feel free point out where I'm dumb and forgetting something that'll make make the compiler reee at me or if there's a simpler solution to what I just proposed

LaVieEstDure:
Also regarding
> If I remember correctly from ROS 1, you had to specify whether a subscription was for a message or a constant pointer to said message, right? If that's the case, then this PR is bringing back functionality from ROS 1 that could be useful for when a Box is needed.

I think all the messages were passed around as const shared pointers. The template specification didnt change anything, it was just improved over time to be more hygienic to write. Subscriber<Msg> still took a callback that had took, in effect, a Boost const shared ptr but you could pass in a callback that took const Msg& and it turned implicitly into that semantically equivalent type.

rosidl_runtime_rs/src/traits.rs
```diff
@@ -144,6 +144,19 @@ pub trait Message: Clone + Debug + Default + 'static + Send + Sync {
     fn from_rmw_message(msg: Self::RmwMsg) -> Self;
 }
 
+/// Allows boxed values to be used by users in Publishers and Subscribers
```
nnmm, line 147:
```suggestion
/// Allows boxed values to be used by users in Publishers and Subscriptions.
// See https://github.com/ros2-rust/ros2_rust/pull/235 for a discussion of possible alternatives.
```

nnmm:
> I think the demo there is the wrong approach. You're trying to specialise a Fn trait which Rust doesnt support. Would it work if instead of having `callback: CallbackTrait` which we have to specialise, we had `callback: Fn(CallbackArgTrait)` and we implemented CallbackArgTrait for all generated Message types which forwarded Message function implementations, and manually implemented the CallbackArgTrait for Box where the associated type of the implementation was the type of what the box was holding

That's definitely better, but it also has its problems: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b5d797bbde4864f45b0e9a33635e8b22

Another approach would be to simply have two functions, `create_subscription()` and `create_subscription_for_heap_message()` or something. Then we can store the callback internally as 
```
enum Callback {
    Regular: Box<dyn Fn(T) + 'static + Send>,
    Heap: Box<dyn Fn(Box<T>) + 'static + Send>
}
```

But I don't like that so much either. So, let's go ahead with this PR, and revisit it when needed. I have just one small change to request.

nnmm (approved):
Thanks for your patience :)

Issue #237, opened 2022-07-27
Create new unit test package


nnmm:
We need a unit test package that can have dependencies on message packages (unlike `rclrs` itself).

Tests that do not need a message type should still live in `rclrs`.

PR #239, opened 2022-07-30
Add `rclrs_test` package for unit/integration tests

christopherjreid:
This package is meant to test behaviors with dependencies that don't
belong in the rclrs crate - namely, msgs

Addresses #237

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -95,7 +95,7 @@ jobs:
     - name: Run cargo test on Rust packages
       run: |
         cd ${{ steps.build.outputs.ros-workspace-directory-name }}
-        . /opt/ros/${{ matrix.ros_distribution }}/setup.sh
+        . install/setup.sh
         for path in $(colcon list | awk '$3 == "(ament_cargo)" && $1 != "examples_rclrs_minimal_pub_sub" && $1 != "examples_rclrs_minimal_client_service" { print $2 }'); do
         cd $path
         echo "Running cargo test in $path"
--- a/rclrs_tests/Cargo.toml
+++ b/rclrs_tests/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "rclrs_tests"
+version = "0.2.0"
+authors = ["Chris Reid <christopherjreid@outlook.com>"]
+edition = "2021"
+
+[lib]
+path = "src/lib.rs"
+
+[dependencies]
+anyhow = {version = "1", features = ["backtrace"]}
+test_msgs = {version = "*"}
+
+[dependencies.rclrs]
+version = "*"
+
+[dependencies.rosidl_runtime_rs]
+version = "*"
--- a/rclrs_tests/package.xml
+++ b/rclrs_tests/package.xml
@@ -1,0 +1,19 @@
+<?xml version="2.0"?>
+<?xml-model
+   href="http://download.ros.org/schema/package_format3.xsd"
+   schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>rclrs_tests</name>
+  <version>0.2.0</version>
+  <description>Package containing tests for rclrs that make use of msgs and other dependencies</description>
+  <maintainer email="christopherjreid@outlook.com">Chris Reid</maintainer>
+  <license>Apache License 2.0</license>
+
+  <depend>test_msgs</depend>
+  <depend>rclrs</depend>
+  <depend>rosidl_runtime_rs</depend>
+
+  <export>
+    <build_type>ament_cargo</build_type>
+  </export>
+</package>
--- a/rclrs_tests/src/graph_tests.rs
+++ b/rclrs_tests/src/graph_tests.rs
@@ -1,0 +1,238 @@
+use rclrs::{
+    Context, Node, NodeBuilder, RclrsError, TopicNamesAndTypes, QOS_PROFILE_SYSTEM_DEFAULT,
+};
+
+use test_msgs::{msg, srv};
+
+struct TestGraph {
+    node1: Node,
+    node2: Node,
+}
+
+fn construct_test_graph(namespace: &str) -> Result<TestGraph, RclrsError> {
+    let context = Context::new([])?;
+    Ok(TestGraph {
+        node1: NodeBuilder::new(&context, "graph_test_node_1")
+            .namespace(namespace)
+            .build()?,
+        node2: NodeBuilder::new(&context, "graph_test_node_2")
+            .namespace(namespace)
+            .build()?,
+    })
+}
+
+#[test]
+fn test_publishers() -> Result<(), RclrsError> {
+    let graph = construct_test_graph("test_publishers_graph")?;
+
+    let _node_1_empty_publisher = graph
+        .node1
+        .create_publisher::<msg::Empty>("graph_test_topic_1", QOS_PROFILE_SYSTEM_DEFAULT)?;
+    let _node_1_basic_types_publisher = graph
+        .node1
+        .create_publisher::<msg::BasicTypes>("graph_test_topic_2", QOS_PROFILE_SYSTEM_DEFAULT)?;
+    let _node_2_default_publisher = graph
+        .node2
+        .create_publisher::<msg::Defaults>("graph_test_topic_3", QOS_PROFILE_SYSTEM_DEFAULT)?;
+
+    std::thread::sleep(std::time::Duration::from_millis(100));
+
+    assert_eq!(
+        graph
+            .node1
+            .count_publishers(&(graph.node1.namespace() + "/graph_test_topic_1"))?,
+        1
+    );
+    assert_eq!(
+        graph
+            .node1
+            .count_publishers(&(graph.node1.namespace() + "/graph_test_topic_2"))?,
+        1
+    );
+    let node_1_publisher_names_and_types = graph
+        .node1
+        .get_publisher_names_and_types_by_node(&graph.node1.name(), &graph.node1.namespace())?;
+
+    let types = node_1_publisher_names_and_types
+        .get(&(graph.node1.namespace() + "/graph_test_topic_1"))
+        .unwrap();
+    assert!(types.contains(&"test_msgs/msg/Empty".to_string()));
+
+    let types = node_1_publisher_names_and_types
+        .get(&(graph.node1.namespace() + "/graph_test_topic_2"))
+        .unwrap();
+    assert!(types.contains(&"test_msgs/msg/BasicTypes".to_string()));
+
+    let node_2_publisher_names_and_types = graph
+        .node2
+        .get_publisher_names_and_types_by_node(&graph.node2.name(), &graph.node2.namespace())?;
+
+    let types = node_2_publisher_names_and_types
+        .get(&(graph.node2.namespace() + "/graph_test_topic_3"))
+        .unwrap();
+    assert_eq!(
+        graph
+            .node2
+            .count_publishers(&(graph.node2.namespace() + "/graph_test_topic_3"))?,
+        1
+    );
+    assert!(types.contains(&"test_msgs/msg/Defaults".to_string()));
+
+    Ok(())
+}
+
+#[test]
+fn test_subscriptions() -> Result<(), RclrsError> {
+    let mut graph = construct_test_graph("test_subscriptions_graph")?;
+
+    let _node_1_defaults_subscription = graph.node1.create_subscription::<msg::Defaults, _>(
+        "graph_test_topic_3",
+        QOS_PROFILE_SYSTEM_DEFAULT,
+        |_msg| {},
+    )?;
+    let _node_2_empty_subscription = graph.node2.create_subscription::<msg::Empty, _>(
+        "graph_test_topic_1",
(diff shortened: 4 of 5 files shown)


christopherjreid:
@nnmm Is this package structure what you envisioned for #237?

christopherjreid:
Based on previous feedback, I'm wondering about structuring this as 

- Setup a graph with a couple of nodes, publishers, subscribers, services and clients
- Run tests for all the graph.rs stuff

Unfortunately it doesn't look like Rust supports a gtest-like `SetUp()` and `TearDown()`, so that means (I think) that this would all go into a single test. Does that make sense to do?

nnmm:
Thanks for tackling this!

> @nnmm Is this package structure what you envisioned for #237?

Yep. A few early-stage comments:
- We probably want `test_msgs` rather than `std_msg` for this package. Though I just compiled it and noticed that we have a bug in `rosidl_runtime_rs` or `rosidl_generator_rs` that we need to fix first :/
- Returning `Result` from the tests makes them a little more concise.
- You don't need `mod tests` inside `graph_tests.rs`. You can just put `#[cfg(test)]` directly on the modules in `lib.rs`. Maybe it's even possible to just add a single `#![cfg(test)]` (with an exclamation mark) to the top `lib.rs` instead.

> Unfortunately it doesn't look like Rust supports a gtest-like `SetUp()` and `TearDown()`, so that means (I think) that this would all go into a single test. Does that make sense to do?

I think it would also be fine to have a freestanding setup function that you call in each test. Something like `run_with_test_nodes<F>(test_fn: F)` perhaps, which spins up the test nodes, runs the provided function, and then cleans up. I'd try that approach, and if spinning up the test nodes is too expensive, we can still group them into fewer tests.

christopherjreid:
Hrm, looks like adding the `test_msgs` dependency also pulls lots of other depedencies in... `nav_msgs, sensor_msgs, diagnostic_msgs, ` and more

```
error: failed to load source for dependency `diagnostic_msgs`

Caused by:
  Unable to update /workspace/install/diagnostic_msgs/share/diagnostic_msgs/rust

Caused by:
  failed to read `/workspace/install/diagnostic_msgs/share/diagnostic_msgs/rust/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
  ```
  
  Not sure what's going on with that....

File: rclrs_tests/package.xml [hunk 1]
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+<?xml-model
+   href="http://download.ros.org/schema/package_format3.xsd"
+   schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>rclrs_tests</name>
+  <version>0.1.0</version>
```
esteve, line 7: Please change this to 0.2.0, when we eventually submit these packages to the ROS buildfarm, they all must have the same version. Thanks.

rclrs_tests/package.xml
```diff
@@ -0,0 +1,23 @@
...
+  <description>Package containing tests for rclrs that make use of msgs and other dependencies</description>
+  <maintainer email="christopherjreid@outlook.com">Chris Reid</maintainer>
+  <license>Apache License 2.0</license>
+
+  <build_depend>test_msgs</build_depend>
+  <build_depend>rclrs</build_depend>
+  <build_depend>rosidl_runtime_rs</build_depend>
+
+  <exec_depend>test_msgs</exec_depend>
```
esteve, line 16: These dependencies should be declared as `<build_depend>` as well since we need them at build time. You can simplify this by just using `<depend>` which combines both build and exec dependencies.

esteve (changes requested):

christopherjreid:
After rebasing and adding all the msg dependencies, it's building and running now. I'm getting a spurious failure with `count_publishers` and `count_subscribers` - after constructing 2 publishers/subscribers, they occasionally return `3` instead. This does not seem to happen when I run a single test individually (e.g. `cargo test test_publishers`), so seems to be concurrency related. I'd say it's quite likely they're conflicting with `test_topic_names_and_types`, which creates a single publisher and subscriber. I might just distribute that functionality into `test_publishers` and `test_subscribers`

nnmm:
> After rebasing and adding all the msg dependencies, it's building and running now. I'm getting a spurious failure with `count_publishers` and `count_subscribers` - after constructing 2 publishers/subscribers, they occasionally return `3` instead. This does not seem to happen when I run a single test individually (e.g. `cargo test test_publishers`), so seems to be concurrency related. I'd say it's quite likely they're conflicting with `test_topic_names_and_types`, which creates a single publisher and subscriber. I might just distribute that functionality into `test_publishers` and `test_subscribers`

Yes, that's annoying. I think it's also fine to run `cargo test` with `--jobs 1` to avoid crosstalk.

christopherjreid:
Updated to combine everything into a single test - constructing the graph, then calling extracted functions for checking publishers, subscriptions, etc. I know that's a bit of an anti-pattern, but I felt like it was the best way to handle the concurrency issues.

Looks like it still has a compilation failure on GH though, so gonna troubleshoot that later

christopherjreid:
I can't reproduce that error locally - looks like a problem with the export dependencies of `test_msgs`? Not sure how to proceed...

nnmm:
@christopherjreid You can try adding a line to the [workflow](https://github.com/ros2-rust/ros2_rust/blob/main/.github/workflows/rust.yml) that sources the install directory. Probably

```
. ${{ steps.build.outputs.ros-workspace-directory-name }}/install/setup.bash
```

but if that's not the right path, you might need to dig around a bit by adding `ls` commands etc.

christopherjreid:
Thanks for the tips! That, and some test changes (apparently Galactic/cyclonedds do not allow publishing to the same topics with different types) got all the tests to pass!

So this is ready for official review now

rclrs_tests/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+[lib]
+path = "src/lib.rs"
+
+[dependencies]
+anyhow = {version = "1", features = ["backtrace"]}
+test_msgs = {version = "*"}
+diagnostic_msgs = {version = "*"}
+nav_msgs = {version = "*"}
+sensor_msgs = {version = "*"}
+stereo_msgs = {version = "*"}
```
nnmm, line 16: These packages seem to be unused (except test_msgs), could you remove them?

rclrs_tests/src/graph_tests.rs
```diff
@@ -0,0 +1,144 @@
+use rclrs::{Context, Node, RclrsError, TopicNamesAndTypes, QOS_PROFILE_SYSTEM_DEFAULT};
+
+use test_msgs::{msg, srv};
+
+#[test]
+fn foo() -> Result<(), RclrsError> {
```
nnmm, line 6: `foo` was probably a placeholder, right?

nnmm:
Looks good in general, I'm just wondering if it's really necessary to group everything in one test because of the interference problem. Would using separate namespaces for each test solve the problem too?

rclrs_tests/Cargo.toml
christopherjreid: `cargo build` won't run without this. Some sort of transitive dependency error?

christopherjreid:
I just added a new commit that creates a new test case for publishers, and one for subscribers. I feel like this is a lot noisier than the previous method, though there is definitely value in having "smaller" (or at least more targeted) test cases like this. What do you think, should I continue on this path for the remaining 3 test cases? (topics, services, and clients)?

nnmm:
> I just added a new commit that creates a new test case for publishers, and one for subscribers. I feel like this is a lot noisier than the previous method, though there is definitely value in having "smaller" (or at least more targeted) test cases like this. What do you think, should I continue on this path for the remaining 3 test cases? (topics, services, and clients)?

Hmm, I'd tend towards writing the tests in the most idiomatic way for now and just solve the interference problem by running them sequentially for now (i.e. adding `--jobs 1` to CI for this package). Maybe we'll find a better approach for test isolation in the future, involving `ROS_DOMAIN_ID` or so.

Is that fine with you too @esteve?

rclrs_tests/Cargo.toml
nnmm: I can't reproduce this locally, for me everything builds fine (with `colcon` and `cargo`). Can you show me the exact steps and error?

rclrs_tests/package.xml
nnmm: This one is not yet done.

christopherjreid: Ahh, I changed the xml version instead, sorry!

nnmm: Thank you! Could you also change the version number in `Cargo.toml` to 0.2?

christopherjreid:
Haven't looked at this in a couple days, but noticed CI had a test failure. Running this locally, I notice there's about a 1/15 chance or so that it randomly fails one of the tests, because the publisher/subscription/service/client that -should- exist does not.

Nothing within the test file itself should have concurrency issues anymore - is there some under-the-hood race condition that I'm hitting? Should I sleep for a bit after constructing the graph?

nnmm:
> Haven't looked at this in a couple days, but noticed CI had a test failure. Running this locally, I notice there's about a 1/15 chance or so that it randomly fails one of the tests, because the publisher/subscription/service/client that -should- exist does not.
> 
> Nothing within the test file itself should have concurrency issues anymore - is there some under-the-hood race condition that I'm hitting? Should I sleep for a bit after constructing the graph?

It seems like nodes won't get immediately notified of other entities (topics, publishers etc.), but only asynchronously.

As far as I can see in https://github.com/ros2/rclcpp/blob/rolling/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpphttps://github.com/ros2/rclcpp/blob/rolling/rclcpp/test/rclcpp/node_interfaces/test_node_graph.cpp, they use a `wait_for_graph_change()` function with a timeout. Maybe you can look into porting that function?

nnmm:
Hi @christopherjreid, I've been quiet since I've been on vacation, but please let me know if there's anything I can help you with here :) If you prefer, I could also take a stab at porting `wait_for_graph_change()`.

christopherjreid:
I would like to do it, just have been very busy (renovating our house). Hopefully I'll have some time this week to take a stab at it!

nnmm:
Okay, great – there's no rush at all.

christopherjreid:
Not sure why CI is failing. I'm rebased on main...

nnmm:
Not your fault. I just reran CI on another PR, and it failed too. We should report this error in https://github.com/ros-tooling/setup-ros or so, or check if someone already did:

```
  The following packages have unmet dependencies:
   libignition-math6-dev : Depends: libignition-cmake2-dev (>= 2.13.0) but 2.12.1-2~jammy is to be installed
  W: Target Packages (main/binary-amd64/Packages) is configured multiple times in /etc/apt/sources.list.d/ros-latest.list:1 and /etc/apt/sources.list.d/ros2-latest.list:1
  W: Target Packages (main/binary-all/Packages) is configured multiple times in /etc/apt/sources.list.d/ros-latest.list:1 and /etc/apt/sources.list.d/ros2-latest.list:1
  E: Unable to correct problems, you have held broken packages.
```

nnmm:
The CI error should now be resolved – but you also need to rebase on main to get clippy fixes that were made necessary by upgrading the Rust compiler.

christopherjreid:
Ready for merging now, I think!

nnmm (commented):
LGTM with some comments. I have some ideas for future improvement, but they don't need to happen in this PR, e.g. checking not only that the graph contains X, but also that the graph contains _only_ X.

@esteve Could you also take a look, since you requested changes here a while back?

rclrs_tests/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+authors = ["Chris Reid <christopherjreid@outlook.com>"]
+edition = "2021"
+
+[lib]
+path = "src/lib.rs"
+
+[dependencies]
+anyhow = {version = "1", features = ["backtrace"]}
+test_msgs = {version = "*"}
+#diagnostic_msgs = {version = "*"}
```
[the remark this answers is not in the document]
christopherjreid: Removed!

.github/workflows/rust.yml
```diff
@@ -95,11 +95,16 @@ jobs:
...
         cd ${{ steps.build.outputs.ros-workspace-directory-name }}
-        . /opt/ros/${{ matrix.ros_distribution }}/setup.sh
+        . install/setup.sh
         for path in $(colcon list | awk '$3 == "(ament_cargo)" && $1 != "examples_rclrs_minimal_pub_sub" && $1 != "examples_rclrs_minimal_client_service" { print $2 }'); do
         cd $path
         echo "Running cargo test in $path"
-        cargo test
+        if [ $path == "rclrs_tests" ]
+        then
+          cargo test --jobs 1
```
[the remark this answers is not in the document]
christopherjreid: Its no longer needed after adding the namespaces as you suggested!

nnmm (approved):
LGTM. Unfortunately I can't merge until @esteve removes his "changes requested" review.

esteve (approved):
@christopherjreid thanks for iterating with us!

PR #240, opened 2022-07-30
Small bugfix for sequences of WStrings

nnmm:
Message packages containing unbounded sequences of WStrings, like test_msgs, would not compile because of this.

There is a special path for sequences of Strings/WStrings in the rosidl_generator_rs template in the `into_rmw_message()` function, but it wasn't taken for `WStrings` because of a copy-paste error.


nnmm:
fyi @christopherjreid – after this you should be able to use test_msgs in #239.

esteve (approved):
@nnmm thanks for the detailed explanation for this PR, very helpful!

PR #247, opened 2022-08-14
Switch from actions-rs to dtolnay/rust-toolchain

nnmm:
The reason is that actions-rs became unmaintained (https://www.reddit.com/r/rust/comments/vyx4oj/actionsrs_organization_became_unmaintained/).
Also, dtolnay's rust-toolchain is simpler (at least in terms of code size).
Both are about equally fast (14s vs 12s in the runs I looked at, so 2s slower, but that doesn't matter compared to other steps).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -52,11 +52,8 @@ jobs:
         required-ros-distributions: ${{ matrix.ros_distribution }}
 
     - name: Setup Rust
-      uses: actions-rs/toolchain@v1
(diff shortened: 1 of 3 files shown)


nnmm:
Apparently a new version of Rust was released since the last commit which has some new clippy lints. I have added a commit to make it happy.

jhdcs (approved):
Looks good to me. Mr. Tolnay tends to make excellent packages, so I'm pretty sure this will work well for us.

PR #248, opened 2022-08-15
Flatten the module hierarchy

nnmm:
The modules for clients/services/publishers/subscriptions are moved up from the node/ directory
into the main directory.

This way, the node module only contains the implementation of the actual Node type.

Even though logically, a subscription for instance is a "part of" a node, the module layout should
not mirror this hierarchy, or we'd also move the node module below the context module and the
directories would become too nested.

This also mirrors the layout in rclcpp.


nnmm:
I haven't done the clippy fixes from https://github.com/ros2-rust/ros2_rust/pull/247 in here, since I'm thinking that one can be merged relatively quickly.

jhdcs (approved):
These seem to be fairly simple changes. I'm not seeing any immediately obvious issues.

esteve (approved):
@nnmm thanks!

PR #249, opened 2022-08-20
Adds Guard Conditions

christopherjreid:
- Adds the Guard Condition struct encapsulating rcl_guard_condition_t.
- Adds optional callbacks and a trigger method to approximate the rclcpp
  implementation.
- Adds an `add_guard_condition` method to WaitSet to add the
  GuardCondition to the WaitSet

### DIFF
--- a/rclrs/CHANGELOG.rst
+++ b/rclrs/CHANGELOG.rst
@@ -5,6 +5,8 @@ Changelog for package rclrs
 0.3
 ----------------
 * Loaned messages (zero-copy) (`#212 <https://github.com/ros2-rust/ros2_rust/pull/212>`_)
+* Graph queries (`#234 <https://github.com/ros2-rust/ros2_rust/pull/234>`_)
+* Guard conditions (`#249 <https://github.com/ros2-rust/ros2_rust/pull/249>`_)
 
 0.2 (2022-07-21)
 ----------------
--- a/rclrs/src/node.rs
+++ b/rclrs/src/node.rs
@@ -5,8 +5,9 @@ pub use self::graph::*;
 
 use crate::rcl_bindings::*;
 use crate::{
-    Client, ClientBase, Context, ParameterOverrideMap, Publisher, QoSProfile, RclrsError, Service,
-    ServiceBase, Subscription, SubscriptionBase, SubscriptionCallback, ToResult,
+    Client, ClientBase, Context, GuardCondition, ParameterOverrideMap, Publisher, QoSProfile,
+    RclrsError, Service, ServiceBase, Subscription, SubscriptionBase, SubscriptionCallback,
+    ToResult,
 };
 
 use std::cmp::PartialEq;
@@ -68,6 +69,7 @@ pub struct Node {
     pub(crate) rcl_node_mtx: Arc<Mutex<rcl_node_t>>,
     pub(crate) rcl_context_mtx: Arc<Mutex<rcl_context_t>>,
     pub(crate) clients: Vec<Weak<dyn ClientBase>>,
+    pub(crate) guard_conditions: Vec<Weak<GuardCondition>>,
     pub(crate) services: Vec<Weak<dyn ServiceBase>>,
     pub(crate) subscriptions: Vec<Weak<dyn SubscriptionBase>>,
     _parameter_map: ParameterOverrideMap,
@@ -189,6 +191,26 @@ impl Node {
         Ok(client)
     }
 
+    /// Creates a [`GuardCondition`][1].
+    ///
+    /// A weak pointer to the `GuardCondition` is stored within this node.
+    /// When this node is added to a wait set (e.g. when calling `spin_once`[2]
+    /// with this node as an argument), the guard condition can be used to
+    /// interrupt the wait.
+    ///
+    /// [1]: crate::GuardCondition
+    /// [2]: crate::spin_once
+    pub fn create_guard_condition(
+        &mut self,
+        context: &Context,
+        callback: Option<Box<dyn Fn() + Send + Sync>>,
+    ) -> Arc<GuardCondition> {
+        let guard_condition = Arc::new(GuardCondition::new(context, callback));
+        self.guard_conditions
+            .push(Arc::downgrade(&guard_condition) as Weak<GuardCondition>);
+        guard_condition
+    }
+
     /// Creates a [`Publisher`][1].
     ///
     /// [1]: crate::Publisher
@@ -252,6 +274,13 @@ impl Node {
 
     pub(crate) fn live_clients(&self) -> Vec<Arc<dyn ClientBase>> {
         self.clients.iter().filter_map(Weak::upgrade).collect()
+    }
+
+    pub(crate) fn live_guard_conditions(&self) -> Vec<Arc<GuardCondition>> {
+        self.guard_conditions
+            .iter()
+            .filter_map(Weak::upgrade)
+            .collect()
     }
 
     pub(crate) fn live_services(&self) -> Vec<Arc<dyn ServiceBase>> {
--- a/rclrs/src/node.rs
+++ b/rclrs/src/node.rs
@@ -191,7 +191,7 @@ impl Node {
         Ok(client)
     }
 
-    /// Creates a [`GuardCondition`][1].
+    /// Creates a [`GuardCondition`][1] with no callback.
     ///
     /// A weak pointer to the `GuardCondition` is stored within this node.
     /// When this node is added to a wait set (e.g. when calling `spin_once`[2]
@@ -200,12 +200,33 @@ impl Node {
     ///
     /// [1]: crate::GuardCondition
     /// [2]: crate::spin_once
-    pub fn create_guard_condition(
-        &mut self,
-        context: &Context,
-        callback: Option<Box<dyn Fn() + Send + Sync>>,
-    ) -> Arc<GuardCondition> {
-        let guard_condition = Arc::new(GuardCondition::new(context, callback));
+    pub fn create_guard_condition(&mut self) -> Arc<GuardCondition> {
+        let guard_condition = Arc::new(GuardCondition::new_with_rcl_context(
+            &mut self.rcl_context_mtx.lock().unwrap(),
+            None::<fn()>,
+        ));
+        self.guard_conditions
+            .push(Arc::downgrade(&guard_condition) as Weak<GuardCondition>);
+        guard_condition
+    }
+
+    /// Creates a [`GuardCondition`][1] with a callback.
+    ///
+    /// A weak pointer to the `GuardCondition` is stored within this node.
+    /// When this node is added to a wait set (e.g. when calling `spin_once`[2]
+    /// with this node as an argument), the guard condition can be used to
+    /// interrupt the wait.
+    ///
+    /// [1]: crate::GuardCondition
+    /// [2]: crate::spin_once
+    pub fn create_guard_condition_with_callback<F>(&mut self, callback: F) -> Arc<GuardCondition>
+    where
+        F: Fn() + Send + Sync + 'static,
+    {
+        let guard_condition = Arc::new(GuardCondition::new_with_rcl_context(
+            &mut self.rcl_context_mtx.lock().unwrap(),
+            Some(callback),
+        ));
         self.guard_conditions
             .push(Arc::downgrade(&guard_condition) as Weak<GuardCondition>);
         guard_condition
--- a/rclrs/src/node.rs
+++ b/rclrs/src/node.rs
@@ -203,7 +203,7 @@ impl Node {
     pub fn create_guard_condition(&mut self) -> Arc<GuardCondition> {
         let guard_condition = Arc::new(GuardCondition::new_with_rcl_context(
             &mut self.rcl_context_mtx.lock().unwrap(),
-            None::<fn()>,
+            None,
         ));
         self.guard_conditions
             .push(Arc::downgrade(&guard_condition) as Weak<GuardCondition>);
@@ -225,7 +225,7 @@ impl Node {
     {
         let guard_condition = Arc::new(GuardCondition::new_with_rcl_context(
             &mut self.rcl_context_mtx.lock().unwrap(),
-            Some(callback),
+            Some(Box::new(callback) as Box<dyn Fn() + Send + Sync>),
         ));
         self.guard_conditions
             .push(Arc::downgrade(&guard_condition) as Weak<GuardCondition>);
--- a/rclrs/src/wait.rs
+++ b/rclrs/src/wait.rs
@@ -369,7 +369,7 @@ mod tests {
     fn guard_condition_in_wait_set_readies() -> Result<(), RclrsError> {
         let context = Context::new([])?;
 
-        let guard_condition = Arc::new(GuardCondition::new(&context, None));
+        let guard_condition = Arc::new(GuardCondition::new(&context));
 
         let mut wait_set = WaitSet::new(0, 1, 0, 0, 0, 0, &context)?;
         wait_set.add_guard_condition(Arc::clone(&guard_condition))?;
--- a/rclrs/src/wait.rs
+++ b/rclrs/src/wait.rs
@@ -24,7 +24,9 @@ use std::time::Duration;
 use std::vec::Vec;
 
 mod exclusivity_guard;
+mod guard_condition;
 use exclusivity_guard::*;
+pub use guard_condition::*;
 
 /// A struct for waiting on subscriptions and other waitable entities to become ready.
 pub struct WaitSet {
@@ -36,6 +38,8 @@ pub struct WaitSet {
     // even in the error case.
     subscriptions: Vec<ExclusivityGuard<Arc<dyn SubscriptionBase>>>,
     clients: Vec<ExclusivityGuard<Arc<dyn ClientBase>>>,
+    // The guard conditions that are currently registered in the wait set.
+    guard_conditions: Vec<ExclusivityGuard<Arc<GuardCondition>>>,
     services: Vec<ExclusivityGuard<Arc<dyn ServiceBase>>>,
 }
 
@@ -45,6 +49,8 @@ pub struct ReadyEntities {
     pub subscriptions: Vec<Arc<dyn SubscriptionBase>>,
     /// A list of clients that have potentially received responses.
     pub clients: Vec<Arc<dyn ClientBase>>,
+    /// A list of guard conditions that have been triggered.
+    pub guard_conditions: Vec<Arc<GuardCondition>>,
     /// A list of services that have potentially received requests.
     pub services: Vec<Arc<dyn ServiceBase>>,
 }
@@ -105,6 +111,7 @@ impl WaitSet {
             rcl_wait_set,
             _rcl_context_mtx: context.rcl_context_mtx.clone(),
             subscriptions: Vec::new(),
+            guard_conditions: Vec::new(),
             clients: Vec::new(),
             services: Vec::new(),
         })
@@ -116,6 +123,7 @@ impl WaitSet {
     /// [`WaitSet::new`].
     pub fn clear(&mut self) {
         self.subscriptions.clear();
+        self.guard_conditions.clear();
         self.clients.clear();
         self.services.clear();
         // This cannot fail – the rcl_wait_set_clear function only checks that the input handle is
@@ -156,6 +164,38 @@ impl WaitSet {
         }
         .ok()?;
         self.subscriptions.push(exclusive_subscription);
+        Ok(())
+    }
+
+    /// Adds a guard condition to the wait set.
+    ///
+    /// # Errors
+    /// - If the guard condition was already added to this wait set or another one,
+    ///   [`AlreadyAddedToWaitSet`][1] will be returned
+    /// - If the number of guard conditions in the wait set is larger than the
+    ///   capacity set in [`WaitSet::new`], [`WaitSetFull`][2] will be returned
+    ///
+    /// [1]: crate::RclrsError
+    /// [2]: crate::RclReturnCode
+    pub fn add_guard_condition(
+        &mut self,
+        guard_condition: Arc<GuardCondition>,
+    ) -> Result<(), RclrsError> {
+        let exclusive_guard_condition = ExclusivityGuard::new(
+            Arc::clone(&guard_condition),
+            Arc::clone(&guard_condition.in_use_by_wait_set),
+        )?;
+
+        unsafe {
+            // SAFETY: Safe if the wait set and guard condition are initialized
+            rcl_wait_set_add_guard_condition(
+                &mut self.rcl_wait_set,
+                &*guard_condition.rcl_guard_condition.lock().unwrap(),
+                std::ptr::null_mut(),
+            )
+            .ok()?;
+        }
+        self.guard_conditions.push(exclusive_guard_condition);
         Ok(())
     }
 
@@ -262,6 +302,7 @@ impl WaitSet {
         let mut ready_entities = ReadyEntities {
             subscriptions: Vec::new(),
             clients: Vec::new(),
+            guard_conditions: Vec::new(),
             services: Vec::new(),
         };
         for (i, subscription) in self.subscriptions.iter().enumerate() {
@@ -275,6 +316,7 @@ impl WaitSet {
                     .push(Arc::clone(&subscription.waitable));
             }
         }
+
         for (i, client) in self.clients.iter().enumerate() {
             // SAFETY: The `clients` entry is an array of pointers, and this dereferencing is
             // equivalent to
@@ -284,6 +326,19 @@ impl WaitSet {
                 ready_entities.clients.push(Arc::clone(&client.waitable));
             }
         }
+
+        for (i, guard_condition) in self.guard_conditions.iter().enumerate() {
+            // SAFETY: The `clients` entry is an array of pointers, and this dereferencing is
+            // equivalent to
+            // https://github.com/ros2/rcl/blob/35a31b00a12f259d492bf53c0701003bd7f1745c/rcl/include/rcl/wait.h#L419
+            let wait_set_entry = unsafe { *self.rcl_wait_set.guard_conditions.add(i) };
+            if !wait_set_entry.is_null() {
+                ready_entities
+                    .guard_conditions
+                    .push(Arc::clone(&guard_condition.waitable));
+            }
+        }
+
         for (i, service) in self.services.iter().enumerate() {
             // SAFETY: The `services` entry is an array of pointers, and this dereferencing is
             // equivalent to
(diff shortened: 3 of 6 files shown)


rclrs/src/wait.rs
```diff
@@ -107,6 +115,7 @@ impl WaitSet {
     /// [`WaitSet::new`].
     pub fn clear(&mut self) {
         self.subscriptions.clear();
+        self.guard_conditions.clear();
```
christopherjreid, line 118: I think I should iterate over each guard condition here and set it's `in_use_by_wait_set` flag to false?

nnmm:
I'll check it out this evening!

rclrs/src/wait.rs
christopherjreid: Updated, but now I'm wondering if we need a `Drop` impl for `WaitSet` that also properly "releases" the guard conditions. @nnmm  thoughts?

rclrs/src/wait/guard_condition.rs
```diff
@@ -0,0 +1,133 @@
...
+    rcl_guard_condition: Arc<Mutex<rcl_guard_condition_t>>,
+    /// An optional callback to call when this guard condition is triggered.
+    callback: Option<Box<dyn Fn(usize)>>,
+    /// A flag to indicate if this guard condition has already been assigned to a wait set.
+    pub(crate) in_use_by_wait_set: Arc<AtomicBool>,
+    /// A count for the number of times this guard condition was triggered, but no callback was assigned.
+    unread_count: usize,
+}
+
+impl Drop for GuardCondition {
```
nnmm, line 18: Users will probably want to use the `GuardConditions` from a separate thread than the one the wait set is running in. So we should check if we can implement `Sync` and `Send`.

rclrs/src/wait/guard_condition.rs
```diff
@@ -0,0 +1,133 @@
...
+            // SAFETY: No precondition for this function (besides passing in a valid guard condition)
+            rcl_guard_condition_fini(&mut *self.rcl_guard_condition.lock().unwrap());
+        }
+    }
+}
+
+impl GuardCondition {
+    /// Creates a new guard condition
+    pub fn new(context: &Context) -> Self {
+        let mut guard_condition = unsafe { rcl_get_zero_initialized_guard_condition() };
```
nnmm, line 30: Obligatory "Getting a zero-initialized value is always safe." comment ;)

rclrs/src/wait/guard_condition.rs
```diff
@@ -0,0 +1,133 @@
...
+        unsafe {
+            // SAFETY: The context must be valid. No other preconditions for this function.
```
nnmm, line 32: Well, and the guard condition itself is zero-initialized.

nnmm (commented):
Great job as usual. I left a round of comments.

rclrs/src/wait.rs
```diff
@@ -150,6 +166,38 @@ impl WaitSet {
...
+    /// - If the guard condition was already added to this wait set or another one,
+    ///   [`AlreadyAddedToWaitSet`][1] will be returned
+    /// - If the number of guard condition in the wait set is larger than the
+    ///   capacity set in [`WaitSet::new`], [`WaitSetFull`][2] will be returned
+    ///
+    /// [1]: crate::RclrsError
+    /// [2]: crate::RclReturnCode
+    pub fn add_guard_condition(
+        &mut self,
+        guard_condition: Arc<Mutex<GuardCondition>>,
```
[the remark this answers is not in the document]
nnmm: Correction: I see now why you have wrapped it in a Mutex – because the `trigger()` method needs a mutable self.

The docs sound like the `rcl_trigger_guard_condition()` function can run concurrently with itself, and the "no" in thread-safe is only wrt the fini function: https://github.com/ros2/rcl/blob/4b125b1af0e2e2c8c7dd0c8e18b5a8d36709058c/rcl/include/rcl/guard_condition.h#L197

So, in theory, we could make `trigger()` take a shared reference and cast it to a mutable pointer (raw mutable pointers may alias). But it's a bit spicy.

If we don't do that, then I'd propose moving the `Mutex` inside the guard condition. The guard condition will always need to be accessed from different threads, so we can save the user some work there.

rclrs/src/wait/guard_condition.rs
nnmm: I found that in my demo, I also needed a `Send` bound on the callback.

christopherjreid:
Updated to address all comments!

rclrs/src/wait/guard_condition.rs
```diff
@@ -0,0 +1,149 @@
+use crate::rcl_bindings::*;
+use crate::{Context, RclrsError, ToResult};
+
+use std::sync::{atomic::AtomicBool, Arc, Mutex, MutexGuard};
+
+/// A struct for encapsulating a guard condition - a waitable trigger
```
nnmm, line 6: Personally, I was a bit confused when I first encountered guard conditions. A description like this may help others:

```suggestion
/// A waitable entity used for waking up a wait set manually.
///
/// If a wait set that is currently waiting on events should be interrupted from a separate thread, this can be done
/// by adding an `Arc<GuardCondition>` to the wait set, and calling `trigger()` on the same `GuardCondition` while
/// the wait set is waiting.
///
/// The guard condition may be reused multiple times, but like other waitable entities, can not be used in
/// multiple wait sets concurrently.
```

rclrs/src/wait/guard_condition.rs
```diff
@@ -0,0 +1,149 @@
...
+
+// SAFETY: rcl_guard_condition is the only member that doesn't implement Send, and it is designed to be accessed from other threads
+unsafe impl Send for GuardCondition {}
+
+// SAFETY: The mutexes and atomic members ensure synchronized access to members, and the callback is reentrant
+unsafe impl Sync for GuardCondition {}
+
+impl GuardCondition {
+    /// Creates a new guard condition
+    pub fn new(context: &Context, callback: Option<Box<dyn Fn() + Send>>) -> Self {
```
nnmm, line 62: Another usability idea: We can return `Arc<Self>` here. You'll never want to use one without `Arc`.

rclrs/src/wait/guard_condition.rs
```diff
@@ -0,0 +1,156 @@
...
+        ws.add_guard_condition(Arc::clone(&gc))?;
+        gc.trigger()?;
+
+        assert_eq!(atomic_bool.load(Ordering::Relaxed), true);
+        ws.wait(Some(std::time::Duration::from_millis(10)))?;
+
+        Ok(())
+    }
+
+    fn assert_send<T: Send>() {}
```
esteve, line 148: These assertions and tests are not necessary, I understand the rationale for them, but we don't test the existence of every trait we use.

rclrs/src/wait/guard_condition.rs
```diff
@@ -0,0 +1,156 @@
...
+    use std::sync::atomic::Ordering;
+
+    #[test]
+    fn test_guard_condition() -> Result<(), RclrsError> {
+        let context = Context::new([])?;
+
+        let atomic_bool = Arc::new(std::sync::atomic::AtomicBool::new(false));
+        let atomic_bool_for_closure = Arc::clone(&atomic_bool);
+
+        let gc = GuardCondition::new(
```
esteve, line 110: Minor nitpick, I prefer long names for variables (i.e. `guard_condition`), even if it's just tests.

rclrs/src/wait.rs
```diff
@@ -275,6 +313,7 @@ impl WaitSet {
                     .push(Arc::clone(&subscription.waitable));
             }
         }
+
```
esteve, line 316: There should be a block similar to subscriptions/clients/services for guard conditions as well.

esteve (changes requested):

rclrs/src/wait/guard_condition.rs
```diff
@@ -0,0 +1,156 @@
...
+
+// SAFETY: rcl_guard_condition is the only member that doesn't implement Send, and it is designed to be accessed from other threads
+unsafe impl Send for rcl_guard_condition_t {}
+
+// SAFETY: The mutexes and atomic members ensure synchronized access to members, and the callback is reentrant
+unsafe impl Sync for GuardCondition {}
+
+impl GuardCondition {
+    /// Creates a new guard condition.
+    pub fn new(context: &Context, callback: Option<Box<dyn Fn() + Send + Sync>>) -> Arc<Self> {
```
esteve, line 65: I disagree about returning an `Arc` favoring usability. For me, conventions and the principle of least surprise make APIs more usable, in this case, given that a `GuardCondition` behaves like other waitable entities, I'd return a `Self` here and add a `create_guard_condition` method to `Node`, which would return an `Arc`.

In this case `rclcpp` is being inconsistent, and `rclpy` fixed this inconsistency https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/node.py#L1696

esteve (changes requested):

rclrs/src/wait/guard_condition.rs
nnmm: I disagree. These are not comparable to other traits like `Clone`, `Message` etc. in that they are the only auto traits, i.e.  implemented _entirely implicitly_ (apart from `Unpin`, `UnwindSafe`, `RefUnwindSafe`, which we don't care about). I would never write an assertion that `Clone` or `Message` is implemented.

Have you seen https://github.com/ros2-rust/ros2_rust/issues/253? If we accidentally make a change that makes these types not `Sync` or `Send`, we will break code.

Also, in the future, as soon as we start caring about semantic versioning, tests like these provide protection against breaking backwards compatiblity by changing auto trait implementations.

rclrs/src/wait/guard_condition.rs
nnmm: All the other user-facing methods for creating waitable entities (`Node::create_subscription` etc.) already return `Arc`, so if you want `GuardCondition` to behave like other waitable entities, that's actually in favor of making this method return an `Arc` too.

Python is not really comparable, everything is an `Arc` (or at least `Rc`) there basically.

rclrs/src/wait/guard_condition.rs
esteve: Yes, I had seen #253, but that issue could have been detected with better tests that actually test the functionality that's provided. Testing for traits smells like bad code in my opinion, but if it's a common pattern in Rust that I'm not aware of, do you know of an example project or a documentation reference that recommends something like this?

rclrs/src/wait/guard_condition.rs
esteve: > All the other user-facing methods for creating waitable entities (`Node::create_subscription` etc.) already return `Arc`, so if you want `GuardCondition` to behave like other waitable entities, that's actually in favor of making this method return an `Arc` too.

No, because `Subscription::new` for example, returns a `Self`, not an `Arc<Self>`. `create_X` are not comparable with `::new` as they need to add the waitable to an internal list in the `Node`

> Python is not really comparable, everything is an `Arc` (or at least `Rc`) there basically.

I was comparing it at the API level, not at the functional level, I thought that was implied.

rclrs/src/wait/guard_condition.rs
nnmm: @esteve: https://rust-lang.github.io/api-guidelines/interoperability.html#types-are-send-and-sync-where-possible-c-send-sync

esteve: @nnmm thanks, that was really useful, glad to be proved wrong :slightly_smiling_face:

rclrs/src/wait/guard_condition.rs
nnmm: > > All the other user-facing methods for creating waitable entities (`Node::create_subscription` etc.) already return `Arc`, so if you want `GuardCondition` to behave like other waitable entities, that's actually in favor of making this method return an `Arc` too.
> 
> No, because `Subscription::new` for example, returns a `Self`, not an `Arc<Self>`. `create_X` are not comparable with `::new` as they need to add the waitable to an internal list in the `Node`

That's what I meant with "user-facing" – `Subscription::new` etc. are private.

But this touches on an interesting point: I would say `GuardCondition::new` can be public, since a guard condition doesn't always need to be attached to a node, does it? There _might_ be a use case of attaching it to a node, if you don't `spin()` the node but add it to a wait set, but there is also a use for using a standalone guard condition with a wait set, i.e. where the guard condition is external to the node in the wait set.

> 
> > Python is not really comparable, everything is an `Arc` (or at least `Rc`) there basically.
> 
> I was comparing it at the API level, not at the functional level, I thought that was implied.

Hmm, I'm not sure what you're getting at with the reference to `rclpy`, sorry ... might need to read up on guard conditions in `rclcpp` and `rclpy`.

rclrs/src/wait/guard_condition.rs
nnmm: I think there might have been a git error – it's not in the latest revision yet.

rclrs/src/wait.rs
nnmm: I thought about it some more, and I agree we can add it back if we can make guard conditions implement `Eq`, so that you can find out _which_ guard condition triggered. That might make it useful.

rclrs/src/wait/guard_condition.rs
nnmm: I wouldn't mind your suggestion of a free function, but it would look a little strange. I think @esteve should let us know if he still thinks the current state of things inconsistent after considering the points on private vs public, and the use case for guard conditions that are _not_ part of the node.

Otherwise, the next best thing imo would be to add both the `GuardCondition::new() -> Self` version and a `GuardCondition::new_arc() -> Arc<Self>` version.

Then there is the other, unrelated point of adding `Node::create_guard_condition()`. I agree with Esteve here that we probably want it – sure, guard conditions are not logically part of a node, but it could be potentially useful (and `rclpy` has such a method). E.g. for the dynamic message GUI I had only one node, and I added that node and the guard condition to the wait set separately. With this added method (which would store a `Weak<GuardCondition>` inside the node), the guard condition could be added to the wait set as part of adding the node to the wait set, making the code shorter. We need to clearly document the difference between `Node::create_guard_condition()` and `GuardCondition::new(_arc)()` though.

Of course `Node::create_guard_condition()` should return an `Arc`, I think we all agree there.

nnmm:
BTW, you could add guard conditions (and graph support, which I forgot) to the new `rclrs/CHANGELOG.rst` file in here too.

christopherjreid:
All comments addressed, except for the open discussion on the return type of `GuardCondition::new`.
We may need a new discussion on the impl of `PartialEq for GuardCondition`. I believe that -only- checking eq on the underlying `rcl_guard_condition` is sufficient, but I'm open to thoughts on that. @nnmm suggested a static counter that would serve as an ID for each `GuardCondition`

rclrs/src/wait/guard_condition.rs
```diff
@@ -0,0 +1,173 @@
...
+            rcl_guard_condition_fini(&mut *self.rcl_guard_condition.lock().unwrap());
+        }
+    }
+}
+
+impl PartialEq for GuardCondition {
+    fn eq(&self, other: &Self) -> bool {
+        // Because GuardCondition controls the creation of the rcl_guard_condition, each unique GuardCondition should have a unique
+        // rcl_guard_condition. Thus comparing equality of this member should be enough.
+        Arc::ptr_eq(&self.rcl_guard_condition, &other.rcl_guard_condition)
```
nnmm, line 68: Did you add the inner `Arc` for this purpose, or is it required by something else too?

rclrs/src/wait/guard_condition.rs
esteve: It's strange to me that there's no `Node::create_guard_condition()` in `rclcpp`, but there is in `rclpy`. In this case I'd favor having that method to make the API more consistent with the rest of waitable entities.

`rclpy` even keeps track of the guard conditions, which are triggered on every `spin`, so this inconsistency between `rclpy` and `rclcpp` is rather intriguing.

nnmm: @christopherjreid, I agree with Esteve, please add `Node::create_guard_condition()`.

rclrs/src/wait/guard_condition.rs
christopherjreid: It's needed so that we can lock the inner rcl_guard_condition in the `trigger` function

nnmm: I tried it, the `trigger` function works fine without the `Arc`. So I'd say we remove it and use some other indicator of equality, like equality of the `impl` pointers: https://github.com/ros2/rcl/blob/3fa178ad6c60bf16447654a01a4f68e8d400b517/rcl/include/rcl/guard_condition.h#L41

christopherjreid: Pushed!

rclrs/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rclrs"
-version = "0.2.0"
```
nnmm, line 3: Let's maybe do the version update in bulk for all packages.

rclrs/src/node.rs
```diff
@@ -189,6 +190,17 @@ impl Node {
         Ok(client)
     }
 
+    /// Creates a [`GuardCondition`][1].
+    ///
+    /// [1]: create::GuardCondition
```
nnmm, line 195:
```suggestion
    /// [1]: crate::GuardCondition
```

rclrs/src/wait/guard_condition.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+}
+
+impl Eq for GuardCondition {}
+
+// SAFETY: rcl_guard_condition is the only member that doesn't implement Send, and it is designed to be accessed from other threads
+unsafe impl Send for rcl_guard_condition_t {}
+
+// SAFETY: The mutexes and atomic members ensure synchronized access to members, and the callback is reentrant
+unsafe impl Sync for GuardCondition {}
```
nnmm, line 81: Please remove this, it shouldn't be necessary. All members implement `Sync`.

rclrs/src/node.rs
```diff
@@ -189,6 +190,17 @@ impl Node {
...
+    pub fn create_guard_condition(
+        &self,
+        context: &Context,
+        callback: Option<Box<dyn Fn() + Send + Sync>>,
+    ) -> Result<Arc<GuardCondition>, RclrsError> {
+        Ok(Arc::new(GuardCondition::new(context, callback)))
```
nnmm, line 201: This function should also add the guard condition to a node-internal list of guard conditions, which are then added to the wait set in `spin_once()`, like subscriptions etc.

rclrs/src/wait/guard_condition.rs
nnmm: Thanks!

rclrs/Cargo.toml
christopherjreid: Pushed!

rclrs/src/wait/guard_condition.rs
christopherjreid: And the test confirms it!

rclrs/CHANGELOG.rst
```diff
@@ -5,11 +5,13 @@ Changelog for package rclrs
 0.3
 ----------------
 * Loaned messages (zero-copy) (`#212 <https://github.com/ros2-rust/ros2_rust/pull/212>`_)
+* Graph support (`#234 <https://github.com/ros2-rust/ros2_rust/pull/234>`_)
```
nnmm, line 8: Maybe this could be `Graph queries` instead of `Graph support`, since there is also the graph listener in rclcpp which we don't have yet.

rclrs/src/wait/guard_condition.rs
```diff
@@ -0,0 +1,173 @@
...
+}
+
+impl Eq for GuardCondition {}
+
+// SAFETY: rcl_guard_condition is the only member that doesn't implement Send, and it is designed to be accessed from other threads
+unsafe impl Send for rcl_guard_condition_t {}
+
+impl GuardCondition {
+    /// Creates a new guard condition.
+    pub fn new(context: &Context, callback: Option<Box<dyn Fn() + Send + Sync>>) -> Self {
```
nnmm, line 82: For slightly better usability, we could make this take an un-boxed callback and box it ourselves in the function body.

nnmm (commented):
(Probably) final comments from my side

rclrs/src/wait/guard_condition.rs
nnmm: @esteve you haven't commented on this yet: 

> I think @esteve should let us know if he still thinks the current state of things (i.e. returning an `Arc` from `GuardCondition::new()`) inconsistent after considering the points on private vs public, and the use case for guard conditions that are _not_ part of the node.

rclrs/src/wait/guard_condition.rs
christopherjreid: Hrm. I played around with this a bit, and couldn't figure out how to get around the `Sized` requirement without the `Box`. Any tips?

nnmm: You can write
```
    pub fn new<F>(context: &Context, callback: Option<F>) -> Self
    where
        F: Fn() + Send + Sync + 'static,
```
and below
```
        Self {
            rcl_guard_condition: Mutex::new(guard_condition),
            callback: callback.map(|cb| Box::new(cb) as Box<dyn Fn() + Send + Sync>),
            in_use_by_wait_set: Arc::new(AtomicBool::new(false)),
        }
```

rclrs/src/wait/guard_condition.rs
esteve: I'd still prefer `GuardCondition::new` to return a `GuardCondition` to keep it consistent with the other waitables, and then give users `Node::create_guard_condition` (which returns an `Arc`) as the entry point for the API.

rclrs/src/wait/guard_condition.rs
christopherjreid: Pushed! I thought I tried this but I think I tried to cast inside the box instead of outside, which obviously wouldn't work.

rclrs/CHANGELOG.rst
esteve: I agree about graph queries, graph support basically includes that and things like `wait_for_service`, which is not implemented yet.

examples/message_demo/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "examples_rclrs_message_demo"
-version = "0.2.0"
+version = "0.3.0"
```
esteve, line 3: Please revert the version changes, we'll make these changes when we prepare for the 0.3.0 release. Thanks.

rclrs/Cargo.toml
esteve: @christopherjreid @nnmm This should be part of the release process in a separate PR, not part of an existing PR.

examples/message_demo/Cargo.toml
christopherjreid: Reverted!

esteve:
@christopherjreid in the future, could you push commit by commit, instead of force-push? It's easier to see the changes that happen while reviewing a PR. We'll squash all the commits in the end when we merge the PR, so there's no need to squash them while the PR is being worked on. Thanks.

rclrs/src/wait/guard_condition.rs
nnmm: Ok. thanks, I'll resolve this then. Note though that `Node::create_guard_condition` has different behavior than `Arc::new(GuardCondition::new())` – with the latter, the guard condition is not attached to the node.

rclrs/src/node.rs
```diff
@@ -189,6 +191,26 @@ impl Node {
...
+    /// A weak pointer to the `GuardCondition` is stored within this node.
+    /// When this node is added to a wait set (e.g. when calling `spin_once`[2]
+    /// with this node as an argument), the guard condition can be used to
+    /// interrupt the wait.
+    ///
+    /// [1]: crate::GuardCondition
+    /// [2]: crate::spin_once
+    pub fn create_guard_condition(
+        &mut self,
+        context: &Context,
```
nnmm, line 205: Sorry, I have one more comment.

It would be nice if this method
- took an un-boxed callback like `GuardCondition::new()`
- did not require passing in the context. The node already has the context, but only the `rcl_context_t`, not the `Context`. To make this work, the `GuardCondition` could have a non-public method `new_with_rcl_context()` that takes an `rcl_context_t` (or a `Mutex<rcl_context_t>`, you get the idea) . Then `GuardCondition::new()` and `Node::create_guard_condition()` could call `GuardCondition::new_with_rcl_context()`.

nnmm:
I also took a look at the clippy error – I hadn't thought about that. Maybe it would be nicest to have two methods:
- `GuardCondition::new()` (does not take a callback)
- `GuardCondition::new_with_callback()` (takes a callback, i.e. not enclosed in an `Option`)

That way the callback type could always be deduced. I think it also makes the call sites prettier.

Alternatively, we could just revert to taking boxed callbacks.

esteve:
@christopherjreid I agree with @nnmm about adding an explicit `GuardCondition::new_with_callback()`, instead of an `Option`

christopherjreid:
Hrm, so we're looking at:
`
pub fn new(...);
pub fn new_with_callback(...);
pub(crate) fn new_with_rcl_context(...);
pub(crate) fn new_with_rcl_context_with_callback(...);
` 
?

nnmm:
> Hrm, so we're looking at:
> 
> ```
> pub fn new(...);
> 
> pub fn new_with_callback(...);
> 
> pub(crate) fn new_with_rcl_context(...);
> 
> pub(crate) fn new_with_rcl_context_with_callback(...);
> ```
> 
> ?

No, the rcl_context version can receive a boxed Option callback

rclrs/src/wait/guard_condition.rs
```diff
@@ -0,0 +1,186 @@
...
+        F: Fn() + Send + Sync + 'static,
+    {
+        Self::new_with_rcl_context(&mut context.rcl_context_mtx.lock().unwrap(), Some(callback))
+    }
+
+    /// Creates a new guard condition by providing the rcl_context_t and an optional callback.
+    /// Note this function enables calling `Node::create_guard_condition`[1] without providing the Context separately
+    ///
+    /// [1]: Node::create_guard_condition
+    pub(crate) fn new_with_rcl_context<F>(context: &mut rcl_context_t, callback: Option<F>) -> Self
```
nnmm, line 98: Nitpick: If this took a `Box<dyn Fn>`, we could make this non-generic, which is a little better for compile times and you wouldn't need the `::<fn()>` (though it's a nice trick!). But it's not really important, take it or leave it.

nnmm (approved):
LGTM!

rclrs/src/wait/guard_condition.rs
christopherjreid: I was thinking the same thing actually. I'll make this change real quick!

nnmm (approved):

esteve (approved):
@christopherjreid LGTM! Thank you so much for iterating with us and for your patience :slightly_smiling_face:

PR #250, opened 2022-08-21
Fix flaky unit test due to parallel test execution

nnmm:
In addition, this PR makes the output of `get_node_names_with_enclaves` more understandable by using the `NodeNameInfo` type.

### DIFF
--- a/rclrs/src/node/graph.rs
+++ b/rclrs/src/node/graph.rs
@@ -200,9 +200,7 @@ impl Node {
     }
 
     /// Returns a list of all node names with enclaves.
-    pub fn get_node_names_with_enclaves(
-        &self,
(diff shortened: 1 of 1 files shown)


nnmm:
@jhdcs @esteve Please review.

jhdcs (approved):
This seems straightforward, though I'm surprised that nodes from other tests would show up in each test.

nnmm:
Yes, because of parallel test execution. There aren't that many possible `ROS_DOMAIN_ID`s, or you could just isolate each test by setting that variable to a unique number at the beginning of each test. But test isolation is something we should do in the future.

nnmm:
Thanks for the review!

PR #252, opened 2022-09-10
Add notes on unsafe comments, client library consistency, and reviewing to contributing.md

nnmm:
Motivation: This documents several guidelines that have so far been implicit. I hope they represent a consensus opinion :)

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -61,6 +61,25 @@ Do not cast references to raw pointers where it's not needed. Since references c
 
 Generally, a `rcl` type should have a `Drop` impl that calls its `fini()` function. However, there are some cases where this is not possible, e.g. when the type is part of another `rcl` type that owns and finalizes it, or when the `fini()` function takes extra arguments. In these cases, it needs to be decided individually if and where to call `fini()`, and care must be taken to not return early from a function with `?` before the relevant `fini()` functions have been called.
 
+
+## Consistency with major ROS 2 client libraries
+Consistency with the two major ROS 2 client libraries, `rclcpp` and `rclpy`, is a goal of this project – and in particular `rclcpp`.
+
+That means that
+- Roughly the same concepts should exist in all client libraries, e.g. services, parameters, etc.
+- Types, functions, argument names etc. should have the same or similar names as their corresponding items in `rclcpp`/`rclpy` where possible
+- Code should be structured into similar packages by default, e.g. `rosidl_generator_<lang>`, `rcl<lang>` etc.
+- Features should be built on top of `rcl` by default, instead of re-implementing them from scratch
+
+This does _not_ mean that
+- No variability between languages is allowed. What is easy to do, and what is considered idiomatic, is quite different between languages, and thus ROS 2 allows for considerable variability between client languages.
+- Features should always be ported 1:1. Do not rely on the assumption that the best implementation for `rclrs` is whatever `rclcpp` does, but understand the motivation and status of the implementation in the other client libraries. For instance, sometimes a design is implemented incompletely by a client library, the feature is later found to have problems (e.g. the deadlocking of sync client calls in Python), or the client library is constrained by backwards compatibility.
+- Internal consistency does not matter. It does.
+- No deviations from the above guidelines are possible.
+
+In summary, understand the context of the features that you port, in order to avoid [_cargo-culting_](https://en.wikipedia.org/wiki/Cargo_cult_programming), and to be mindful of [Chesterton's fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence).
+
+
 ## License
 Any contribution that you make to this repository will
 be under the Apache 2 License, as dictated by that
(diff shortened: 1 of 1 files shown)


jhdcs (dismissed):
These appear to be good additions to me. I'll have to work on my reviewing style to comply with the new guidelines.

docs/CONTRIBUTING.md
```diff
@@ -61,6 +68,32 @@ Do not cast references to raw pointers where it's not needed. Since references c
...
+- Types, functions, argument names etc. should have the same or similar names as their corresponding items in `rclcpp`/`rclpy` where possible
+- Code should be structured into similar packages by default, e.g. `rosidl_generator_<lang>`, `rcl<lang>` etc.
+- Features should be built on top of `rcl` by default, instead of re-implementing them from scratch
+
+This does _not_ mean that
+- No variability between languages is allowed. What is easy to do, and what is considered idiomatic, is quite different between languages, and thus ROS 2 allows for considerable variability between client languages.
+- Features should always be ported 1:1. Do not rely on the assumption that the best implementation for `rclrs` is whatever `rclcpp` does, but understand the motivation and status of the implementation in the other client libraries. For instance, sometimes a design is implemented incompletely by a client library, the feature is later found to have problems (e.g. the deadlocking of sync client calls in Python), or the client library is constrained by backwards compatibility.
+- Internal consistency does not matter. It does.
```
esteve, line 84: I'd add external consistency as well, API should follow the principle of least surprise, where there are no quirks in how similar functions are called and the types they return.

docs/CONTRIBUTING.md
```diff
@@ -61,6 +68,32 @@ Do not cast references to raw pointers where it's not needed. Since references c
...
+- No deviations from the above guidelines are possible.
+
```
esteve, line 86: Here I'd add excessively complex designs in C++ or Python may be caused by limitations of such languages and it's best to strive for clean simple functionality in Rust, where zero-cost abstractions are preferred. I recall that https://github.com/ros2-rust/ros2_rust/pull/210 started to get a rather complex design when in the end it only needed to be a free function.

docs/CONTRIBUTING.md
```diff
@@ -61,6 +68,32 @@ Do not cast references to raw pointers where it's not needed. Since references c
...
+In summary, understand the context of the features that you port, in order to avoid [_cargo-culting_](https://en.wikipedia.org/wiki/Cargo_cult_programming), and to be mindful of [Chesterton's fence](https://en.wiktionary.org/wiki/Chesterton%27s_fence).
```
esteve, line 87: I think the addition of Chesterton's fence here might be harmful, given that in other parts of this document it's advised to understand the rationale for a given design in `rclcpp` in order to be able to not port features 1:1

esteve (changes requested):

docs/CONTRIBUTING.md
nnmm: Hmm, yes you could read it as "be wary of deviating from the design of other client libraries", but I meant it more like "if you find the design of rclcpp and rclpy illogical, that's probably a signal that you don't understand the problem they are solving there".

But I see that it might be misleading, so I'm removing it.

docs/CONTRIBUTING.md
nnmm: Ok, I added that to the closely related item above.

nnmm (commented):
Thanks for the comments!

docs/CONTRIBUTING.md
esteve: I meant more in the sense of consistence in the `rclrs` API itself, over the years `rclcpp` and `rclpy` have their own quirks and sometimes the API is not entirely coherent. For example, `rclcpp`'s [create_subscription](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/node.hpp#L221-L222) has the QoS parameter before the callback, where [create_service](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/node.hpp#L316-L317) has them swapped. It's a delicate line, because on the one hand we want to make it easy for people to port their applications and mimic the existing APIs, but if such APIs are not consistent, it may make writing new applications more confusing.

esteve:
Regarding contributions, something that personally makes reviews more difficult for me is force-pushes and squashing while the PR is being reviewed, because I can't point to a commit to check if a particular feedback has been addressed. Could you add a section about that? We squash commits when we merge a PR, so there's not much change in how merge notes look like.

docs/CONTRIBUTING.md
jhdcs: In my opinion, if we have to make a decision to mimic `rclcpp`'s API or `rclpy`'s API, we should probably go for `rclcpp` - Rust is a bit closer to C++ than to Python, so (theoretically) it's more likely that a user of `rclcpp` would be using `rclrs`. At least that's my reasoning.

But yeah, ideally the APIs would be consistent across the board.

nnmm:
> Regarding contributions, something that personally makes reviews more difficult for me is force-pushes and squashing while the PR is being reviewed, because I can't point to a commit to check if a particular feedback has been addressed. Could you add a section about that? We squash commits when we merge a PR, so there's not much change in how merge notes look like.

Added. I think all comments should be addressed now.

jhdcs (approved):

nnmm:
@esteve I assume all your comments have been addressed and this can be merged, yes?

esteve (approved):
@nnmm nice work!

PR #254, opened 2022-09-11
Test for Send and Sync implementations, and add missing implementations

nnmm:
The tutorial did not work anymore because `Publisher` stopped implementing `Send` after https://github.com/ros2-rust/ros2_rust/pull/212.

This makes sure, with tests, that all types that can implement `Send` and `Sync` do so. This will avoid similar breakages in the future.

Please double-check my reasoning in the safety justifications!

### DIFF
--- a/rclrs/src/subscription/readonly_loaned_message.rs
+++ b/rclrs/src/subscription/readonly_loaned_message.rs
@@ -45,3 +45,9 @@ where
         }
     }
 }
+
+// SAFETY: The functions accessing this type, including drop(), shouldn't care about the thread
+// they are running in. Therefore, this type can be safely sent to another thread.
+unsafe impl<'a, T> Send for ReadOnlyLoanedMessage<'a, T> where T: RmwMessage {}
+// SAFETY: This type has no interior mutability, in fact it has no mutability at all.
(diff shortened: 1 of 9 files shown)


rclrs/src/subscription/readonly_loaned_message.rs
```diff
@@ -45,3 +45,9 @@ where
         }
     }
 }
+
+// SAFETY: The functions accessing this type, including drop(), shouldn't care about the thread
+// they are running in. Therefore, this type can be safely sent to another thread.
+unsafe impl<'a, T> Send for ReadOnlyLoanedMessage<'a, T> where T: RmwMessage {}
+// SAFETY: This type has no interior mutability, in fact it has no mutability at all.
```
jhdcs, line 52:
```suggestion
// SAFETY: This type is not mutable.
```

jhdcs (approved):
Just a minor comment suggestion, but it's not a huge deal. This looks good to me.

rclrs/src/subscription/readonly_loaned_message.rs
nnmm: Thanks for your suggestion – I like the previous version a tiny bit better though since "interior mutability" directly references the criterion for a type to be `Sync`. Since you already approved, I assume leaving it like that is also fine with you :)

PR #262, opened 2022-09-18
Add dynamic message functionality

nnmm:
**This is an overview PR to see the changes at a high level.** I recommend you check out this branch and run `cargo doc` to see what the API looks like. For the review, it will be split into smaller parts.

# What are dynamic messages?
Dynamic messages are messages whose type is only known at runtime. You create dynamic messages by specifying the type as a string, e.g. "geometry_msgs/msg/Twist". The dynamic message module then looks for a shared _introspection type support_ library, loads it, and parses the structure of the C message struct from it. That structure contains information about what fields the message contains and what their types are, e.g. `field 'angular' of type 'Vector3' at offset 24`. Using the structure, the dynamic message module can then read out the value of each field, and even modify it. Basically, it allows accessing the fields of the message as a HashMap from string to a value like this:

```
enum Value<'msg> {
  Float(&'msg f32),
  Double(&'msg f64),
  ...
  FloatArray(&'msg [f32]),
  ...
  FloatSequence(&'msg Sequence<f32>),
  ...
  FloatBoundedSequence(DynamicBoundedSequence<'msg, f32>)
  ..
}
```

This is useful for writing generic tools such as introspection tools, bridges to other communication systems, or nodes that manipulate messages à la `topic_tools`.

In C++, a similar thing exists in https://github.com/osrf/dynamic_message_introspection and https://github.com/facontidavide/ros2_introspection.

# Features:
* Subscription and publisher for dynamic messages
* Conversion between statically typed and dynamically typed messages
* Safe access of fields – no `unsafe` required in user code

# Non-features:
* Dynamic services and clients
* Serde impls for the dynamic type (requires compiling the Rust message package into a dynamic library)

# Notes
* The entire `dynamic_message` module is behind a feature flag that will be disabled by default. Thus, there is no impact on compile times for the majority of users who do not need this.
* I still need to finish writing some more tests

# Future work:
* Factor out the common parts of `Subscription` and `DynamicSubscription` (and the same for publishers) – this can also help compilation times by monomorphizing Subscription.

### DIFF
--- a/examples/dynamic_pub_sub/Cargo.toml
+++ b/examples/dynamic_pub_sub/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "dynamic_pub_sub"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+rclrs = { version = "0.3", features = ["dyn_msg"] }
+anyhow = {version = "1", features = ["backtrace"]}
--- a/examples/dynamic_pub_sub/src/main.rs
+++ b/examples/dynamic_pub_sub/src/main.rs
@@ -1,0 +1,22 @@
+use anyhow::{Error, Result};
+use std::env;
+
+fn main() -> Result<(), Error> {
+    let context = rclrs::Context::new(env::args())?;
+
+    let mut node = rclrs::create_node(&context, "dynamic_subscriber")?;
+
+    let mut num_messages: usize = 0;
+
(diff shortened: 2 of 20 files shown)


nnmm:
@eduidl, you might be interested in this, since I know you wrote a client library for Rust as well as a command-line tool that uses message introspection :)

esteve:
@nnmm thanks for all the work, it's unfortunate we couldn't merge all the related PRs earlier, but if you ever come back to `ros2-rust`, we'd happy to review any future work. I'm closing this PR, but feel free to reopen it :slightly_smiling_face:

PR #266, opened 2022-09-24
Extend SubscriptionCallback trait to ReadOnlyLoanedMessage

nnmm:
With this change, loaned messages can now be used as the argument to subscription callbacks, just like regular messages.

This was not a straightforward extension because
1. `ReadOnlyLoanedMessage` has a lifetime. This required a higher-rank trait bound, which in turn required refactoring the `ArgTuple` trait a bit.
2. The `take_loaned_message()` method was restricted to `Subscription<T: RmwMessage>,` i.e. not available for subscriptions on idiomatic messages. This has been fixed and now the method is implemented for `Subscription<T: Message>` like the other `take*` methods – though you still will get an RMW-native loaned message even if `T` is the idiomatic message type, because idiomatic message types can never be loaned.

### DIFF
--- a/examples/minimal_pub_sub/src/zero_copy_subscriber.rs
+++ b/examples/minimal_pub_sub/src/zero_copy_subscriber.rs
@@ -12,7 +12,7 @@ fn main() -> Result<(), Error> {
     let _subscription = node.create_subscription::<std_msgs::msg::UInt32, _>(
         "topic",
         rclrs::QOS_PROFILE_DEFAULT,
-        move |msg: std_msgs::msg::UInt32| {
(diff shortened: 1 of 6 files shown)


christopherjreid (dismissed):
Everything here looks good to me!

jhdcs (dismissed):
Looks good to me!

esteve (commented):
@nnmm would it make sense to add an example in the examples package demonstrating this feature?

nnmm:
@esteve We already had an example, but I added a commit now to make it use the new callback signature.

jhdcs (approved):

PR #267, opened 2022-09-25
Fix compilation error in examples

soblin:
Hi when I build the workspace the examples gave following compile error:

```
error[E0632]: cannot provide explicit generic arguments when `impl Trait` is used in argument position
  --> src/zero_copy_subscriber.rs:12:52
   |
12 |     let _subscription = node.create_subscription::<std_msgs::msg::UInt32, _>(
   |                                                    ^^^^^^^^^^^^^^^^^^^^^  ^ explicit generic argument not allowed
   |                                                    |
   |                                                    explicit generic argument not allowed
   |
   = note: see issue #83701 <https://github.com/rust-lang/rust/issues/83701> for more information

For more information about this error, try `rustc --explain E0632`.
```

Looks like I could fix this by removing part of the type declarations.

I'm really new to Rust lang, so this PR may not be the best solution.

### DIFF
--- a/examples/message_demo/src/message_demo.rs
+++ b/examples/message_demo/src/message_demo.rs
@@ -143,20 +143,16 @@ fn demonstrate_pubsub() -> Result<(), Error> {
         rclrs::QOS_PROFILE_DEFAULT,
     )?;
 
-    let _idiomatic_subscription = node
-        .create_subscription::<rclrs_example_msgs::msg::VariousTypes, _>(
-            "topic",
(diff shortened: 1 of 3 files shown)


nnmm:
Hi, this error occurs because we require Rust 1.63 and your compiler is apparently older than that.

I think the proper solution would be to set the minimum required Rust version in Cargo.toml to 1.63.

soblin:
Thank you for your fat response!

PR #268, opened 2022-09-25
Require at least Rust 1.63

esteve:
This PR should prevent issues like #267 by requiring a minimum version of the Rust toolchain.


nnmm (approved):
Thanks. Looks good to me.

PR #269, opened 2022-09-28
Add support for constants to message generation

nnmm:
This will produce:

```
impl VariousTypes {
    /// binary, hexadecimal and octal constants are also possible
    pub const TWO_PLUS_TWO: i8 = 5;
    /// Only unbounded strings are possible
    pub const PASSWORD: &'static str = "hunter2";
    /// As determined by Edward J. Goodwin
    pub const PI: f32 = 3.0;
 }
 ```

Implementation notes:
* Comments on constants are not supported in Foxy, hence the `getattr`
* I changed `constant_value_to_rs` to operate on an rosidl type identifier, not a string. That was more straightforward both for the caller and the implementation of the function.
* I theoretically could have mapped strings to a static variable of type `BoundedString`, but having a constant of type `&str` seemed more useful.

### DIFF
--- a/rclrs_example_msgs/msg/VariousTypes.msg
+++ b/rclrs_example_msgs/msg/VariousTypes.msg
@@ -32,3 +32,10 @@ NestedType[2] nested_array
 NestedType[] nested_seq_unbounded
 NestedType[<=3] nested_seq_bounded
 
+
+# binary, hexadecimal and octal constants are also possible
+int8 TWO_PLUS_TWO = 5
+# Only unbounded strings are possible
+string PASSWORD = "hunter2"
+# As determined by Edward J. Goodwin
+float32 PI = 3.0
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: rclrs_example_msgs/msg/VariousTypes.msg [hunk 1]
```diff
@@ -32,3 +32,10 @@ NestedType[2] nested_array
 NestedType[] nested_seq_unbounded
 NestedType[<=3] nested_seq_bounded
 
+
+# binary, hexadecimal and octal constants are also possible
+int8 TWO_PLUS_TWO = 5
+# Only unbounded strings are possible
+string PASSWORD = "hunter2"
```
esteve, line 39: :rofl:

esteve (approved):
Great work! After this, the last remaining thing to be added in the generator are actions, IIRC

nnmm:
Thanks for the quick review! Yes, actions, and I'd also like to have comments for regular members.

PR #272, opened 2022-10-01
Format all code with group_imports = StdExternalCrate

nnmm:
This non-default `rustfmt` option was mentioned in https://blog.rust-lang.org/inside-rust/2022/09/29/announcing-the-rust-style-team.html. I think it's nice to have a consistent order for `use` statements.

I didn't create a `.rustfmt.toml` file with this setting because the setting requires nightly Rust, and I'm not sure if we want to require that for development.

### DIFF
--- a/examples/message_demo/src/message_demo.rs
+++ b/examples/message_demo/src/message_demo.rs
@@ -1,7 +1,7 @@
-use anyhow::{Error, Result};
 use std::convert::TryInto;
 use std::env;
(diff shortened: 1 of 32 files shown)


esteve (dismissed):

nnmm:
Could you reapprove? I needed to rebase after #249 was merged.

esteve (approved):

PR #276, opened 2022-10-04
Add WaitSet::new_for_node()

nnmm:
This function is important for users who directly use wait sets, since the `Node::live_subscriptions()` etc. functions are not public.


jhdcs (approved):
Seems simple enough. I'm assuming that there are some functions coming down the pipe that will make use of this?

nnmm:
Thanks! Not really, I just noticed that this function was needed while working on something that probably won't land in ros2-rust.

Issue #278, opened 2022-10-06
Make wait set growable


nnmm:
Currently, on each iteration of `rclrs::spin()`, the wait set is regenerated from scratch.

It would be more efficient to be able to push new waitable entities, if any, into the existing wait set – and re-allocating storage if the capacity is not sufficient. This could work a bit like a `Vec` – doubling capacity each time the capacity is exceeded.

PR #282, opened 2022-10-11
Fix a segfault when re-using wait set

nnmm:
`rcl_wait()` modifies the entries in `rcl_wait_set.subscriptions` so that some of them will be null pointers, causing a segfault if called a second time.

As a short term-fix, this makes it so that the wait set can only be used once.

The wait set should be made reusable as part of https://github.com/ros2-rust/ros2_rust/issues/278

### DIFF
--- a/rclrs/src/lib.rs
+++ b/rclrs/src/lib.rs
@@ -46,7 +46,7 @@ pub use wait::*;
 ///
 /// [1]: crate::RclReturnCode
 pub fn spin_once(node: &Node, timeout: Option<Duration>) -> Result<(), RclrsError> {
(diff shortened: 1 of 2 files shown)


esteve (dismissed):
@nnmm Good catch!

nnmm:
@esteve thanks, writing my own application on top of `rclrs` has been really helpful for finding bugs and usability issues like #276. Could you reapprove (after force-push for clippy)?

esteve:
@nnmm yeah, I agree, using an application to drive features is an excellent way of finding out what's missing and what can be improved

esteve (approved):

PR #283, opened 2022-10-14
Fix a bug in graph query function and add regression test

nnmm:
It's an easy mistake to make – in https://github.com/ros2/rcl/blob/rolling/rcl/include/rcl/graph.h, some functions have a `no_demangle` argument and others have a `no_mangle` argument. The `no_mangle` argument should be `true`, but it was false.


christopherjreid:
Whoops! :)

esteve (approved):

Issue #215, opened 2022-07-04
Use std::os::raw types instead of libc crate


nnmm:
It seems that the libc crate is only really useful in `#[no_std]` environments. https://stackoverflow.com/questions/44436515/should-i-use-libcc-char-or-stdosrawc-char

We should see if we can replace all uses of `libc` with `std::os::raw`.

Tacha-S:
I found libc crate use of the following items within this project.

- `c_void`
  - It is used in `*mut c_void`, in which case both [libc](https://docs.rs/libc/latest/libc/enum.c_void.html) and [std::os::raw](https://doc.rust-lang.org/nightly/std/os/raw/type.c_void.html) are equivalent to C's `void*`.
- `c_char`(Both libc and std are used.)
  - Exist in [std::os::raw](https://doc.rust-lang.org/nightly/std/os/raw/type.c_char.html).
- `c_ushort`
  - Exist in [std::os::raw](https://doc.rust-lang.org/nightly/std/os/raw/type.c_ushort.html).
- `uintptr_t`
  - Not implemented in `std::os::raw`, but may be [replaced by `usize`](https://rust-lang.github.io/unsafe-code-guidelines/layout/scalars.html#isize-and-usize).
- `realloc`
  - Not implemented in `std::os::raw`. Compatibility with C is not specified, but [std::alloc::realloc](https://doc.rust-lang.org/std/alloc/fn.realloc.html) is the implementation in Rust.
- `memcpy`
  - Not implemented in `std::os::raw`, but [`copy_nonoverlapping`](https://doc.rust-lang.org/std/ptr/fn.copy_nonoverlapping.html) is semantically equivalent to C `memcopy`.
- `size_t`
  - Not implemented in `std::os::raw`. However, since `--size_t_is_usize` is enabled in bindgen, it can be replaced by `usize`.([ref](https://internals.rust-lang.org/t/pre-rfc-usize-is-not-size-t/15369))

https://github.com/ros2-rust/ros2_rust/blob/0c262973b5abe0b8b66bca0285fae3ec59c0176f/rclrs/build.rs#L37

I consider `c_void`, `c_char`, `c_ushort`, `uintptr_t` to be fine, what about `realloc`, `memcpy`, `size_t`?

nnmm:
I agree about replacing the integer aliases like you say. We could also use `std::ffi` instead of `std::os::raw`, I actually think that's a bit nicer.

[...]

PR #284, opened 2022-10-16
Remove libc dependencies

Tacha-S:
- Replace libc crate with `std::ffi`
- Remove realloc and memcpy workaround https://github.com/ros2-rust/ros2_rust/issues/215#issuecomment-1279723455


esteve:
@Tacha-S this is looking great, thanks! Let us know when it's ready for review

Tacha-S:
I was developing with 1.64.
But unfortunately 1.63(CI version) requires core::ffi instead of std::ffi.

https://doc.rust-lang.org/src/std/ffi/mod.rs.html#157

Which version should be support?

nnmm:
I don't fully understand – the compiler error sounds like 1.63 has `core::ffi`, but the [release notes](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#c-compatible-ffi-types-in-core-and-alloc) and [`core::ffi` docs](https://doc.rust-lang.org/core/ffi/type.c_short.html) look like that module was introduced with `1.64`? Maybe these types were available but experimental in 1.63.

Either way, I think it would be nice to support 1.63 if it's not too ugly. Maybe that requires using `std::os::raw` after all.

Tacha-S:
[`core::ffi` was unstable](https://github.com/rust-lang/rust/pull/98315) before 1.64.0.

rosidl_generator_rs/resource/msg_rmw.rs.em
```diff
@@ -89,9 +90,8 @@ impl rosidl_runtime_rs::SequenceAlloc for @(type_name) {
     unsafe { @(package_name)__@(subfolder)__@(type_name)__Sequence__fini(seq as *mut _) }
   }
   fn sequence_copy(in_seq: &rosidl_runtime_rs::Sequence<Self>, out_seq: &mut rosidl_runtime_rs::Sequence<Self>) -> bool {
-      out_seq.resize_to_at_least(in_seq.len());
-      out_seq.clone_from_slice(in_seq.as_slice());
-      true
+    // SAFETY: This is safe since a the point is guaranteed to be valid/initialized.
```
jhdcs, line 93:
```suggestion
    // SAFETY: This is safe since the pointer is guaranteed to be valid/initialized.
```

jhdcs (changes requested):
Looks pretty good to me! I only saw one spot where a comment's phrasing seemed a little odd.

That being said, a second pair of eyes would be useful, I think. Just in case I missed something.

rosidl_runtime_rs/src/traits.rs
```diff
@@ -42,7 +42,7 @@ pub trait RmwMessage: Clone + Debug + Default + Send + Sync + Message {
     const TYPE_NAME: &'static str;
 
     /// Get a pointer to the correct `rosidl_message_type_support_t` structure.
-    fn get_type_support() -> libc::uintptr_t;
+    fn get_type_support() -> usize;
```
nnmm, line 45: While we're here – could we make this pointer and the one in the `Service` trait a `*const c_void`?

jhdcs (approved):
Looks good to me!

@nnmm does everything look okay to you?

nnmm (approved):
Nice work, thank you! :)

PR #287, opened 2022-10-27
Parsing of the message structure from introspection typesupport library

nnmm:
This module provides a nice tree representation of the message structure.


jhdcs (approved):
This looks all right to me. Thank you!

nnmm:
Wonderful @jhdcs, thanks for the quick review! I'll leave this open for one or two days to give @esteve a chance to see it too.

esteve:
@nnmm I won't be able to review this until next week or maybe even later. Given that this PR is not a priority, would you be ok with waiting until then? Thanks

nnmm:
@esteve Sure. I'll probably open one or more other PRs for this feature that you can batch review when you have time.

nnmm:
I forgot that I had written a test for this. Added now.

rclrs/src/dynamic_message/message_structure.rs
```diff
@@ -0,0 +1,254 @@
...
+    // That function must be unsafe, since it is possible to safely create a garbage non-null
+    // pointer.
+    unsafe fn new(
+        type_id: u8,
+        string_upper_bound: Option<NonZeroUsize>,
+        inner: *const rosidl_message_type_support_t,
+    ) -> Self {
+        #[cfg(all(not(ros_distro = "foxy"), not(ros_distro = "galactic")))]
+        use rosidl_typesupport_introspection_c_field_types::*;
+        match u32::from(type_id) {
```
esteve, line 123: Wouldn't this look cleaner with a lookup table?

rclrs_tests/src/lib.rs
```diff
@@ -1,5 +1,8 @@
 #![cfg(test)]
 
 mod client_service_tests;
+// Disabled in Foxy due to https://github.com/ros2/rosidl/issues/598
+#[cfg(all(not(ros_distro = "foxy"), not(ros_distro = "galactic")))]
```
esteve, line 5: Why was this disabled for Galactic? It seemed that only Foxy was affected

esteve (changes requested):

rclrs/src/dynamic_message/message_structure.rs
nnmm: Hmm probably! I'll try it out.

rclrs_tests/src/lib.rs
nnmm: No, actually also Galactic. I was wrong at first – should correct the comment.

rclrs/src/dynamic_message/message_structure.rs
nnmm: Two issues with that:
* The lookup table would be brittle, because the order of entries has to match the order of variants in the enum, but there's no clean way to check that it does without incurring overhead afaik
* The string, sequence and message cases are not a pure lookup, but have logic in their match arms. So the lookup table would have to exclude those and handle them specially, which would make this function look less clean again.

So I'd leave it as it is.

rclrs_tests/src/lib.rs
nnmm: As proof, I removed the galactic part temporarily: https://github.com/ros2-rust/ros2_rust/actions/runs/3452502455/jobs/5762451384

nnmm:
@esteve could you take another look please?

rclrs_tests/src/lib.rs
esteve: In that case, can you update the comment to mention `galactic` as well, instead of only `foxy`? Thanks

rclrs/src/dynamic_message/message_structure.rs
esteve: Fair enough. The code as it is is not the prettiest, but does the job.

rclrs_tests/src/lib.rs
esteve: @nnmm can you update the comment here? I can push the change, but given that this PR is part of a bigger PR, I thought of leaving it to you.

esteve:
@nnmm closing for now since we have had time to revisit this PR in a long time. Let us know if you'd like to work on it in the future. Thanks!

PR #288, opened 2022-10-27
Vendor rcl_interfaces as a private module into rclrs

nnmm:
The parameter services require the rcl_interfaces message and service types. I think this option is preferable to publishing rcl_interfaces on crates.io.

The script included in this PR produces the `vendor` module inside `rclrs` by copying the generated code for the `rcl_interfaces` package and its dependency `builtin_interfaces` and adjusting the submodule paths in the code.
If these packages, or the `rosidl_generator_rs` package, get changed, we can update the `vendor` module by running this script.

### DIFF
--- a/rclrs/package.xml
+++ b/rclrs/package.xml
@@ -16,6 +16,8 @@
   <build_depend>libclang-dev</build_depend>
   <build_depend>rosidl_runtime_rs</build_depend>
   <build_depend>rcl</build_depend>
+  <depend>builtin_interfaces</depend>
+  <depend>rcl_interfaces</depend>
 
   <export>
     <build_type>ament_cargo</build_type>
--- a/rclrs/src/lib.rs
+++ b/rclrs/src/lib.rs
@@ -15,6 +15,7 @@ mod publisher;
 mod qos;
 mod service;
 mod subscription;
+mod vendor;
 mod wait;
 
 mod rcl_bindings;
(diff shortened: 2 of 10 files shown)


esteve:
LGTM, but I'll leave it to @jhdcs to approve it as I see that CI is failing.

@nnmm can you update the description of this PR and add a docstring to the Python script with a description of what it does and how to use it? Thanks

For a future PR, or maybe even this one, I'd add the test_interfaces messages so that we can integrate the tests here, I prefer tests to be in the same crate/package, they are easier to maintain and also serve as examples of how to use the API

nnmm:
@esteve I had planned to just create an alias `ExampleMessage` for one of the `rcl_interfaces` messages that we can use in doctests. For most tests, the contents of the message being used don't matter. So I propose to do that for now, and if/when we see a need for `test_msgs` specifically, we can consider vendoring that too.

jhdcs (dismissed):
This looks fine to me as well, though I don't see a failing CI. Was that a previous commit?

nnmm:
@jhdcs Yes, that was a previous commit. I've also added the comment requested by @esteve.

nnmm:
@jhdcs Could you reapprove this? I've added a commit to add these dependencies to `package.xml`, since we will still depend on the C libraries from these packages.

jhdcs (approved):

nnmm:
Thanks!

PR #289, opened 2022-10-28
Rename *Base traits to *Waitable

nnmm:
This is better in my opinion for a few reasons:
* "Base" suggests inheritance, but unlike in `rclcpp`, there is no inheritance in `rclrs`
* The functionality provided by e.g. `SubscriptionBase` in `rclcpp` and `rclrs` is very different, so using the same name for both suggests an equivalence that isn't there
* `*Waitable` is more descriptive, since the purpose of these traits is to interface subscriptions, services, and clients with wait sets

### DIFF
--- a/rclrs/src/client.rs
+++ b/rclrs/src/client.rs
@@ -43,7 +43,7 @@ impl Drop for ClientHandle {
 /// Trait to be implemented by concrete Client structs.
 ///
 /// See [`Client<T>`] for an example.
-pub trait ClientBase: Send + Sync {
+pub trait ClientWaitable: Send + Sync {
     /// Internal function to get a reference to the `rcl` handle.
     fn handle(&self) -> &ClientHandle;
     /// Tries to take a new response and run the callback or future with it.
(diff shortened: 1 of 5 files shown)


jhdcs (commented):
I think that the change makes sense, but would like @esteve to weigh in before approving anything.

nnmm:
@esteve Could you take a look here?

esteve:
@nnmm I get the point about moving away from `*Base` to avoid any confusion with `rclcpp`, but `*Waitable` doesn't really strike me as the right one, because `*Waitable` sounds as if they can be waited on directly (I know, they interact with `waitset`s), but the methods they describe for example for `SubscriptionWaitable` are `handle` and `execute`. I'd prefer something like `SubscriptionExecutable`, `SubscriptionRunnable`, `SubscriptionTrigger` or similar, probably `SubsciptionRunnable` would be my preference.

esteve:
Closing. We might want to revisit this again in the future. Thanks @nnmm

PR #290, opened 2022-10-28
Make constructor methods for pub/sub/etc. take rcl_node_t mutex

nnmm:
The constructors are not public, so we're free to use internal types as arguments.

This signature makes it possible, or at least easier, to write parameter services that are owned by the node: With this change, the Node does not need to be constructed before the parameter services are constructed, which would somewhat contradict the node owning these services.


jhdcs (approved):
The changes look good to me!

esteve (approved):
@nnmm LGTM

Issue #246, opened 2022-08-14
Add Extend and FromIterator impl for rosidl_runtime_rs::{String, WString}


nnmm:
It would be nice if the string types in `rosidl_runtime_rs` could be built/extended from iterators, [like the regular string type](https://doc.rust-lang.org/std/string/struct.String.html#impl-Extend%3Cchar%3E).

For this, we'd need implementations for `Extend<char>` and `FromIterator<char>` for both types, at least.

maspe36:
An important thing to note, is that this feature gives the string types interior mutability. So these string types would have to lose the Sync impl.

https://github.com/ros2-rust/ros2_rust/blob/main/rosidl_runtime_rs/src/string.rs#L230

nnmm:
I think we wouldn't need to lose the `Sync` marker, since the `Extend` trait takes `&mut self` and not `&self`.

maspe36:
Ahh, gotcha. Yeah even rust's `String` is `Sync` 
https://doc.rust-lang.org/std/string/struct.String.html#impl-Sync-for-String

Updated in the PR

maspe36:
I believe this issue should be closed now that #293 has been merged

PR #293, opened 2022-11-11
Extend string types

maspe36:
Implemented `Extend<char>`, `Extend<&'a char>`, `FromIterator<char>`, and `FromIterator<&'a char>` for String and WString.

Of note here is that with this change, String and WString now have interior mutability so they are no longer able to implement `Sync` safely.

Original Issue: https://github.com/ros2-rust/ros2_rust/issues/246

### DIFF
--- a/rosidl_runtime_rs/src/string.rs
+++ b/rosidl_runtime_rs/src/string.rs
@@ -3,6 +3,8 @@ use std::ffi::CStr;
 use std::fmt::{self, Debug, Display};
 use std::hash::{Hash, Hasher};
 use std::ops::{Deref, DerefMut};
+use std::ptr;
+use std::ptr::slice_from_raw_parts;
 
 #[cfg(feature = "serde")]
 mod serde;
@@ -135,7 +137,7 @@ macro_rules! string_impl {
                 };
                 // SAFETY: Passing in a zeroed string is safe.
                 if !unsafe { $init(&mut msg as *mut _) } {
-                    panic!("Sinit failed");
+                    panic!("$init failed");
                 }
                 msg
             }
@@ -224,10 +226,53 @@ macro_rules! string_impl {
             }
         }
 
+        impl FromIterator<char> for $string {
+            fn from_iter<I: IntoIterator<Item = char>>(iter: I) -> Self {
+                let mut buf = <$string>::default();
+                buf.extend(iter);
+                buf
+            }
+        }
+
+        impl<'a> FromIterator<&'a char> for $string {
+            fn from_iter<I: IntoIterator<Item = &'a char>>(iter: I) -> Self {
+                let mut buf = <$string>::default();
+                buf.extend(iter);
+                buf
+            }
+        }
+
+        impl Extend<char> for $string {
+            fn extend<I: IntoIterator<Item = char>>(&mut self, iter: I) {
+                let mut v = self.to_vec();
+                let iterator = iter.into_iter();
+
+                iterator.for_each(|c| {
+                    v.push(c as $char_type);
+                });
+
+                // SAFETY: It's okay to pass a non-zero-terminated string here since assignn
+                // uses the specified length and will append the 0 to the dest string itself.
+                if !unsafe {
+                    $assignn(
+                        self as *mut _,
+                        v.as_ptr() as *const _,
+                        v.len(),
+                    )
+                } {
+                    panic!("$assignn failed");
+                }
+            }
+        }
+
+        impl<'a> Extend<&'a char> for $string {
+            fn extend<I: IntoIterator<Item = &'a char>>(&mut self, iter: I) {
+                self.extend(iter.into_iter().cloned());
+            }
+        }
+
         // SAFETY: A string is a simple data structure, and therefore not thread-specific.
         unsafe impl Send for $string {}
-        // SAFETY: A string does not have interior mutability, so it can be shared.
-        unsafe impl Sync for $string {}
 
         impl SequenceAlloc for $string {
             fn sequence_init(seq: &mut Sequence<Self>, size: usize) -> bool {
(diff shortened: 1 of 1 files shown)


rosidl_runtime_rs/src/string.rs
```diff
@@ -135,7 +137,7 @@ macro_rules! string_impl {
                 };
                 // SAFETY: Passing in a zeroed string is safe.
                 if !unsafe { $init(&mut msg as *mut _) } {
-                    panic!("Sinit failed");
+                    panic!("$init failed");
```
nnmm, line 140: Good catch!

rosidl_runtime_rs/src/string.rs
```diff
@@ -224,10 +226,47 @@ macro_rules! string_impl {
...
+                iterator.for_each(|c| {
+                    v.push(c as $char_type);
+                });
+
+                // SAFETY: It's okay to pass a non-zero-terminated string here since assignn
+                // uses the specified length and will append the 0 to the dest string itself.
+                if !unsafe { $assignn(self as *mut _, v.as_ptr() as *const _, v.len()) } {
+                    panic!("$assignn failed");
+                }
+            }
```
nnmm, line 259: The strings are UTF-8/UTF-16, but this implementation only works for ASCII strings – you can't cast a `char` to `$char_type`. One way to do it properly would be to get a `std::string::String` from `$string`, extend that, and convert it back to the `$string` type. It might be possible to still do better than that (performance-wise), but that would be a good start.

nnmm (commented):
Thanks for your PR! I left a comment.

rosidl_runtime_rs/src/string.rs
maspe36: Oops, good call. Fixed

rosidl_runtime_rs/src/string.rs
```diff
@@ -224,6 +225,59 @@ macro_rules! string_impl {
...
+
+                    let filtered_bytes: Vec<$unsigned_char_type> = buf
+                        .into_iter()
+                        .filter(|&c| c != (0 as $unsigned_char_type))
+                        .collect();
+
+                    for encoded_char in filtered_bytes {
+                        v.push(encoded_char as $char_type);
+                    }
+                });
```
nnmm, line 265: You could probably simplify this to something like 
```
let mut s = self.to_string();
s.extend(iter);
*self = Self::from(&s);
```

This should be more performant too (partly because `String::extend()` reserves an appropriate amount of space upfront if that information is available from the iterator's `size_hint()` function). It would be possible to get even more efficient and eliminate the back-and-forth conversion, but that requires dropping into some quite low-level code, so I'm also fine with not doing that. I'll leave it up to you whether you're interested in that or not.

maspe36: Ah yeah this is better. Simplified implementation, a few extra ergonomic trait impls for the String types, and as you mentioned likely more performant. Thanks for the help :+1:

rosidl_runtime_rs/src/string.rs
```diff
@@ -224,6 +225,42 @@ macro_rules! string_impl {
             }
         }
 
+        impl From<&std::string::String> for $string {
```
nnmm, line 228: Since the existing `From<&str>` impl is more general, I think this impl isn't necessary.

rosidl_runtime_rs/src/string.rs
```diff
@@ -112,7 +113,7 @@ pub struct StringExceedsBoundsError {
 
 // There is a lot of redundancy between String and WString, which this macro aims to reduce.
 macro_rules! string_impl {
-    ($string:ty, $char_type:ty, $unsigned_char_type:ty, $string_conversion_func:ident, $init:ident, $fini:ident, $assignn:ident, $sequence_init:ident, $sequence_fini:ident, $sequence_copy:ident) => {
+    ($string:ty, $char_type:ty, $unsigned_char_type:ty, $string_conversion_func:ident, $encoding_func:ident, $init:ident, $fini:ident, $assignn:ident, $sequence_init:ident, $sequence_fini:ident, $sequence_copy:ident) => {
```
nnmm, line 116: Could you remove the `$encoding_func` again?

rosidl_runtime_rs/src/string.rs
```diff
@@ -3,6 +3,7 @@ use std::ffi::CStr;
 use std::fmt::{self, Debug, Display};
 use std::hash::{Hash, Hasher};
 use std::ops::{Deref, DerefMut};
+use std::ptr;
```
nnmm, line 6: I think this can be removed too – it's more consistent with the rest of the code to use the full path below.

rosidl_runtime_rs/src/string.rs
```diff
@@ -224,6 +225,42 @@ macro_rules! string_impl {
...
+
+        impl<'a> FromIterator<&'a char> for $string {
+            fn from_iter<I: IntoIterator<Item = &'a char>>(iter: I) -> Self {
+                let mut buf = <$string>::default();
+                buf.extend(iter);
+                buf
+            }
+        }
+
+        impl Extend<char> for $string {
```
nnmm, line 250: Small nitpick: I think it's nice to have trait impls in alphabetical order for readability. Not important though.

nnmm (commented):
LGTM, just some small changes to tidy up are remaining.

rosidl_runtime_rs/src/string.rs
maspe36: Done

rosidl_runtime_rs/src/string.rs
maspe36: Done

rosidl_runtime_rs/src/string.rs
maspe36: So in `*self = Self::from(&s);`, `&s` is actually of type `&String` and `From` isn't implemented for that type. So I removed this impl by just using `Self::from(s.as_str())` in the `Extend` impl

rosidl_runtime_rs/src/string.rs
maspe36: Sure thing, alphabetized all the trait impl's in this module.

rosidl_runtime_rs/src/string.rs
```diff
@@ -126,21 +126,6 @@ macro_rules! string_impl {
             ) -> bool;
         }
 
-        impl Default for $string {
```
maspe36, line 129: Alphabetizing the trait impl's

rosidl_runtime_rs/src/string.rs
nnmm: That's correct, though using `Self::from(&s)` should work too due to [Deref coercion](https://doc.rust-lang.org/book/ch15-02-deref.html#implicit-deref-coercions-with-functions-and-methods).

nnmm (approved):
Thanks for making these changes!

maspe36:
>Thanks for making these changes!

Thanks for all the guidance :)

PR #294, opened 2022-11-27
Return Arc from the create_node function to match other create_X functions

esteve:
This PR makes `create_node` consistent with other `create_X` functions by return an `Arc`. Additionally added fine-grained locks so that `create_X` functions don't need to take a mutable reference to `self`.

Returning an `Arc` is also akin to `rclcpp`'s behavior, which returns a `shared_ptr` from `create_node`.


jhdcs (approved):
These changes look good to me! Thank you very much!

esteve:
@jhdcs thanks for the prompt review!

PR #297, opened 2022-12-12
Swapped usage of rosidl_cmake over to the new rosidl_pycommon.

maspe36:
As of [rosidl 3.3.0](https://github.com/ros2/rosidl/commit/9348ce9b466335590dc334aab01f4f0dd270713b) (specifically, this [commit](https://github.com/ros2/rosidl/pull/696/files#diff-330d15435e7274b3796bb084ab591a6be34c8afeb8acdcc90e93074cc33bcc12)), the rosidl_cmake Python module was moved to a new rosidl_pycommon package and the Python module in rosidl_cmake was deprecated.

This was throwing these deprecation warnings when building for rolling. 
```
Starting >>> builtin_interfaces
Starting >>> unique_identifier_msgs
--- stderr: unique_identifier_msgs                                                                                 
/opt/ros/rolling/local/lib/python3.10/dist-packages/rosidl_cmake/__init__.py:19: UserWarning: The 'rosidl_cmake' Python module is deprecated. Use 'rosidl_pycommon' instead.
  warnings.warn(
---
Finished <<< unique_identifier_msgs [2.73s]
--- stderr: builtin_interfaces                               
/opt/ros/rolling/local/lib/python3.10/dist-packages/rosidl_cmake/__init__.py:19: UserWarning: The 'rosidl_cmake' Python module is deprecated. Use 'rosidl_pycommon' instead.
  warnings.warn(
---
Finished <<< builtin_interfaces [2.92s]
```


rosidl_generator_rs/package.xml
```diff
@@ -13,7 +13,8 @@
   <buildtool_depend>rosidl_runtime_rs</buildtool_depend>
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
-  <buildtool_export_depend>rosidl_cmake</buildtool_export_depend>
+  <buildtool_export_depend condition="$ROS_DISTRO != rolling">rosidl_cmake</buildtool_export_depend>
```
maspe36, line 16: I may be interpreting [REP 149](https://www.ros.org/reps/rep-0149.html#id20) incorrectly, but to me it reads as if I should be able to use a less/greater than for ROS distros. Unfortunately that doesn't work, and the xml fails to parse. Not sure what to do if rosidl_cmake 3.3.0+ gets back ported to Humble or when Iron comes out

nnmm: Reads that way to me too. Did you try only `>` or also `>=`?

nnmm:
This would fit the bill for `package.xml`, right?

```
<buildtool_export_depend condition="$ROS_DISTRO == foxy or $ROS_DISTRO == galactic or $ROS_DISTRO == humble">rosidl_cmake</buildtool_export_depend>
<buildtool_export_depend condition="$ROS_DISTRO != foxy and $ROS_DISTRO != galactic and $ROS_DISTRO != humble">rosidl_pycommon</buildtool_export_depend>
```

Then in CMake we could do something a little more elegant like @esteve proposed.

rosidl_generator_rs/package.xml
maspe36: Coming back to this now, looks to only be a problem with less than operators.

Fails to parse 
`$ROS_DISTRO <= rolling`
`$ROS_DISTRO < rolling`

Successfully parses 
`$ROS_DISTRO >= rolling`
`$ROS_DISTRO > rolling`

Thankfully this means we can handle this a bit more concisely in the `package.xml`

```xml
<buildtool_export_depend condition="rolling > $ROS_DISTRO">rosidl_cmake</buildtool_export_depend>
<buildtool_export_depend condition="$ROS_DISTRO >= rolling">rosidl_pycommon</buildtool_export_depend>
```
And then we can swap the usage of `rolling` to `iron` once its out.

maspe36:
> The changes assume that rolling is the next ROS distro, however it's more akin to main in a GitHub repo that gets tagged and release, so these changes won't work with the next ROS distro iron

The next patch set doesn't assume this and I don't think will break when `iron` is released.

> create a new branch for humble and make main target iron and greater, and backport any changes to humble, galactic, etc.

I think eventually this may need to happen, but for now that's probably more work than its worth. Especially since we can have optional dependencies in the `package.xml` based on distro.

> check the version of one of the rosidl base packages in CMake (e.g. rosidl_parser) and if it's greater or equal than 3.3.0, then don't call find_package on rosidl_cmake

Is this really better than doing the same check from the `package.xml` in the CMake?

```cmake
if("$ENV{ROS_DISTRO}" STRLESS_EQUAL "humble")
    find_package(rosidl_cmake REQUIRED)
    ament_export_dependencies(rosidl_cmake)
endif()
```
Only benefit I see would be if it gets backported but since this is a deprecation I think @nnmm is right, this probably won't get backported.

maspe36:
Look like the build is failing for the same reason mentioned [here](https://github.com/ros2-rust/ros2_rust/pull/300)

nnmm (dismissed):
From my understanding, this looks good. I'll wait a little bit before merging to give others the chance to take a look too.

rosidl_generator_rs/package.xml
```diff
@@ -13,7 +13,8 @@
   <buildtool_depend>rosidl_runtime_rs</buildtool_depend>
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
-  <buildtool_export_depend>rosidl_cmake</buildtool_export_depend>
+  <buildtool_export_depend condition="humble >= $ROS_DISTRO">rosidl_cmake</buildtool_export_depend>
+  <buildtool_export_depend condition="$ROS_DISTRO > humble">rosidl_pycommon</buildtool_export_depend>
   <buildtool_export_depend>rosidl_runtime_rs</buildtool_export_depend>
   <buildtool_export_depend>rosidl_typesupport_c</buildtool_export_depend>
   <buildtool_export_depend>rosidl_typesupport_interface</buildtool_export_depend>
```
esteve, line 20: Please add `buildtool_depend` and `buildtool_export_depend` for `ros_environment`, which provides access to the `ROS_DISTRO` environment variable. Although usually when building locally `ROS_DISTRO` is available, it's not guaranteed unless `ros_environment` is added as a dependency.

esteve (changes requested):

rosidl_generator_rs/package.xml
maspe36: Done

nnmm:
@maspe36 Not quite "tonight" as promised, but I have an MR open with the fix, just waiting for review by a maintainer: https://github.com/ros2-rust/ros2_rust/pull/309

nnmm:
@maspe36 If you rebase this onto main, CI should be green.

nnmm (commented):
Don't we also need `rosidl_pycommon` in `rosidl_generator_rs/CMakeLists.txt` too? Otherwise LGTM, but I can't merge before @esteve revisits his requested changes.

maspe36:
@esteve @nnmm [rosidl_pycommon](https://github.com/ros2/rosidl/tree/rolling/rosidl_pycommon) is a python only package, you cannot find it with `find_package`, and you cannot export it with `ament_export_dependencies`. Having it in the `package.xml` should be sufficient for rosdep

nnmm (approved):
Sounds good to me. Thanks for taking care of this, and sorry it took forever to merge!

esteve (approved):

esteve:
@maspe36 thanks again for all the patience! :slightly_smiling_face:

Issue #214, opened 2022-07-04
Optimize Serde implementations for rosidl_runtime_rs types


nnmm:
Particularly:
- The [String Deserialize](https://github.com/ros2-rust/ros2_rust/blob/a351772cea9a84770afc4a597e55b2c735c1569a/rosidl_runtime_rs/src/string/serde.rs#L11) implementation first creates a Rust String, but it could potentially directly construct a `rosidl_runtime_rs::String`. Same for `WString` and the sequences types (which use a temporary `Vec`).
- The [String Serialize](https://github.com/ros2-rust/ros2_rust/blob/a351772cea9a84770afc4a597e55b2c735c1569a/rosidl_runtime_rs/src/string/serde.rs#L21) implementation also creates a Rust String even though it wouldn't need to.

The sequence serialization implementations are already as efficient as they can be, I believe.

This is a tricky task that is best suited for someone who's quite familiar with the Serde data model, so I'm adding the needs-rust-expert label.

PR #300, opened 2023-02-18
Improve efficiency of deserializing strings

nnmm:
This addresses part of https://github.com/ros2-rust/ros2_rust/issues/214.

I recommend watching the excellent https://www.youtube.com/watch?v=BI_bHCGRgMY tutorial. I mostly followed the implementation of the standard library `String` type.

### DIFF
--- a/rosidl_runtime_rs/src/string/serde.rs
+++ b/rosidl_runtime_rs/src/string/serde.rs
@@ -1,13 +1,98 @@
-use serde::{de::Error, Deserialize, Deserializer, Serialize, Serializer};
+use std::fmt;
 
-use super::{BoundedString, BoundedWString, String, WString};
+use serde::{
+    de::{Error, SeqAccess, Visitor},
+    Deserialize, Deserializer, Serialize, Serializer,
+};
+
+use super::{
+    rosidl_runtime_c__String__assignn, rosidl_runtime_c__U16String__assignn, BoundedString,
+    BoundedWString, String, WString,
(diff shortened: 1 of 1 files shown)


esteve:
@nnmm cool! Looks like there's an issue with the generated Python files for `rclrs_example_msgs`, but it's not cause by this PR, might be a problem in `rolling`.

nnmm:
@esteve yeah. Unfortunately the workaround suggested in https://github.com/ros-tooling/action-ros-ci/issues/795 didn't seem to actually work around, so we'd have to wait for the next Rolling sync. Can we somehow make CI allow failure (for this specific check or entirely) until then?

esteve:
@nnmm looks like that's a feature that users have requested (https://github.com/orgs/community/discussions/15452), I don't know of any other way to allow a job to fail. I think the best we can do now is disable the `rolling` CI job until the next sync.

esteve:
@nnmm never mind, looks like `continue-on-error` might work in this case: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

esteve:
I've created https://github.com/ros2-rust/ros2_rust/pull/302 to allow the `rolling` job to fail, let's see if it works.

nnmm:
@esteve Could you take another look now that CI is green?

esteve (approved):
@nnmm thanks!

PR #302, opened 2023-02-26
Workaround for ament_lint issues. Run CI every night.

esteve:


nnmm:
Looks like the same error happens also on humble.

esteve:
@nnmm @jhdcs It seems to be caused by `pydocstyle` >= 6.2.0 https://github.com/ament/ament_lint/pull/428, which gets installed by `ros-tooling/setup-ros` in https://github.com/ros-tooling/setup-ros/blob/master/src/package_manager/pip.ts#L51

Options I can think of:
 - disable `ament_pep257` for the `rclrs_example_msgs` package temporarily (https://github.com/ros2-rust/ros2_rust/blob/main/rclrs_example_msgs/CMakeLists.txt#L22)
 - manually build https://github.com/ament/ament_lint and install it after running `ros-tooling/setup-ros`
 - manually patch `ament_pep257/ament_pep257/main.py` to update it for the new `pydocstyle` version

3 seems easiest, 2 would be the most "correct" one (but every CI run will take longer) and I'd try to avoid 1

nnmm:
@esteve Have you seen https://github.com/ament/ament_lint/pull/428? I think it has already been fixed, we just need to wait for a Rolling sync: https://github.com/ros-tooling/action-ros-ci/issues/795#issuecomment-1430211489

nnmm:
I did try using the prerelease/testing APT repo as suggested in the latter link, but it didn't work.

jhdcs:
My personal vote would be going for the most correct option - however, how much time would this be adding to CI?

esteve:
> My personal vote would be going for the most correct option - however, how much time would this be adding to CI?

This would mean rebuilding `ament_lint`, which won't take long (it's just a bunch of CMake scripts), and I'm assuming that installing all its development dependencies won't take long either, I don't know for sure, but it's worth a try.

esteve:
> I did try using the prerelease/testing APT repo as suggested in the latter link, but it didn't work.

Yeah, that's weird, there are in fact .debs with newer releases of `ament_pep257` for both `humble` and `rolling` in the testing repo:

http://packages.ros.org/ros2-testing/ubuntu/pool/main/r/ros-rolling-ament-cmake-pep257/
http://packages.ros.org/ros2-testing/ubuntu/pool/main/r/ros-humble-ament-cmake-pep257/

I don't know why they don't get installed. I've updated `setup-ros` to v0.6 in case that's why the action was not picking up the testing repository.

esteve:
`setup-ros@v0.6` dropped support for Galactic as it's no longer supported by Open Robotics (support ended in November https://www.ros.org/reps/rep-2000.html#galactic-geochelone-may-2021-november-2022)

esteve:
Unfortunately, until `ament_lint` gets tagged for `foxy` and the latest release includes https://github.com/ament/ament_lint/pull/431, the `foxy` job will continue failing. I filed https://github.com/ament/ament_lint/issues/434, so hopefully we can use the `ros-testing` repo with `foxy` at some point.

nnmm:
@esteve There is also another error in the rolling build:

```
Starting >>> rclrs_example_msgs
  --- output: rclrs
  error[E0308]: mismatched types
     --> src/subscription/message_info.rs:120:19
      |
  120 |             data: rmw_message_info.publisher_gid.data,
      |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected an array with a fixed size of 24 elements, found one with 16 elements
```

esteve:
@nnmm yeah, saw that, but I'm more interested now in getting `humble` and `foxy` passing. Looks like the `publisher_gid` field got updated to use 16 bytes, instead of 24, in rolling (see https://github.com/ros2/rmw/pull/345)

nnmm (approved):
LGTM!

nnmm:
I hope it's fine if I merge your PR this time @esteve, to unblock other PRs.

esteve:
@nnmm sure! Thanks for merging this PR :slightly_smiling_face:

PR #308, opened 2023-03-12
Use ros2-testing debs for rolling

esteve:
This PR uses the Debian packages from the testing repository for rolling so we can catch API incompatibilities earlier. Given that the rolling job is marked as `continue-on-error` and so it won't cause the entire CI fail, in case there are any build errors, PRs can still be merged.


nnmm (approved):
Sounds good to me!

PR #311, opened 2023-05-11
Remove workaround for pydocstyle

esteve:
Now that there's a new tagged version of `ament_lint` (https://github.com/ament/ament_lint/releases/tag/0.9.8), this should no longer be needed.


esteve:
No longer needed, already in `main` via #321

Issue #320, opened 2023-07-09
`extern` block uses type `u128`, which is not FFI-safe


maspe36:
This new warning is causing CI to fail 

```
warning: `extern` block uses type `u128`, which is not FFI-safe
    --> /home/sam/rust_ws/src/ros2_rust/rclrs_tests/target/debug/build/rclrs-8555e18c0c9bb180/out/rcl_bindings_generated.rs:6343:31
     |
6343 |         dynamic_type_builder: *mut rosidl_dynamic_typesupport_dynamic_type_builder_t,
     |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not FFI-safe
     |
     = note: 128-bit integers don't currently have a known stable ABI

```

Seems to be coming from auto-generated idl code. Investigate and hopefully resolve

esteve:
It's interesting that this only happens with `rolling`. As I see it, we have two options:

 - disable the dynamic message feature (`dyn_msg`) from the `clippy` check
 - disable the warning about `u128` (via `improper-ctypes`) from the `clippy` check

I believe that disabling the check from `clippy` should be fine in our case, as we're not exposing this API.

@maspe36 what do you think?

maspe36:
This is also happening with `iron` in my PR to add support for it. I'm trying to hunt down the root cause and depending on that it may indeed make sense to disable the warning or dynamic message feature from clippy.

The first build I found where this was failing the rolling [build #974](https://github.com/ros2-rust/ros2_rust/actions/runs/4998294792/jobs/8953551442). 

Current theory is that this is related to the version uptick in rosidl_generator_c from 3.3.1 to 4.1.1 as that is a major version change between builds #974 and [#973](https://github.com/ros2-rust/ros2_rust/actions/runs/4986815835/jobs/8927975778). I'll keep digging

maspe36:
Looks like this is cause by the new [Runtime Interface Reflection feature (#215)](https://github.com/ros2/rmw_implementation/pull/215) in the [rmw_implementation](https://github.com/ros2/rmw_implementation) package. 

[...]

PR #321, opened 2023-07-13
Update Rust toolchain to 1.71.0 and remove pydocstyle workaround

esteve:


esteve:
@maspe36 I used the workaround I suggested in https://github.com/ros2-rust/ros2_rust/issues/320#issuecomment-1633977366 and it seems to build fine locally. Let me know if you prefer the changes in PR to be split into separate PRs, I think it'd be cleaner, but given that I was already fixing `clippy` warnings I thought of putting adding the fixes all together.

esteve:
@maspe36 never mind, I've created #322 that only addresses #320 and disables the `clippy` warning

rclrs/src/vendor/rcl_interfaces/msg.rs
```diff
@@ -404,6 +404,7 @@ pub mod rmw {
         pub description: rosidl_runtime_rs::String,
         pub additional_constraints: rosidl_runtime_rs::String,
         pub read_only: bool,
+        pub dynamic_typing: bool,
```
maspe36, line 407: Does this break backwards compatibility with humble?

maspe36 (changes requested):
The [Dockerfile](https://github.com/ros2-rust/ros2_rust/blob/main/Dockerfile#L15) also has a reference to rust 1.63 that should probably be upgraded to 1.70

esteve:
> The [Dockerfile](https://github.com/ros2-rust/ros2_rust/blob/main/Dockerfile#L15) also has a reference to rust 1.63 that should probably be upgraded to 1.70

Done. I updated it to 1.71.0, which was released yesterday.

rclrs/src/vendor/rcl_interfaces/msg.rs
esteve: Humble does have a `dynamic_typing` field as well (https://github.com/ros2/rcl_interfaces/blob/humble/rcl_interfaces/msg/ParameterDescriptor.msg#L25), so this should be ok.

maspe36 (approved):

esteve:
@maspe36 thanks for reviewing these changes :slightly_smiling_face:

PR #322, opened 2023-07-13
Disable clippy warnings for u128 types

### LINKED ISSUE #320: `extern` block uses type `u128`, which is not FFI-safe

esteve:
Fixes #320

### DIFF
--- a/rclrs/src/rcl_bindings.rs
+++ b/rclrs/src/rcl_bindings.rs
@@ -3,6 +3,7 @@
 #![allow(non_upper_case_globals)]
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
+#![allow(improper_ctypes)]
 #![allow(clippy::all)]
 #![allow(missing_docs)]
 
--- a/rclrs/src/rcl_bindings.rs
+++ b/rclrs/src/rcl_bindings.rs
@@ -3,12 +3,10 @@
 #![allow(non_upper_case_globals)]
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
-
 // Added to avoid clippy complaining about u128 types not being FFI safe
 // Caused by https://github.com/ros2/ros2/issues/1374 in iron and newer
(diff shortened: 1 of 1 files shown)


rclrs/src/rcl_bindings.rs
```diff
@@ -3,6 +3,7 @@
 #![allow(non_upper_case_globals)]
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
+#![allow(improper_ctypes)]
```
maspe36, line 6: I think it might be a good idea to separate this not allow from the rest of the blocks and add a comment referencing that its coming from ROS 2's new runtime reflection feature in Iron and Rolling. I can see this list getting quite long and goofy over time

maspe36 (changes requested):

rclrs/src/rcl_bindings.rs
esteve: Done in 80bb0acdad8aec8fae60130e5ac62765c69fbc19

esteve: I had to fix the formatting because `rustfmt` was complaining.

maspe36 (approved):

esteve:
@maspe36 thank you so much for the review :slightly_smiling_face:

PR #323, opened 2023-07-23
Developer workflow improvements to build scripts

maspe36:
- Upgraded libloading to 0.8
- Added instructions to rerun build scripts if dependent files have changed
- Enabled bindgen to invalidate the built crate whenever any transitively included header files from the wrapper change
- Removed some default true options from our bindgen builder

This enables smoother development workflows when doing things like editing our build scripts, or swapping the source'd version of ROS without requiring a new workspace or having to manually go through and removing the `target/` folder for each crate.

### DIFF
--- a/rclrs/build.rs
+++ b/rclrs/build.rs
@@ -4,6 +4,7 @@ use std::path::{Path, PathBuf};
 
 const AMENT_PREFIX_PATH: &str = "AMENT_PREFIX_PATH";
 const ROS_DISTRO: &str = "ROS_DISTRO";
+const BINDGEN_WRAPPER: &str = "src/rcl_wrapper.h";
 
 fn get_env_var_or_abort(env_var: &'static str) -> String {
     if let Ok(value) = env::var(env_var) {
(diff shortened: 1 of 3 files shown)


rclrs/build.rs
```diff
@@ -21,9 +22,8 @@ fn main() {
     println!("cargo:rustc-cfg=ros_distro=\"{ros_distro}\"");
 
     let mut builder = bindgen::Builder::default()
-        .header("src/rcl_wrapper.h")
+        .header(BINDGEN_WRAPPER)
         .derive_copy(false)
-        .allowlist_recursively(true)
```
maspe36, line 26: This is true by default ([source](https://docs.rs/bindgen/latest/src/bindgen/options/mod.rs.html#1372))

rclrs/build.rs
```diff
@@ -37,10 +37,14 @@ fn main() {
         .allowlist_var("rcutils_.*")
         .allowlist_var("rosidl_.*")
         .layout_tests(false)
-        .size_t_is_usize(true)
```
maspe36, line 40: This is true by default as well ([source](https://docs.rs/bindgen/latest/src/bindgen/options/mod.rs.html#1588))

esteve (approved):

esteve:
@maspe36 thanks. I tend to like to make options explicit (e.g. setting `size_t_is_usize` to `true` even if it's the default), but I don't it'll change in the future, so it's ok.

maspe36:
I appreciate wanting to be explicit about what we use, but there are a number of bindgen options we depend on, that we don't explicitly list here. I think this incomplete list actually adds noise which makes it harder to grok. 

If you want, I'm happy to do another PR where we explicitly list all of the bindgen options that default to true?

esteve:
> If you want, I'm happy to do another PR where we explicitly list all of the bindgen options that default to true?

@maspe36 not needed, I agree with you that it's inconsistent to only add some options but not others, it looks cleaner after this PR, thanks.

Issue #121, opened 2022-04-22
Add support for clocks


nnmm:
See https://github.com/ros2/rcl/blob/master/rcl/include/rcl/time.h

Since this is a complex feature, it would be good to have some prior experience with the code base and/or seek early-stage feedback from other contributors.

DS3a:
Quick clarification... I need to create a `time.rs` which has the same functions that are in `time.c`, using the bindings from `rcl_bindings_generated.rs` right?

DS3a:
Can someone let me know if I'm going in the right direction? so that I can make changes in case I'm wrong?

nnmm:
@DS3a appreciate the question. So, you should use the functions from time.c, via our generated bindings, but rclrs should not just wrap the functions 1:1. That would not be an optimal API. For instance, you should make use of the Rust standard library facilities (like the `std::time::Duration` type). I strongly recommend to familiarize yourself with how the time functionality from `rcl` is used in `rclcpp` and/or `rclpy` first. That should hopefully give you a more concrete idea.

I'd also recommend to tackle this issue in smaller steps, not port the full functionality at once.

Hope this helps. Let us know if you have further questions!

DS3a:
> @DS3a to understand better how time is handled in ROS2, you can also have a look at https://design.ros2.org/articles/clock_and_time.html
> 
> Regarding types, using `std::time::Duration` for example, might be possible to use, but I encourage you to use the native ROS2 types (e.g. `rclrs::Duration`, wrapping `rcl_duration_t`) as it'll provide better integration with the rest of ROS2. There may be conversion utilities between ROS2 and Rust types to make their usage easier, though. Have a look at the APIs of both `rclcpp` and `rclpy` to get inspiration.

Got it... I did start looking into it to get ideas as to how to implement it in rust... I've started building `duration.rs` which is a dependancy for `time.rs`. and both of them are a dependancy for `clock`.

[...]

PR #325, opened 2023-07-31
Add time source and clock API to nodes

luca-della-vedova:
Progress towards #121 
Hi! I'm opening this very much WIP PR to get some feedback on general direction and decisions, as well as getting some input on how to deal with some of the hurdles I've encountered.

This PR adds three modules:

### time.rs

Includes a `Time` structure, currently quite bare and can just be constructed / converted. I intentionally left most of the `Duration`, as well as different operations with times which could lead to overflows out of the PR to keep it simple.

### clock.rs

Adds a `ClockType` enum that wraps the `rcl_clock_type_t` enum but removes the uninitialized case, making all the variants valid. It also adds a `Clock` type that is meant to be the main source of truth for clock information.
The clock mimics how `rclcpp::Clock` works by using the same `rcl` function calls. It can be set to either of the variants and it offers APIs to fetch its type, as well as getting the current time, enabling / disabling ros time override, and setting the current ros time (which usually comes from a `/clock` topic).

### time_source.rs

Similar to how `rclcpp::TimeSource` / `rclpy::TimeSource` works, it keeps a list of associated clocks and reads the `use_sim_time` parameter and drive the attached clocks from a `/clock` callback if the parameter is set to true. However, parameter callbacks are not implemented so this can only be set once at node execution time.

### Test it!

If you run the minimal pub/sub example and a gazebo instance you can run the subscriber:

```
ros2 run examples_rclrs_minimal_pub_sub minimal_subscriber --ros-args --param use_sim_time:=value
```

And you should see, depending on the parameter, the wall time or gazebo time being printed.

Docs are still a WIP and that will be my next step but I would be happy to have some input on some of the main pain points I found (as well as get any feedback on the high level direction).

### Pain points

* Parameter callbacks are missing but we know this already. Is it OK if the implementation does not actually subscribe to live parameter updates but only initializes clock once on node startup time? Changing clock type at runtime can be quite tricky and is where most edge cases could come up (i.e. comparing clocks of different types) so my instinct is that this is OK for now.
* I had to introduce a dependency on `rosgraph_msgs` since this is the type of the `/clock` topic. I believe this should be OK since `rclcpp` also [depends on it](https://github.com/ros2/rclcpp/blob/22a954e1b00ea2ba810666baa92a3c66fe1a8654/rclcpp/CMakeLists.txt#L17). However, I couldn't extract the contained clock the way `rclcpp` does it because we vendor `builtin_interfaces` and the type for the clock in the vendored package and the one included when adding a `rosgraph_msgs` dependency are different. I guess the solution to this would be vendoring `rosgraph_msgs`? Or not vendoring anything at all?
* Node creation. `TimeSource`s need a reference to an existing node so they can create subscriptions for the `/clock` topic. This means that the time source can only be attached to a node after it is created. Furthermore, we have a bunch of different ways to create nodes, the main node builder creates a `Node`, utility functions create either a `Node` or an `Arc<Node>`, so the attachment is a bit hacky. If the main node builder returned an `Arc<Node>` the code would be a fair bit cleaner since we could do all the creation and attachment at the builder level but not sure this would be a welcome change.
* I couldn't figure out how to spin a separate thread to run the `/clock` subscription so for now the callbacks and spinning happens in the main thread which is quite not ideal (with multiple callbacks queued when spinning, the clock would freeze until its turn to run comes). However I couldn't find an immediate way to spawn a second thread to spin only a set of topics since there seems to be no concept of callback groups in `rclrs`.
* Comparison between different clock types. @mxgrey recommended having stronger constraints for comparing `Time` instances compared to other client libraries. The current implementation has each `Time` contain a weak pointer to the clock that generated it and have an API that allows comparison and returns an `Option` that is `None` if the two times are not comparable because they were generated from different clocks, _even if the clocks are of the same type_. This is a stronger constraint than other libraries but should also be safer.
* Another idea that would tie in nicely with the above comparison behavior but would be another deviation from other client libraries, would be to change the behavior of `RosTime` slightly. In other rcl libraries it means "Transparently use either a simulated clock or the system time depending on the `use_sim_time` parameter value."
This behavior however, creates issues when the parameter value is updated at runtime because users might suddenly try to compare a system time to a simulated time and have their code behave unexpectedly since the library would allow that.
An idea that floated is to have `RosTime` mean exclusively "simulation time", and the inner clock type changing between `RosTime` and `SystemTime` depending on the `use_sim_time` parameter. Now whenever the parameter value changes, the underlying node clock is recreated with a new variant, and all existing `Time` objects will explicitly fail to be compared, giving the user a strong warning that they might be falling into undefined behavior.
This is quite a change from other rcl libraries so I thought I'd check what the feeling is before going into it.

A last pain point I found is dealing with `TimeSource` and different `Clock` types: Other libraries for example `rclcpp` [here](https://github.com/ros2/rclcpp/blob/22a954e1b00ea2ba810666baa92a3c66fe1a8654/rclcpp/src/rclcpp/time_source.cpp#L88-L101), throw exceptions if the user builds a `TimeSource` with a `Clock` that is not set to `RosTime` _and_ the `use_sim_time` parameter is set to `true`. I tentatively made it an `Error` instead but I'm also not happy about this decision. The alternatives for the case of the user manually setting the node's clock type to `SystemTime` and setting the `use_sim_time` parameter are:
- `Panic`, consistent with `rclcpp` and `rclpy`, warns the user in the strongest possible way that something might be wrong with their configuration.
- `Result`, leaves the user the responsibility of checking how each clock attaching operation went. Does not however deal with the case of attaching a non `RosTime` clock while the `use_sim_time` is set to `false`, just for the parameter to be updated later in a parameter callback. `rclcpp` warns about this with a [parameter setting result](https://github.com/ros2/rclcpp/blob/22a954e1b00ea2ba810666baa92a3c66fe1a8654/rclcpp/src/rclcpp/time_source.cpp#L455-L466)
- `Ignore`, assume that if the user sets the clock to `SystemTime` they want the system time regardless and are OK with ignoring the `use_sim_time` parameter, also a bit dangerous and a deviation from how other libraries do it.

Following feedback in https://github.com/ros2-rust/ros2_rust/pull/325#discussion_r1281616926, the parameter is now not read if the clock is not `RosTime` so the current behavior is `Ignore`.

TODOs (on top of the above):

- [x] Looking into spinning a parallel thread for `/clock` subscription, instead of using the main node's executor. I believe this is not currently possible.
- [x] Delve deep into types and overflows and make sure behavior is consistent. Since we don't do any mathematical operations with times and types in the clock message are 32 bits, while internal storage is 64 bits, we are safe.
- [x] Fully go through all the `rcl` calls and make sure we return results when needed and ignore return values when not.
- [x] Add support for jump callbacks (might be a future PR). Will move to future PR to avoid overcomplicating this.
- [x] Tests!
- [x] Docs!
- [x] Clean `cargo fmt` and `cargo clippy`, as well as sanity check against the contribution guide


rclrs/src/time_source.rs
```diff
@@ -0,0 +1,167 @@
...
+        if !matches!(clock_type, ClockType::RosTime) && *self._ros_time_active.lock().unwrap() {
+            return Err(ClockMismatchError(clock_type));
+        }
+        if let Some(last_msg) = self._last_time_msg.lock().unwrap().clone() {
+            let nanoseconds: i64 =
+                (last_msg.clock.sec as i64 * 1_000_000_000) + last_msg.clock.nanosec as i64;
+            Self::update_clock(&clock, nanoseconds);
+        }
+        // TODO(luca) this would allow duplicates to be stored in the vector but it seems other
+        // client libraries do the same, should we check and no-op if the value exists already?
```
jhdcs, line 91: Perhaps an unordered set of some sort would work? That would make it so we don't have to check.

jhdcs (commented):
Just a quick review for now, I'll have to do a more thorough one when I get a bit more time. Still, hopefully this is at least slightly useful...

rclrs/src/clock.rs
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    /// Helper function to initialize a default clock, same behavior as `rcl_init_generic_clock`.
+    /// Needed because functions that initialize a clock take as an input a mutable reference
+    /// to a clock and don't actuall return one, so we need a function to generate one. Doing this
+    /// instead of a `Default` implementation allows the function to be private and avoids
+    /// exposing a public API to create an invalid clock
+    // SAFETY: Getting a default value is always safe.
+    unsafe fn init_generic_clock() -> rcl_clock_t {
```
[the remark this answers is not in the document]
luca-della-vedova: It would, again not sure of the approach but the reason I did it this way is written in the comment above.
The main issue is that this function creates an invalid uninitialized clock and in this way it will be impossible for users to call it since it's a private member function. Right now users can only create a clock through `new` that is guaranteed to create a valid one, but having a default trait would add a `Clock::default()` function that creates an invalid clock and I believe would be a bit of a footgun.

rclrs/src/time_source.rs
luca-della-vedova: Yea agreed, the hash set storage was one of my proposals in the [upstream issue](https://github.com/ros2/rclcpp/issues/2253), will check how that discussion goes and open matching PRs among the different client libraries to keep implementations consistent.

rclrs/src/clock.rs
```diff
@@ -0,0 +1,120 @@
...
+                rcl_disable_ros_time_override(&mut *clock);
+            }
+        }
+    }
+
+    pub fn now(&self) -> Time {
+        let mut clock = self._rcl_clock.lock().unwrap();
+        let mut time_point: i64 = 0;
+        unsafe {
+            // SAFETY: The function will only fail if the clock is not initialized
```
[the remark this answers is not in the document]
luca-della-vedova: Fair enough. The clock cannot be uninitialized at this point with the "private default constructor" trick above, otherwise it could be if we implement a default constructor, so this function should never fail.
Still, even having an uninitialized clock would not be undefined behavior, just the `rcl` function doing nothing and the above zero initialization being returned as the result. Changed the note https://github.com/ros2-rust/ros2_rust/pull/325/commits/978fd2eeef990978a08f9f47b8231ffb66f96f19

rclrs/src/node.rs
```diff
@@ -71,6 +71,9 @@ pub struct Node {
     pub(crate) guard_conditions_mtx: Mutex<Vec<Weak<GuardCondition>>>,
     pub(crate) services_mtx: Mutex<Vec<Weak<dyn ServiceBase>>>,
     pub(crate) subscriptions_mtx: Mutex<Vec<Weak<dyn SubscriptionBase>>>,
+    _clock: Arc<Mutex<Clock>>,
+    // TODO(luca) set to private
+    pub _time_source: Arc<Mutex<Option<TimeSource>>>,
```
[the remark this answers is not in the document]
luca-della-vedova: I believe this is a symptom of one of the pain points I mentioned about how can we create a `TimeSource` in the node builder if it needs to have an `Arc<Node>` as an input parameter. If we solved the issue there this could be made private and the TODO fully removed

rclrs/src/node.rs
```diff
@@ -71,6 +71,9 @@ pub struct Node {
     pub(crate) guard_conditions_mtx: Mutex<Vec<Weak<GuardCondition>>>,
     pub(crate) services_mtx: Mutex<Vec<Weak<dyn ServiceBase>>>,
     pub(crate) subscriptions_mtx: Mutex<Vec<Weak<dyn SubscriptionBase>>>,
+    _clock: Arc<Mutex<Clock>>,
```
mxgrey, line 74: Since the inner `_rcl_clock` is an `Arc<Mutex<rcl_clock_t>>` is it really necessary for this `_clock` to also be wrapped in an `Arc<Mutex<Clock>>`? Its only other field is a `ClockType` which is a simple enum so it's definitely `send` and `sync`.

I would recommend making this field a simple `Clock` and changing `get_clock`  to return `&Clock`.

rclrs/src/clock.rs
```diff
@@ -0,0 +1,170 @@
...
+            ClockType::SystemTime => rcl_clock_type_t::RCL_SYSTEM_TIME,
+            ClockType::SteadyTime => rcl_clock_type_t::RCL_STEADY_TIME,
+        }
+    }
+}
+
+/// Struct that implements a Clock and wraps `rcl_clock_t`.
+pub struct Clock {
+    _type: ClockType,
+    _rcl_clock: Arc<Mutex<rcl_clock_t>>,
```
mxgrey, line 31: I don't believe we ever need to change the value of the inner `rcl_clock_t`. I would recommend removing the `Mutex<>` wrapper and see if this code still compiles. If it does then we don't need the `Mutex<>`, and we'll avoid a lot of unnecessary locking.

luca-della-vedova: I looked at it and I couldn't come up with a way to remove the `Mutex` since the value in `_rcl_clock` could be read and written from different threads at the same time. Still, the `Arc` was unnecessary since an `rcl_clock_t` instance is always owned by a single clock so at least I could remove that layer of indirection https://github.com/ros2-rust/ros2_rust/pull/325/commits/19a73c3eb0a33e562d55f43024f3120964f91fbd.

Specifically, the `set_ros_time` [function](https://github.com/luca-della-vedova/ros2_rust/blob/19a73c3eb0a33e562d55f43024f3120964f91fbd/rclrs/src/clock.rs#L59-L74) would modify the inner `_rcl_clock` by setting its internal time, at the same time the `now` [function](https://github.com/luca-della-vedova/ros2_rust/blob/19a73c3eb0a33e562d55f43024f3120964f91fbd/rclrs/src/clock.rs#L77-L88) would read the internal clock and return it. I'm a bit afraid of what would happen if they were both called from concurrent threads, `rcl` documentation specifies that even though the functions use atomics, concurrent calls on the same `clock` object [are not safe](https://github.com/ros2/rcl/blob/960bb4ad353a43abfde5ecb831bd87823454bfc0/rcl/include/rcl/time.h#L645).

rclrs/src/node.rs
luca-della-vedova: I am quite new at `Arc`, `Mutex` and Rust multithreading so I'll admit not being 100% sure of what I am doing.

The same `Clock` instance has to be shared between the node (so users can do a `node.get_clock().now()`) and the `TimeSource` (so it can be driven by a `/clock` topic) so shared ownership is a necessity and `Arc` makes it a lot easier than having to deal with lifetimes.

However, what you say is true and since the inner `_rcl_clock` is wrapped and thread safe, there is technically no need to have this as an `Arc<Mutex<Clock>>`.
In https://github.com/ros2-rust/ros2_rust/pull/325/commits/fde0f35a85d2315b87323cec127d2b1c2c9c41de I changed it from `Arc<Mutex<Clock>>` to `Arc<Clock>`, removing the cost of double locking and making the user face API a lot cleaner. I'm not 100% sure how to remove the need for `Arc` because of the sharing I mentioned above but hopefully this is already a great win, thanks for catching it!
In the process of removing mutexes I also realized that we can just use atomic booleans instead of having a `Mutex<bool>` and addressed that as well https://github.com/ros2-rust/ros2_rust/pull/325/commits/b808cc4bdd06ef6a76308a3aa07e996bbfd71847, it was a fun investigation

rclrs/src/time_source.rs
luca-della-vedova: Upstream consensus is also to move to a `HashSet` based approach. I tentatively did a vector based avoid duplication in https://github.com/ros2-rust/ros2_rust/pull/325/commits/1f5ea5cae786017d8b05b99a1741010a4253e9c8, however I think it would be tricky to have a `HashSet`  since `Arc` requires the inner type to be hashable and `rcl_clock_t` does not implement `Hash` or `PartialEq`. Hence we would have to write a custom Hash / Eq implementation, I would guess using `Arc::as_ptr` to hash based on the memory address.

rclrs/src/node.rs
luca-della-vedova: Done in https://github.com/ros2-rust/ros2_rust/pull/325/commits/5df379fd4c4c162664e2cd8a6ff09e0056f1f1b4 (and cleaned up a bit in following commits).
I tried to use the `RwLock` but encountered poor results. In principle it should help with a many-readers pattern but in practice even the `now` function requires mutable access because the [C binding uses a mutable raw pointer](https://github.com/luca-della-vedova/ros2_rust/blob/524475d5867586f0b8d7c22f677665c0c30cbbb7/rclrs/src/clock.rs#L100) so I couldn't get it to compile.

I left a `new` function to avoid overcomplicating the Node's builder [code](https://github.com/luca-della-vedova/ros2_rust/blob/524475d5867586f0b8d7c22f677665c0c30cbbb7/rclrs/src/node/builder.rs#L274), but can take it out if you think it clutters the API too much.

The main difference with the suggested implementation is in the fact that `ClockSource` has both a function to set the time and one to enable / disable the time override. I believe the only ways to get rid of the enable / disable logic would be to encode it in the `ClockSource` existence (i.e. automatically enable the override when a `ClockSource` is constructed, disable when it is destructed) or diverge a bit from the other client library and have `RosTime` only encode "Time as dictated by the `/clock` topic", and have it fallback to `SystemTime` if `use_sim_time=false`.

jhdcs (dismissed):
This looks good to me! Though a second pair of eyes reviewing is always appreciated!

examples/minimal_pub_sub/src/minimal_subscriber.rs
```diff
@@ -16,6 +17,8 @@ fn main() -> Result<(), Error> {
             num_messages += 1;
             println!("I heard: '{}'", msg.data);
             println!("(Got {} messages so far)", num_messages);
+            let now = inner_node.get_clock().now();
+            dbg!(now);
```
luca-della-vedova, line 21: Note to self (and reviewers), revert this before merging

luca-della-vedova:
With https://github.com/ros2-rust/ros2_rust/pull/325/commits/4645935757fb87c82e6254c7dc146ddb74b86da1 I tried to change the behavior so that the requested clock type is saved in the `TimeSource`. If it is set to `RosTime` the clock will be recreated with either `RosTime` or `SystemTime` whenever the value of `use_sim_time` changes, making every new time value explicitly not comparable with previous ones.

This required a refactoring that was larger than I expected, I have to clean up a few loose ends but the major points:
* Node `get_clock` went from returning a `Clock` to an `Arc<Mutex<Clock>>`. This is needed because now the underlying clock can be destroyed and recreated if a parameter callback was to toggle the value of `use_sim_time`, so users need to lock it to make sure the value is not changed in another thread. Again there are no parameter callbacks but the API will need to be this way to support them.
* `TimeSource`  now also contains the node's `Clock`, together with its `Option<ClockSource>`. This again was needed because now `TimeSource` can change the `Clock` instance itself so it needs to have write access to it.
* Simplification: I removed the API to attach arbitrary number of clocks to a `TimeSource`. To be honest I didn't see any use case in the other rcl libraries for attaching multiple clocks to the same `TimeSource` and it introduces a fair bit of complexity, so got rid of it.

I feel some of my implementation is not so optimal and can be cleaned up but the bulk of it is there.

mxgrey:
I don't agree with returning an `Arc<Mutex<Clock>>` to the user. I think that creates unnecessary API noise that will confuse and bother end users.

In my mind an ideal implementation would be:
* The clock is stored inside of `TimeSource` as a regular `clock: Clock` field
* We provide a `fn clock(&self) -> &Clock` method on `Node` for users to borrow the current clock. They can choose to clone while borrowing if they want to store the clock for some reason.
* If the `use_sim_time` parameter changes, we use a mutable borrow of `Node` to change the `clock: Clock` field inside of `TimeSource` directly, putting a new clock instance inside.
* If users want to always access the latest clock they can just call `fn clock(&self) -> &Clock` on the `Node`.

esteve:
@luca-della-vedova excellent work! Thank you so much for working on this. I'll have more time to do a proper review by the end of the week, though.

mxgrey:
After giving it some more thought, it might not be straightforward to use shared borrows and mutable borrows when the `Node` instance needs be spread across multiple threads (unless we always move `Node`s around as `Arc<Mutex<Node>>` or `Arc<RwLock<Node>>`, but that has its own baggage). That likely means that it won't be feasible to return a `&Clock` to users from a `Node` method.

Since the likelihood of the clock instance getting changed at any given moment is close to zero, I would recommend storing it in `TimeSource` as `clock: RwLock<Clock>` and then return it from `Node` as simply a clock value by cloning: `fn clock(&self) -> Clock`. Practically all accesses of `Clock` will be read-only so they can all use `RwLock::read()` and not get in each other's way.

luca-della-vedova:
> After giving it some more thought, it might not be straightforward to use shared borrows and mutable borrows when the `Node` instance needs be spread across multiple threads (unless we always move `Node`s around as `Arc<Mutex<Node>>` or `Arc<RwLock<Node>>`, but that has its own baggage). That likely means that it won't be feasible to return a `&Clock` to users from a `Node` method.
> 
> Since the likelihood of the clock instance getting changed at any given moment is close to zero, I would recommend storing it in `TimeSource` as `clock: RwLock<Clock>` and then return it from `Node` as simply a clock value by cloning: `fn clock(&self) -> Clock`. Practically all accesses of `Clock` will be read-only so they can all use `RwLock::read()` and not get in each other's way.

Yap! This sounds very close to my thought process during the implementation. The reason my first stab at the problem used `Arc<Mutex<Clock>>` instead of a `clone()` was keeping correctness in the chance the clock changes at runtime, since there is a potential invalid scenario happening. Now I agree the chance of this happening is close to zero but to be fair the whole "destroy clock, recreate clock" logic in the previous commit was written specifically for this case, since it will only be triggered in the case the parameter changes at runtime.

The scenario that could cause unexpected behavior is:

* User calls `node.clock()`, gets a cloned `Clock` and stores it somewhere.
* In the meantime, the `use_sim_time` parameter is changed in a parameter callback, the underlying `Clock` object changes.
* Since user has a cloned clock this change is not captured and from now on all clock related functions will fail. User has to detect this and do a new `node.clock()`.

On the other hand if it was an `Arc<Mutex<Clock>>` (or better as you mentioned an `Arc<RwLock<Clock>>`):

* User calls `node.clock()`, gets an `Arc<Mutex<Clock>>` and stores it somewhere.
* `use_sim_time` parameter is changed and the underlying clock object is changed.
* Whenever user needs to actually access the clock they need to `lock()` the Mutex, and since it points to the same object their clock will always be up to date and valid.

In general I do agree that the potential unexpected scenario is very uncommon, since in my (limited) experience, pseudo rclcpp:

``` 
auto time = node->clock()->now();
```

is a lot more common than

```
auto clock = node->clock();
[...]
auto time = clock->now();
```

So I guess the question is whether we want a `clone` to happen behind the scenes or force users to lock a Mutex, or find a third way (that I'll spend today investigating)

mxgrey:
I vote strongly on the `clone()` method for several reasons:
* I don't think `Mutex` is even the right choice for managing access to the `Clock`, I believe that should be done with `RwLock` for reasons I mentioned earlier. Anything we do that exposes a specific choice of access manager means we will break the API if we ever tweak the implementation, e.g. to improve performance.
* Forcing users to access clocks through a mutex creates an unnecessary cognitive burden on them, e.g. "Why do I need to lock something just to check the time? What should I do if I get an Err?".
* I think it makes **more** sense that if a user **chooses** to store the clock somewhere, that clock will forever continue to produce self-consistent time values. A user that wants to account for the possibility that the underlying type of clock will change **should** ask for a new clock from the node each time.
* Based on patterns and conventions seen in other client libraries, the vast majority of users will not store the clock and will fetch it from the node each time anyway, giving the expected default behavior.

mxgrey:
One high level remark about the current state of the PR is it uses `.unwrap()` an awful lot. I haven't looked enough through the overall ros2-rust codebase to know how prevalent `.unwrap()` already is, but I think the general attitude in the Rust community is that libraries which use `.unwrap()` (without carefully documenting and ensuring satisfaction of invariants) are not to be considered "production ready".

It's probably okay to merge this PR as-is if we intend to do another pass through the whole `ros2-rust` codebase eliminating uses of `.unwrap()`, but we probably shouldn't consider this to be production-ready code until that's been done.

luca-della-vedova:
> One high level remark about the current state of the PR is it uses `.unwrap()` an awful lot. I haven't looked enough through the overall ros2-rust codebase to know how prevalent `.unwrap()` already is, but I think the general attitude in the Rust community is that libraries which use `.unwrap()` (without carefully documenting and ensuring satisfaction of invariants) are not to be considered "production ready".
> 
> It's probably okay to merge this PR as-is if we intend to do another pass through the whole `ros2-rust` codebase eliminating uses of `.unwrap()`, but we probably shouldn't consider this to be production-ready code until that's been done.

Agreed. The bulk of the unwraps here come from two sources:

Mutex / RwLock lock unwrapping, it seems to be done extensively in this codebase and I didn't find any case of dealing with lock poisoning, so I just followed the way it seems to be done.

[Clock creation returning a Result](https://github.com/ros2-rust/ros2_rust/blob/632e082f272fd27a853409599ca134058e82df2d/rclrs/src/clock.rs#L69-L81). This `Result` is pretty much unwrapped everywhere. If we were to "just" unwrap in the constructor we could clean up a fair bit of code. I left the `Result` to keep the option open but I have a mild recommendation to just unwrap at the constructor level. According to the `rcl_clock_init` [function](https://github.com/ros2/rcl/blob/feb775a49a3c32eb2c619baaff5651c0aa2d099a/rcl/src/rcl/time.c#L97-L117), it will fail in the following cases:

* A null pointer is passed as an input. Technically possible but we would have to go out of our way to manually initialize a null pointer and send it to the unsafe function.
* An invalid enum value is passed as a `ClockType`. Impossible since clock type is an enum and we can't pass an arbitrary value (unlike C enums that are pretty much integers).
* The clock is `RosTime` _and_ a heap memory allocation for the clock storage (an int64 and a bool) fails. Here I would argue that if the OS doesn't have 9 bytes we have much larger issues to worry about.

rclrs/src/node.rs
```diff
@@ -355,6 +361,12 @@ impl Node {
         domain_id
     }
 
+    /// Gets a parameter given the name.
+    /// Returns None if no parameter with the requested name was found.
+    pub fn get_parameter(&self, name: &str) -> Option<ParameterValue> {
+        self._parameter_map.get(name).cloned()
+    }
+
```
luca-della-vedova, line 369: This is also a more controversial change.
Creating this API without adding any of the other `Parameter` APIs (specifically `declare_parameter`) would effectively mean that the behavior is what we would have with `automatically_declare_parameters_from_overrides` set to `true`, which is not the case in all other client libraries that default to the parameter set to `false`.
A possibility could be to remove this API until we have at least a `declare_parameter`, which would however make the whole `RosTime` section of the PR dead code.

esteve (changes requested):

esteve:
@luca-della-vedova

Thanks again for all the hard work, some replies to the pain points part:

> I couldn't figure out how to spin a separate thread to run the /clock subscription so for now the callbacks and spinning happens in the main thread which is quite not ideal

I'd prefer if the `SingleThreadedExecutor` (the only one we have now) didn't spawn new threads, can't remember well, but does the `rclcpp` version use threads for clocks?

> If the main node builder returned an Arc<Node> the code would be a fair bit cleaner since we could do all the creation and attachment at the builder level but not sure this would be a welcome change.

We tend to leave the `Arc<...>` variants of creating a `struct` for the `create_XXX` functions, but if it makes more sense for the builder to retunr an `Arc<...>` we can revisit that decision.

> I guess the solution to this would be vendoring rosgraph_msgs? Or not vendoring anything at all?

We still to vendorize messages because some of the `rclrs` API (that's why vendorize `builtin_interfaces`), so I would be ok vendorizing `rosgraph_mgs` if you need it.

luca-della-vedova:
> > I couldn't figure out how to spin a separate thread to run the /clock subscription so for now the callbacks and spinning happens in the main thread which is quite not ideal
> 
> I'd prefer if the `SingleThreadedExecutor` (the only one we have now) didn't spawn new threads, can't remember well, but does the `rclcpp` version use threads for clocks?

Yes there is an option in `rclcpp` to spin a separate thread for clocks and it [defaults to true](https://github.com/ros2/rclcpp/blob/ea31f94eb453acf3f34b37b85e38284abc514d57/rclcpp/include/rclcpp/time_source.hpp#L65).
This is especially important in nodes that don't spin. There is a reason I put the "debug print now" in the `minimal_subscriber` example, if I were to put it in the `minimal_publisher` that node never spins, hence never processes callbacks / updates its time.
Now nodes that don't spin are probably quite rare outside of examples but this is still a limitation.
The other issue is of course is that time will be effectively frozen until the `/clock` callback is processed, which might be more severe in large and complex callbacks.

> 
> > If the main node builder returned an Arc the code would be a fair bit cleaner since we could do all the creation and attachment at the builder level but not sure this would be a welcome change.
> 
> We tend to leave the `Arc<...>` variants of creating a `struct` for the `create_XXX` functions, but if it makes more sense for the builder to retunr an `Arc<...>` we can revisit that decision.

In the end I went with changing the node builder to create an `Arc`, the time source needs a weak reference to a node to be able to manage the clock subscriptions. This can't be done in the builder unless we build an `Arc`.

> 
> > I guess the solution to this would be vendoring rosgraph_msgs? Or not vendoring anything at all?
> 
> We still to vendorize messages because some of the `rclrs` API (that's why vendorize `builtin_interfaces`), so I would be ok vendorizing `rosgraph_mgs` if you need it.

I'm not sure what the procedure here is. The code "seems to compile" but it is downloading `rosgraph_msgs` behind the scenes and I don't know what the side effects of that are.

rclrs/src/node.rs
mxgrey: I recommend marking this as a `pub(crate)` function for now so that we can use it internally for `use_sim_time` but aren't committing it to the public API.

I expect our next PR will introduce parameters, so this wouldn't be in the public API for long, but it's probably better for it to not be in at all to avoid confusing anyone.

rclrs/src/clock.rs
```diff
@@ -0,0 +1,221 @@
...
+    pub fn clock_type(&self) -> ClockType {
+        self._type
+    }
+
+    /// Returns the current clock's timestamp.
+    pub fn now(&self) -> Time {
+        let mut clock = self._rcl_clock.lock().unwrap();
+        let mut time_point: i64 = 0;
+        unsafe {
+            // SAFETY: No preconditions for his function
```
maspe36, line 93: nit: `his` -> `this`

rclrs/src/clock.rs
```diff
@@ -0,0 +1,221 @@
...
+            rcl_clock_get_now(&mut *clock, &mut time_point);
+        }
+        Time {
+            nsec: time_point,
+            clock: Arc::downgrade(&self._rcl_clock),
+        }
+    }
+
+    /// Helper function to initialize a default clock, same behavior as `rcl_init_generic_clock`.
```
maspe36, line 102: nit: Perhaps consider rewording slightly to something like this? I had a hard time following at first

> Helper function to privately initialize a default clock. This function has the same behavior as `rcl_init_generic_clock`. However, by defining a function like this, instead of implementing `Default`, we avoid exposing a public API to create an invalid clock.

I think the main point to convey is that we'd rather define this privately than in a public API

rclrs/src/time_source.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    /// Returns the clock that this TimeSource is controlling.
+    pub fn get_clock(&self) -> Clock {
+        self._clock.read().unwrap().clone()
+    }
+
+    /// Attaches the given node to to the `TimeSource`, using its interface to read the
+    /// `use_sim_time` parameter and create the clock subscription.
+    pub fn attach_node(&mut self, node: Weak<Node>) {
+        self._node = node;
```
maspe36, line 107: Perhaps I missed something, but do we actually need to store the weak reference to the Node at all here? It seems like we only read the `use_sim_time` parameter at the time of a `TimeSource` structs creation. Unless we plan to re-use these `TimeSource`'s or if it actually does make sense to have them update when the parameter updates, I don't think we should store the weak reference.

luca-della-vedova: Indeed we only read the value of the parameter (and create the clock subscriber) once at creation time right now so we could technically just pass the node as an argument to the builder and just store an `Option<Subscription>` instead.

However, this behavior is not the full implementation and all other client libraries do store the node in the time source. The reason for this is that the time source needs to be able to subscribe to parameter changes and, when the `use_sim_time` parameter changes at runtime, be able to create and destroy subscriptions. This is not possible without holding a reference to the node.
If we just did the "bare minimum" now and didn't pass / store the `node` we would have to change the API once parameters are implemented.

rclrs/src/clock.rs
luca-della-vedova: https://github.com/ros2-rust/ros2_rust/pull/325/commits/ed583f67d44e382015766a325041b8b5389dfd38. Reworded a bit to avoid saying the word `function` three times in a row but kept most of your suggestion.

rclrs/src/clock.rs
luca-della-vedova: https://github.com/ros2-rust/ros2_rust/pull/325/commits/ed583f67d44e382015766a325041b8b5389dfd38

examples/minimal_pub_sub/src/minimal_subscriber.rs
```diff
@@ -6,6 +6,7 @@ fn main() -> Result<(), Error> {
     let context = rclrs::Context::new(env::args())?;
 
     let node = rclrs::create_node(&context, "minimal_subscriber")?;
+    let inner_node = node.clone();
```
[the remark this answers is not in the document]
luca-della-vedova: I just removed all the changes in this file since they weren't meant to land in the final code anyway https://github.com/ros2-rust/ros2_rust/pull/325/commits/ed583f67d44e382015766a325041b8b5389dfd38

rclrs/src/time_source.rs
mxgrey: I wonder how much access an end user actually needs to `TimeSource`. I can't remember ever needing to touch that as a user in any other client library.

Is it feasible to slap `pub(crate)` on enough structs and methods that we can avoid API commitments until the full implementation is ready?

luca-della-vedova: Thanks for the suggestion! Indeed you are right and that part of  the code is very rarely touched by users.
In https://github.com/ros2-rust/ros2_rust/pull/325/commits/382f780edf87a5190d1fdc3b3ea14d9bd3a25a22 I made everything in `time_source.rs` either private or `pub(crate)` so it doesn't export anything as a public API. I also removed the `use_clock_thread` builder argument since it is not implemented yet and sounds like might not be for some time.
This added another unwrap that however falls in the class of the ones I described above (only fail if we can't allocate a few bytes of memory) so it's mostly noise.

rclrs/src/time_source.rs
luca-della-vedova: Simplified it in the end to get rid of the vector of clocks and just have a single clock.
To be honest even though the other client libraries allow having multiple clocks I haven't found a single case of that being done, so imho we can get away with the additional complexity of multiple clocks and just the node's clock.

rclrs/src/time_source.rs
maspe36: > when the `use_sim_time` parameter changes at runtime

This doesn't happen currently, I'm guessing its blocked by the lack of a parameter implementation in this PR?

I would prefer if we did something similar to the parameters impl where we deviate slightly from rclcpp. Forcing end users to wrap all nodes in `Arc` is a large API change for a niche runtime configuration option IMO

But I'll defer to @esteve, @nnmm, and @jhdcs. I think this PR has been open long enough and we can always iterate :)

maspe36 (approved):

rclrs/src/time_source.rs
mxgrey: Realistically I think it'll be close to impossible to implement a whole bunch of sophisticated features down the road without requiring that `Node`s are wrapped in `Arc` unless we take an approach where almost every field inside in the `Node` struct is wrapped in its own `Arc` to hide this implementation detail.

But I don't think that's a great idea because it just hides important information from the user about the lifecycle and identity of a node. For example a user might mistakenly think that they can create a second node instance by cloning an existing one when in reality they would only be getting a new reference to the same node instance.

Putting the Arc on the outside explicitly communicates the lifecycle and shareability properties of the node.

I can imagine implementing a Rust client library that does not use Arc for nodes if it were implemented inside of an ECS framework, but I think such a library should be done as a separate project from this one.

esteve:
@luca-della-vedova now that #332 has been merged, could you rebase this PR? Thanks.

luca-della-vedova:
> @luca-della-vedova now that #332 has been merged, could you rebase this PR? Thanks.

Done and I also added a test that makes sure `use_sim_time=true` behaves as expected (there wasn't one before and it was just in the example code).
In the process I had to add a few `Mutex(_)` wrappers to `TimeSource` fields that I'm not too happy about, I'll be poking at the code to see if it can be improved.

esteve:
@jhdcs @maspe36 do you guys have more feedback? This would be a nice feature to have for 0.4

jhdcs (approved):
This looks good to me for including in a 0.4 release. Sure, there may be some rough edges, but we can smooth those out in future more-narrowly-tailored PRs.

esteve (approved):
@luca-della-vedova thank you so much for this contribution!

PR #332, opened 2023-10-05
Add preliminary support for parameters

luca-della-vedova:
## Description

This PR brings preliminary support for parameters to `rclrs`. It's fairly complex and some choices have been made for ergonomics and safety that don't necessarily match what other client libraries do. At this stage feedback on whether the behavior and those deviations are OK is probably more productive that going too deep into code review.

There are extensive unit tests implemented in parameter.rs. These can be a good example of what the experience of using the library would be like.

~~Note: This PR is currently built on top of #325 since it refactors some of its usage of parameters, depending on the preferred order we can either look at this after or I can rebase this on top of main and look at the time source PR later, but they are still somewhat interconnected.~~

### What's implemented

* Parameter getting and setting API, as well as parsing of command line overrides (this was present but not accessible).
* Both declared and non declared parameters.

### What's not implemented

* Parameter services.
* Parameter callbacks, such as on set, pre set and post set callbacks.
* Automatic parameter declaration from overrides.

### API Overview

The main API that users are expected to interact with is `node.declare_parameter(...)` which returns a parameter builder object that can be used to set its properties before finalizing it and declaring the parameter.
The lifetime of the parameter is tied to the object returned by the declaration. A parameter can only be declared once and once the returned object is destroyed it will be automatically undeclared.
The returned object depends on the type of parameter and offers different APIs.

Three types of parameters have been implemented:

1. `Mandatory`, built through `node.declare_parameter(name, default_value).[options].mandatory()`. This is a parameter that must always have its value set. Its setter must provide a value and its getter will always return a value.
2. `Optional`, built through `.optional()` builder function. Differs from the mandatory since it accepts an `Option<T>` as a default value and can be set to `None`. The getter as a consequence will return an `Option<T>` as well.
3. `ReadOnly`, built through `.read_only()` builder function. Cannot be changed and only offers a `get` API, making it actually impossible to edit.

#### Parameter typing and interaction

The main way users are expected to do operations on parameters is through the object returned by the builder and its associated getters and setters. This allows better compile time control over the type and less error handling to be done by the user, for example:

```
let int_param = node.declare_parameter("int_param", 10).mandatory();
// Getter will always return an i64
assert_eq!(int_param.get(), 10);
// Setter only accepts an i64
assert!(int_param.set(100).is_ok());
```

This is opposed to the parameter API in the other libraries that leaves the user to make sure the type returned by the getter is the expected type.

Of course there must be support for dynamic typing, but this still works with the proposed API where now users declare a parameter of enum type, like so:

```
let dynamic_param = node.declare_parameter("dynamic_int", ParameterValue::Integer(10)).mandatory();
// Setter now accepts any object of ParameterValue type
assert!(dynamic_param.set(ParameterValue::Double(1.0)).is_ok());
```

The dynamic nature of the parameter is implied from the user setting its value as an enum rather than a matching primitive type. This is in contrast to other client libraries that use a boolean flag for this purpose.

#### Parameter lifetime and declaration

As mentioned before parameters will be declared and undeclared automatically based on the lifetime of the object returned by the declaration.

#### Undeclared parameter API

Users can still access all parameters, declared or undeclared, through their name and value. However this is hidden behind a `node.use_undeclared_parameters()` that returns an interface that has the more familiar `.set(name, value)` and `.get(name)`.
The idea is that users must explicitly express their intent to allow setting undeclared parameters in their code. When services are implemented they will behave accordingly when trying to set a non existing parameter.

#### No ParameterDescriptor argument

Other libraries such as `rclcpp`, use a `ParameterDescriptor` object as an input to the declaration function. However this is non ideal in several ways:

* The message contains a type and a name. However these properties are already included in the `declare_parameter` function signature so they are unnecessary and silently ignored.
* `read_only` is a boolean parameter which is saved as a parameter property. Every time a parameter is set the code needs to check it and return an error if it is read only. This is not allowed with the proposed API that doesn't offer a `set` function in the first place.
* The descriptor contains both an `int_range` and a `float_range` field. This is unnecessary information and also a bit confusing if for example we are declaring a statically typed string parameter.

For such reason a builder type API has been implemented. The `read_only` and `dynamic` properties of the descriptor don't exist anymore and are embedded in the type as described in the previous paragraphs. The others are arguments of the builder. For example:

```
let range = ParameterRange {
    lower: Some(0),
    upper: Some(100),
    step: Some(10),
}

let int_param = node.declare_parameter("percentage", 0)
                    .range(range)
                    .description("Describes a percentage, only multiple of 10 are accepted")
                    .constraints("If the robot is of model X only values above 50 are acceptable");
```

Depending on the parameter type only a certain type of range can be set:

* `Integer` param: the `range(..)` function will take an integer range as an argument.
* `Double` param: the `range(..)` function will take a double range.
* Dynamic typing parameter, the input will be `ParameterRanges` struct that contains both a float and an integer range.
* Other types currently have an empty range function that does nothing.

#### More control over ranges

As shown partially above, ranges now are only specified when the correct parameter type applies. In addition, a few changes are present that might differ from how they work in other libraries.

There is an `InvalidRange` error that will be thrown if the user specifies an invalid range, like lower > upper or step <= 0. These are all allowed in `rclcpp` where:

* If lower > upper, the only acceptable parameter values are equal to either lower or upper.
* If step == 0 it will be interpreted as no step (we use `Option` for this).
* If step < 0 it will be still used.

In my opinion these cases, especially the first one, are really not very intuitive and for this reason the proposal is to introduce an error that flags them out.

Furthermore, in the ROS interface, if a range is specified its upper and lower bounds must both be specified, while the step is optional (where step = 0 means no step specified).
The proposal here is to also allow users to specify only one of lower or upper bound, these can be trivially translated to the interface messages by using the float infinity or integer max value.

#### Easily clonable ParameterValue

The `ParameterValue` variants that were not wrapping primitive types and needed heap allocations have been wrapped by an `Arc<[]>` type. This allows both less expensive `parameter.get()` calls, since they return a cloned value, and guarantee immutability outside of the `parameter.set(value)` functions.
Extra care has been put to make the parameter declaration behavior as ergonomic as possible and avoid forcing users to add `Arc::from(type)` or `.into()` in all their calls.

#### Extensive and partially configurable error reporting

Parameter declaration is a complex subject and each library has some discretion over what is acceptable and what is not. In this implementation the following always fail:

* Declaring a parameter with a range that is not valid (as described above).
* Declaring a parameter that was already declared.
* Declaring a parameter of a certain type but its override was of a different type.
* Declaring a parameter with a certain range but its override is out of range.
* Declaring a parameter with a range and a default value but its default value is out of range.

In addition, we use an explicitly different storage for declared and undeclared parameters, upon declaration if there was already an undeclared value the following could happen:

* Previously undeclared value is of a different type than what declared.
* Previously undeclared value is out of the declared range.

This API defaults to ignoring the value set through the undeclared interface if any of these happens. The idea is that at the declaration level the user knows exactly what their parameter should contain. Furthermore if declaration failed it could be tricky for users to reset the parameter.
Still, there is an API to make these cases an explicit error rather than a silently drop behavior, as such:

```
// Note the wrong type
node.use_undeclared_parameters().set("float_param", "foo");
// This returns a type mismatch error.
assert!(node.declare_parameter("float_param", 1.0).tentative().mandatory().is_err());
// This silently overrides the type and returns a parameter with current value = 1.0
let float_param = node.declare_parameter("float_param", 1.0).mandatory().unwrap();
```


mxgrey:
In particular it would be great if any of @nnmm @esteve @jhdcs (and anyone else) can glance over the test code to get a sense of the API that's being proposed before the working group meeting on October 9 (this Monday) so we can have a discussion then.

esteve:
@luca-della-vedova thanks for submitting this PR.

> Note: This PR is currently built on top of https://github.com/ros2-rust/ros2_rust/pull/325 since it refactors some of its usage of parameters, depending on the preferred order we can either look at this after or I can rebase this on top of main and look at the time source PR later, but they are still somewhat interconnected.

Can you elaborate how they are interconnected? A priori they don't seem to be related, would it be possible to remove the changes from #325 from this PR? As it is now, it's not the easiest PR to read. Thanks.

luca-della-vedova:
> @luca-della-vedova thanks for submitting this PR.
> 
> > Note: This PR is currently built on top of #325 since it refactors some of its usage of parameters, depending on the preferred order we can either look at this after or I can rebase this on top of main and look at the time source PR later, but they are still somewhat interconnected.
> 
> Can you elaborate how they are interconnected? A priori they don't seem to be related, would it be possible to remove the changes from #325 from this PR? As it is now, it's not the easiest PR to read. Thanks.

#325 needs to be able to access parameters since it needs to read `use_sim_time` and act on it accordingly so it had to implement a very bare parameter API to be functional.
On the other hand, this PR implements an actual set of APIs for parameters so it includes a commit to delete the temporary implementation from the `Clock` API and bring it to the full implementation.
If the changes were removed this would have to be done again whenever one of the PRs is merged. But I understand where you come from, I removed all the clock related changes in https://github.com/ros2-rust/ros2_rust/pull/332/commits/6dcc21fbcd28a9992616a353d3d9ecb7f27a8764 and will keep a mental note on how to solve the conflicts that might come up.

esteve:
@luca-della-vedova 

> https://github.com/ros2-rust/ros2_rust/pull/325 needs to be able to access parameters since it needs to read use_sim_time and act on it accordingly so it had to implement a very bare parameter API to be functional.

In that case, shouldn't it be opposite? #325 based on this PR?

> But I understand where you come from, I removed all the clock related changes in https://github.com/ros2-rust/ros2_rust/commit/6dcc21fbcd28a9992616a353d3d9ecb7f27a8764 and will keep a mental note on how to solve the conflicts that might come up.

Thanks, I really appreciate it.

luca-della-vedova:
> @luca-della-vedova
> 
> > #325 needs to be able to access parameters since it needs to read use_sim_time and act on it accordingly so it had to implement a very bare parameter API to be functional.
> 
> In that case, shouldn't it be opposite? #325 based on this PR?

In theory yes, my original thought was that since that PR went through some rounds of review and is generally simpler it would have been closer to being merged than this brand new and fresh take on a fairly complex problem that is the one of parameters. Anyway we can now keep them fully decoupled and not have to worry about it anymore.

nnmm:
Thanks for the extensive MR description :) I'm not sure when (or honestly, if) I have time to fully review in the near future. Have you seen my WIP branch for parameters? https://github.com/ros2-rust/ros2_rust/tree/parameter-service

nnmm:
One thing I did there is define a [trait for parameter validators](https://github.com/ros2-rust/ros2_rust/blob/parameter-service/rclrs/src/parameter/constraint.rs), which could be a general-purpose callback or a lower-upper range etc. Do you think that would make sense here too?

(this is not the most important thing, but you gotta start somewhere with these big PRs)

nnmm:
About undeclared parameters - iirc rclcpp doesn't represent undeclared parameters differently in its parameter map, I think it's more like: allowing undeclared parameters changes how parameters can be created, but once created, the declared parameters and the implicitly (without declaration) created parameters look the same. Could be that I'm totally wrong, but maybe you could elaborate a bit on what the functional differences of undeclared vs declared parameters in the map are.

nnmm:
Really impressive work btw!

rclrs/src/parameter.rs
```diff
@@ -3,3 +3,1285 @@ mod value;
...
+struct DeclaredStorage {
+    value: DeclaredValue,
+    kind: ParameterKind,
+    options: ParameterOptionsStorage,
+}
+
+#[derive(Debug)]
+enum ParameterStorage {
+    Declared(DeclaredStorage),
+    Undeclared(Arc<RwLock<ParameterValue>>),
```
mxgrey, line 475: In the current API, the only ways to access an undeclared parameter are
```
node.use_undeclared_parameters().get::<f64>("my_param");
node.use_undeclared_parameters().set("my_param", 1.23);
```
In both cases, the undeclared value will be locked and then cloned+returned (or set) with the parameter map locked the whole time. As far as I can figure we don't stand to gain anything from having an `Arc<RwLock<_>>` for the `Undeclared` variant.

It might make sense to have this `Arc` if we wanted to return some kind of `Undeclared` struct that holds onto a copy of the `Arc`, but I don't think that would be sound since an undeclared parameter could be changed to a declared parameter at any time, and holding an `Arc<RwLock<ParameterValue>>` of the old undeclared parameter would clash with that (e.g. how would the `Undeclared` struct know about the declared constraints?).

It might be possible to have an `Undeclared` struct if we wrap the entire `DeclaredStorage` in an `Arc<RwLock<_>>`, but until we want to provide such a struct I would recommend removing the `Arc<RwLock<_>>` for this.

mxgrey:
> maybe you could elaborate a bit on what the functional differences of undeclared vs declared parameters in the map are.

The gist is that declared parameters carry a lot more information with them. When a parameter is declared it will have some type (we say `kind` in the implementation to avoid keyword collision) and may have a range, description, or other properties. An undeclared parameter does not have any of that.

I imagine in rclcpp they probably just have "sensible default" values for all that extra information to store declared and undeclared parameters the same way. That makes sense in C++ where tagged unions are ... not pleasant (`std::variant` has terrible ergonomics). In this implementation we're taking an approach of "don't do/say more than what is explicitly needed". An undeclared parameter cannot have ranges, constraints, or descriptions because the user never provided any of those, so we should not pretend to have any of them in our implementation.

This will also give us a shortcut when we implement parameter services: parameters tagged with `Declared(_)` will have additional code to ensure its constraints are enforced. Meanwhile `Undeclared(_)` can immediately shortcut that.

luca-della-vedova:
> One thing I did there is define a [trait for parameter validators](https://github.com/ros2-rust/ros2_rust/blob/parameter-service/rclrs/src/parameter/constraint.rs), which could be a general-purpose callback or a lower-upper range etc. Do you think that would make sense here too?
> 
> (this is not the most important thing, but you gotta start somewhere with these big PRs)

I didn't know that the effort started already! Regarding the validators I fully agree we should have a general purpose validator that can be set by the user to a callback, so they can perform arbitrary checks. This will be part of the future work for validation callbacks.
However, the [ParameterDescriptor](https://github.com/ros2/rcl_interfaces/blob/rolling/rcl_interfaces/msg/ParameterDescriptor.msg#L32-L36) message that is returned from the `/describe_parameters` service has a range field so we need to be able to translate whatever constraint the user set to a range object that can be returned to the service call. From my understanding this translation is immediate if `Range` is a separate object but would be trickier if it is just an object that implements a trait and we share a bunch of `Box<dyn ParameterConstraint>` objects.

rclrs/src/node.rs
```diff
@@ -71,7 +71,7 @@ pub struct Node {
     pub(crate) guard_conditions_mtx: Mutex<Vec<Weak<GuardCondition>>>,
     pub(crate) services_mtx: Mutex<Vec<Weak<dyn ServiceBase>>>,
     pub(crate) subscriptions_mtx: Mutex<Vec<Weak<dyn SubscriptionBase>>>,
-    _parameter_map: ParameterOverrideMap,
+    _parameter: ParameterInterface,
```
esteve, line 74: I like this pattern from `rclcpp`, we should do the same for services, subscriptions, etc.

rclrs/src/parameter.rs
```diff
@@ -3,3 +3,1285 @@ mod value;
...
+        }
+        Ok(())
+    }
+}
+
+pub(crate) struct ParameterInterface {
+    _parameter_map: Arc<Mutex<ParameterMap>>,
+    _override_map: ParameterOverrideMap,
+    allow_undeclared: AtomicBool,
+    //_services: ParameterService,
```
esteve, line 632: Could you remove commented code? I think adding a comment with a `NOTE(luca-della-vedova)` with context would be useful, though.

rclrs/src/parameter.rs
```diff
@@ -3,3 +3,1285 @@ mod value;
...
+
+    pub(crate) fn declare_from_iter<U: IntoIterator>(
+        &self,
+        name: &str,
+        default_value: U,
+    ) -> ParameterBuilder<Arc<[U::Item]>>
+    where
+        Arc<[U::Item]>: ParameterVariant,
+    {
+        // TODO(luca) consider passing a FnOnce to initialize the value to declare to do lazy
```
esteve, line 722: Please change `luca` to `luca-della-vedova`, so it's easier to track. Thanks.

esteve (changes requested):

luca-della-vedova:
> @luca-della-vedova thank you much for the PR! I like the changes in the `rclcpp` API, could you add a document or maybe just an explanation as part of the `parameter.rs` file with the divergences from the `rclcpp` API?
> 
> I think this PR is a good foundation for possible refinements in follow up PRs. As mentioned in the meeting from October 9th, these changes could result in a proposal to flesh out the parameters APIs in `rclcpp` and `rclpy`.

I added a link to the PR as well as a few bullet points for the main deviations in https://github.com/ros2-rust/ros2_rust/pull/332/commits/b142ceca6b8cbba778ae1654b807e73f7f415a74. It was a bit hard to be both brief and explicative but I tried to do my best!

rclrs/src/node.rs
```diff
@@ -360,6 +360,84 @@ impl Node {
...
 
+    /// Creates a `ParameterBuilder` that can be used to set parameter declaration options and
+    /// declare a parameter as `Optional`, `Mandatory` or `ReadOnly`.
+    ///
+    /// # Example
+    /// ```
+    /// # use rclrs::{Context, ParameterRange, RclrsError};
+    /// let context = Context::new([])?;
+    /// let node = rclrs::create_node(&context, "domain_id_node")?;
+    /// // Set it to a range of 0-100, with a step of 2
```
maspe36, line 371: Isn't the range `-100 - 100`?

rclrs/src/parameter.rs
```diff
@@ -3,3 +3,1299 @@ mod value;
...
+use crate::rcl_bindings::*;
+use crate::{call_string_getter_with_handle, RclrsError};
+use std::collections::BTreeMap;
+use std::marker::PhantomData;
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc, Mutex, RwLock, Weak,
+};
+
+// This module implements the core logic of parameters in rclrs.
```
maspe36, line 16: Not something that would need to be done for this PR, but we should probably collect all of these divergent changes in a top level markdown file

rclrs/src/parameter.rs
```diff
@@ -3,3 +3,1299 @@ mod value;
...
+    fn validate(&self) -> Result<(), DeclarationError> {
+        if let Some(integer) = &self.integer {
+            integer.validate()?;
+        }
+        if let Some(float) = &self.float {
+            float.validate()?;
+        }
+        Ok(())
+    }
+    fn check_in_range(&self, value: &ParameterValue) -> Result<(), ParameterValueError> {
```
maspe36, line 118: nit: Spacing between impl functions

rclrs/src/parameter.rs
```diff
@@ -3,3 +3,1299 @@ mod value;
...
+        // Declaring a parameter with a different type than what was overridden should return an
+        // error
+        assert!(matches!(
+            node.declare_parameter("declared_int", 1.0).mandatory(),
+            Err(DeclarationError::Override(
+                ParameterValueError::TypeMismatch
+            ))
+        ));
+
+        // If the override does not respect the range we should return an error
```
maspe36, line 846: nit: `range,`

rclrs/src/parameter.rs
```diff
@@ -3,3 +3,1299 @@ mod value;
...
+        assert_eq!(optional_param2.get(), Some(false));
+
+        // This was provided as a parameter override, hence should be set to true
+        let optional_param3 = node
+            .declare_parameter("optional_bool", Some(false))
+            .optional()
+            .unwrap();
+        assert_eq!(optional_param3.get(), Some(true));
+
+        // double_array was overriden to 1.0 2.0 through command line overrides
```
maspe36, line 980: nit: `[1.0, 2.0]`

rclrs/src/parameter.rs
```diff
@@ -3,3 +3,1299 @@ mod value;
...
+        assert_eq!(array_param.get()[1], 2.0);
+
+        let array_param = node
+            .declare_string_array_parameter("string_array", vec!["Hello", "World"])
+            .mandatory()
+            .unwrap();
+        assert_eq!(array_param.get()[0], "Hello".into());
+        assert_eq!(array_param.get()[1], "World".into());
+
+        // If a value is set when undeclared, a following declare_parameter should have the
```
maspe36, line 995: nit: `a following` -> `the following`

rclrs/src/parameter.rs
```diff
@@ -3,3 +3,1299 @@ mod value;
...
+            upper: Some(10),
+            step: Some(-1),
+        };
+        assert!(matches!(
+            node.declare_parameter("int_param", 5)
+                .range(range)
+                .mandatory(),
+            Err(DeclarationError::InvalidRange)
+        ));
+        // Setting parameters out of range range should fail
```
maspe36, line 1103: nit: Duplicate range?

rclrs/src/parameter.rs
```diff
@@ -3,3 +3,1299 @@ mod value;
...
+    /// * Err(DeclarationError::TypeMismatch) if the type of the requested value is different from
+    /// the parameter's type.
+    pub fn set<T: ParameterVariant>(
+        &self,
+        name: &str,
+        value: T,
+    ) -> Result<(), ParameterValueError> {
+        let mut map = self.interface._parameter_map.lock().unwrap();
+        if let Some(entry) = map.storage.get_mut(name) {
+            // If it's declared we can only set if it's the same variant.
```
maspe36, line 615: nit: `declared,`

maspe36 (approved):
I'm not the most familiar person with ROS 2 parameters, but I really like the ergonomics of this impl. 

Thanks for the great work!

nnmm:
> However, the [ParameterDescriptor](https://github.com/ros2/rcl_interfaces/blob/rolling/rcl_interfaces/msg/ParameterDescriptor.msg#L32-L36) message that is returned from the `/describe_parameters` service has a range field so we need to be able to translate whatever constraint the user set to a range object that can be returned to the service call. From my understanding this translation is immediate if `Range` is a separate object but would be trickier if it is just an object that implements a trait and we share a bunch of `Box<dyn ParameterConstraint>` objects.

My understanding is that in `rclcpp`,  the parameter callbacks are validators, and they don't fit the `ParameterDescriptor` message either, so it would be fine to have more general validation. My impression (without researching it really) is that this field is for informational purposes, for humans.

rclrs/src/parameter.rs
nnmm: This is a really different usage of the words "declared" and "undeclared" than `rclcpp` uses imo. In `rclcpp`, there is no internal distinction between a declared and an undeclared parameter - some parameter either exists or it doesn't. The `allow_undeclared_parameters` flag just causes all parameter overrides to create/declare parameters automatically, instead of acting only as defaults for `declare_parameter()`. By contrast, you have a user-facing distinction between the two and do not implement `allow_undeclared_parameters`, right?

The PR in rclcpp that introduced the "declaration" concept was https://github.com/ros2/rclcpp/pull/495.

To reply to @mxgrey's [comment](https://github.com/ros2-rust/ros2_rust/pull/332#issuecomment-1751925649) on the rationale for the distinction:

> The gist is that declared parameters carry a lot more information with them. When a parameter is declared it will have some type (we say `kind` in the implementation to avoid keyword collision) and may have a range, description, or other properties. An undeclared parameter does not have any of that.
> 
> I imagine in rclcpp they probably just have "sensible default" values for all that extra information to store declared and undeclared parameters the same way. That makes sense in C++ where tagged unions are ... not pleasant (`std::variant` has terrible ergonomics). In this implementation we're taking an approach of "don't do/say more than what is explicitly needed". An undeclared parameter cannot have ranges, constraints, or descriptions because the user never provided any of those, so we should not pretend to have any of them in our implementation.

You could also have one parameter map, with parameter's fields for user-provided metadata being optionals, right?

rclrs/src/parameter.rs
```diff
@@ -3,3 +3,1299 @@ mod value;
...
+    /// These are not enforced by the library but are displayed on parameter description requests
+    /// and can be used by integrators to understand complex constraints.
+    pub fn constraints(mut self, constraints: impl Into<Arc<str>>) -> Self {
+        self.options.constraints = constraints.into();
+        self
+    }
+
+    /// Sets the parameter's human readable constraints.
+    /// These are not enforced by the library but are displayed on parameter description requests
+    /// and can be used by integrators to understand complex constraints.
```
nnmm, line 277: Wrong docstring

rclrs/src/parameter/value.rs
```diff
@@ -26,27 +28,257 @@ pub enum ParameterValue {
...
+
+impl ParameterVariant for ParameterValue {
+    type Range = ParameterRanges;
+    fn maybe_from(value: ParameterValue) -> Option<Self> {
+        Some(value)
+    }
+
+    fn kind() -> ParameterKind {
+        ParameterKind::Dynamic
+    }
```
nnmm, line 281: Are there good use cases for dynamic parameters? This functionality always seemed like an antipattern to me, so I left it out from my prototype to keep complexity smaller, at least initially.

rclrs/src/parameter/value.rs
```diff
@@ -26,27 +28,257 @@ pub enum ParameterValue {
...
+}
+
+/// A trait that describes a value that can be converted into a parameter.
+pub trait ParameterVariant: Into<ParameterValue> + Clone {
+    /// The type used to describe the range of this parameter.
+    type Range: Into<ParameterRanges> + Default + Clone;
+    /// Attempts to convert `value` into the requested type.
+    /// Returns `Some(Self)` if the conversion was successful, `None` otherwise.
+    // TODO(luca) should we use try_from?
```
nnmm, line 140: Yes, we should

rclrs/src/parameter/value.rs
```diff
@@ -26,27 +28,257 @@ pub enum ParameterValue {
...
     ///
     /// YAML example: `["abc", ""]`.
-    StringArray(Vec<String>),
+    StringArray(Arc<[Arc<str>]>),
+}
+
+/// Describes the parameter's type. Similar to `ParameterValue` but also includes a `Dynamic`
+/// variant for dynamic parameters.
+#[derive(Clone, Debug, PartialEq)]
+pub enum ParameterKind {
```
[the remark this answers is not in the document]
mxgrey: That's correct. Unset optional parameters and dynamic parameters are two cases where the type constraint cannot be inferred from just looking at what variant is active in the `ParameterValue` enum.

rclrs/src/node.rs
```diff
@@ -360,6 +360,84 @@ impl Node {
...
+    /// param.set(50).unwrap();
+    /// assert_eq!(param.get(), 50);
+    /// // Out of range, will return an error
+    /// assert!(param.set(200).is_err());
+    /// # Ok::<(), RclrsError>(())
+    /// ```
+    pub fn declare_parameter<T: Declarable>(
+        &self,
+        name: &str,
+        default_value: T,
```
nnmm, line 391: I think there should be a way to not pass a default value. Ideally, there would also be a way to declare a parameter with a _fixed_ value, regardless of what override was given (in `rclcpp`, it's the `ignore_overrides` parameter, which is a somewhat ugly way to solve it if you ask me).

So, if we omit the default value from this function and instead add a `default()` function to the parameter builder, then it should be a failure if no override was provided and the default was not set.

rclrs/src/node.rs
```diff
@@ -360,6 +360,84 @@ impl Node {
...
+        name: &str,
+    ) -> ParameterBuilder<'_, Option<T>> {
+        self._parameter.declare::<Option<T>>(name, None)
+    }
+
+    /// Enables usage of undeclared parameters for this node.
+    ///
+    /// Returns a `Parameter` struct that can be used to get and set all parameters.
+    pub fn use_undeclared_parameters(&self) -> Parameters {
+        self._parameter.allow_undeclared();
```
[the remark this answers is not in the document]
mxgrey: In the current API, once undeclared parameters are allowed, they're allowed for the rest of the node's lifespan. It's hard for me to imagine a use case for undoing that allowance or even what the behavior should be to undo it.

In rclcpp this decision is set in stone at initialization, so I think it's reasonable for us to not offer a way to reverse it.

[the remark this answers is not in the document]
nnmm: But I don't see the `allow_undeclared` atomic being read anywhere.

rclrs/src/node.rs
mxgrey: > I think there should be a way to not pass a default value.

That exists for Optional parameters.

For it to exist for Mandatory parameters we would need it to return an error if no parameter override value is found. That should be pretty straightforward by adding another error variant to `DeclarationError`.

> Ideally, there would also be a way to declare a parameter with a fixed value, regardless of what override was given

It should be pretty easy to add this option to the parameter builder.

rclrs/src/node.rs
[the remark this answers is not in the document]
mxgrey: It has been added in anticipation of parameter services. When parameter services are implemented on top of this in the future, this flag will determine whether to reject parameter service requests on undeclared parameters.

Until then it's true that it doesn't have any effect, simply because we've designed the rclrs API in a way that makes it unnecessary to check this for rclrs API calls.

mxgrey:
> You could also have one parameter map, with parameter's fields for user-provided metadata being optionals, right?

It's true that there are many valid ways to structure the internals of the parameter map. The question really boils down to which reads most clearly to the maintainers (the perspective of users doesn't even matter since this is just implementation detail that they can't touch).

Our feeling was that a declared/undeclared split is the most clear because it allows a clear path to upgrade an undeclared parameter into a declared parameter if the user asks for it. There are two important differences between declared and undeclared parameters that make this relevant:
* The user holds a concrete reference to each declared parameter whereas undeclared are stored entirely in the parameter map.
* You cannot redeclare a parameter that has already been declared (unless it has been undeclared). On the other hand there is nothing wrong with upgrading an undeclared parameter to declared.

If the implementation does not have a clear distinction between which parameters are declared or undeclared then implementing those behaviors becomes less obvious (but certainly not impossible).

We are also able to do a slight optimization for undeclared parameters because they do not need to be shared or modified from outside of the parameter map, but I don't think that minor optimization is a strong argument for sticking to the current implementation.

rclrs/src/node.rs
mxgrey: Alternatively, `default_value` could become a field set by the builder and we can get rid of `declare_optional_parameter`. If the user selects `.mandatory()` or `.read_only()` when neither a default nor override value is set then they get back an error when they build.

My only hesitation with this approach is in not sure if the Rust compiler can infer the parameter type if the default value is provided through a method on the builder. I have a personal distaste for turbofish syntax so I'd like to avoid forcing that on the user if possible.

Does this sound like something worth testing out @luca-della-vedova ?

rclrs/src/parameter.rs
luca-della-vedova: Not really, they are all slightly similar but the two previously defined ones are invalid, I guess an alternative could be to declare the first one as mutable and only update single fields but not much would change.

rclrs/src/node.rs
luca-della-vedova: Nice find! Fixed together with all the other nits https://github.com/ros2-rust/ros2_rust/pull/332/commits/25b1bb8b99ee7bd99045df48e575bc2cf4a392a9

rclrs/src/parameter.rs
mxgrey: I think he means the word "range" is repeated in the documentation.

luca-della-vedova: :facepalm: https://github.com/ros2-rust/ros2_rust/pull/332/commits/3e2a7bf2bec3e671a0c3bf877b19166659206928

luca-della-vedova:
> > However, the [ParameterDescriptor](https://github.com/ros2/rcl_interfaces/blob/rolling/rcl_interfaces/msg/ParameterDescriptor.msg#L32-L36) message that is returned from the `/describe_parameters` service has a range field so we need to be able to translate whatever constraint the user set to a range object that can be returned to the service call. From my understanding this translation is immediate if `Range` is a separate object but would be trickier if it is just an object that implements a trait and we share a bunch of `Box<dyn ParameterConstraint>` objects.
> 
> My understanding is that in `rclcpp`, the parameter callbacks are validators, and they don't fit the `ParameterDescriptor` message either, so it would be fine to have more general validation. My impression (without researching it really) is that this field is for informational purposes, for humans.

I suspect the field you are referring to is `additional_constraints` [here](https://github.com/ros2/rcl_interfaces/blob/rolling/rcl_interfaces/msg/ParameterDescriptor.msg#L12-L19), which indeed is just a human readable string to describe complex constraints. I was referring to the ranges at the end of the file.

There are different types of validation happening in rclcpp and other libraries.
There are parameter callbacks that, as you say, are general purpose and don't fit the `ParameterDescriptor` message. I agree we should implement them but we can do it in a followup PR.
Parameter ranges present in the [descriptor message](https://github.com/ros2/rcl_interfaces/blob/rolling/rcl_interfaces/msg/ParameterDescriptor.msg#L32-L36), these are separate from the user provided callbacks and set at declaration level when passing the `ParameterDescriptor` argument to `declare_parameter`. When a parameter is set (or declared) the range check is made [here](https://github.com/ros2/rclcpp/blob/5ffc963e1aef97928a6cc895db18d40673c655f1/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L406-L410) and only if it is successful the user provided callback is called [afterwards](https://github.com/ros2/rclcpp/blob/5ffc963e1aef97928a6cc895db18d40673c655f1/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L411-L416).

I agree that in general having a single place for constraints and unifying both ranged and callbacks into a single callback would be cleaner, but that would make it impossible for the `/describe_callbacks` to distinguish the two.

rclrs/src/parameter/value.rs
luca-della-vedova: I... Honestly don't know of any real use case but I thought I would add the functionality since it is present in the other libraries.
It is also part of [ParameterDescriptor](https://github.com/ros2/rcl_interfaces/blob/fbd32f8b8ef90fb9cefdc90d8d2f5b992fd843bb/rcl_interfaces/msg/ParameterDescriptor.msg#L25) and I thought we should have an implementation for all the fields in the message so when we implement services and users call `/describe_parameters` we have an implementation for all the features in the message fields.
The nice aspect of the proposed implementation is that it is just treated as a different enum value and not as a boolean flag that we need to check all over the place so it doesn't really clutter the code or introduce potential errors too much. If there is consensus to remove it I'm happy to do so though

rclrs/src/parameter/value.rs
luca-della-vedova: I gave it a try but I struggled a fair bit with the error type and it ended up being a lot more verbose without really adding much (I could only get it to work by converting the errors to options through `ok()`) https://github.com/ros2-rust/ros2_rust/pull/332/commits/25cd2a8d680361c60aa9e3b62ce02b1803519aef. The main part I struggled with is the fact that the trait `TryFrom` can have any type as its `Error`, so all the code that uses `try_from(_)` needs to be able to deal with this unless we constrain the type through a quite verbose bound.
Happy to refactor / revert it, or add the trait bounds, I find them a bit tough to get right.

rclrs/src/node.rs
mxgrey: Agreed on converging all those different signatures into new builder methods.

Do you have a branch where you're experimenting with this? I might be able to take a look and get the inference working.

luca-della-vedova: I'll admit not really having a clear idea about what I'm doing since I'm a bit confused by the bounds needed for these functions, but I committed the current refactoring work https://github.com/ros2-rust/ros2_rust/pull/332/commits/be4b977d890f1c6a046c748d9b509d5c5b0f1be8

mxgrey:
I just pushed some substantial changes, some of which address feedback we've received and others which tie up some loose ends. Here's a summary of the changes.

### One entry point for building parameters

We previously had `Node::declare_parameter`, `Node::declare_optional_parameter`, and a few other signatures that can be used to kick off parameter declaration in different ways depending on how you want to pass (or not pass in) a default value. All these different function signatures have been merged into a single `Node::declare_parameter` signature where the default value is set with the builder itself.

Now setting a default value is always optional, but `.mandatory()` and `.read_only()` will give errors if there is no available initial value, i.e. there is no default value, override value, or previously set value to use as an initial value for the parameter.

### Initial value discriminator

For any declaration there are up to three possible values the parameter could be initialized to:
* default value
* override value
* prior value (set while the parameter was undeclared)

Initially we thought we would provide a variety of flags to the builder to determine which value to use as the initial value in any given declaration. For example, `.ignore_override()` would prevent the override value from being used. We could provide another flag to prevent a prior value from being used. Then another flag could say that prior values should be preferred over default values but less than override values. Then ... on and on until we have too many flags to comprehend.

Instead we have settled on allowing users to provide a custom discriminator function. The discriminator would have a signature `FnOnce(AvailableValues<T>) -> Option<T>` where
```
pub struct AvailableValues<T> {
    /// The value given to the parameter builder as the default value.
    pub default_value: Option<T>,
    /// The value given as an override value, usually as a command line argument.
    pub override_value: Option<T>,
    /// A prior value that the parameter was set to before it was declared.
    pub prior_value: Option<T>,
    /// The valid ranges for the parameter value.
    pub range: ParameterRanges,
}
```
There is a default discriminator whose behavior is to prefer `prior_value` over `override_value` and prefer `override_value` over `default_value`.

We could provide a variety of out-of-the-box discriminators for whatever uses cases we expect will be relatively common if anyone has opinions on what those should look like.

We still provide `.ignore_override()` for the sake of convenience and familiarity. When `.ignore_override()` is used, `AvailableValues::override_value`  will always be set to `None`.


### `Arc<str>`

Parameter names are now stored as `Arc<str>` which will be much more memory and CPU efficient. Since parameter names are immutable there was never a benefit to storing them as `String`.

esteve:
@luca-della-vedova thanks for addressing my feedback. I'll have another look, but I'm leaning towards just merging this PR and iterate.

esteve (approved):

esteve:
@jhdcs @nnmm any more feedback? Unless there's a huge blocker, I'd like to merge this before the week ends. Any improvements can go in follow-up PRs.

jhdcs (approved):
I have no further comments.

esteve:
@luca-della-vedova @mxgrey thank you so much for all the hard work and @jhdcs @maspe36 @nnmm thank you for the thorough reviews

PR #333, opened 2023-10-20
Add support for deferred services

mxgrey:
This PR is meant to be an MVP implementation of support for deferred services. This allows users to provide a service callback that does not immediately return a response, but instead is given a `ResponseSender` that it can use to send a response from elsewhere, e.g. a task running in a task pool.

I have some ideas for how to improve the ergonomics (e.g. automatically infer the service type based on the provided callback signature) and maybe allow one `Node::create_service()` function to produce both immediate and deferred services, but I'll save those ideas for a future PR since it may be controversial for being API breaking.

### DIFF
--- a/examples/minimal_client_service/Cargo.toml
+++ b/examples/minimal_client_service/Cargo.toml
@@ -16,6 +16,10 @@ path = "src/minimal_client_async.rs"
 name = "minimal_service"
 path = "src/minimal_service.rs"
 
+[[bin]]
+name = "minimal_deferred_service"
+path = "src/minimal_deferred_service.rs"
+
 [dependencies]
 anyhow = {version = "1", features = ["backtrace"]}
 tokio = { version = "1", features = ["macros", "rt", "rt-multi-thread", "time"] }
--- a/examples/minimal_client_service/src/minimal_deferred_service.rs
+++ b/examples/minimal_client_service/src/minimal_deferred_service.rs
@@ -1,0 +1,37 @@
+use std::{env, sync::mpsc::{self, Receiver}};
+use anyhow::{Error, Result};
+use example_interfaces::srv::{AddTwoInts, AddTwoInts_Request, AddTwoInts_Response};
+
+struct Payload {
+    request: AddTwoInts_Request,
+    response_sender: rclrs::ResponseSender<AddTwoInts_Response>,
+}
+
+fn worker_thread(receiver: Receiver<Payload>) {
+    loop {
+        let Ok(Payload { request, response_sender }) = receiver.recv() else {
+            return;
+        };
+        let sum = request.a + request.b;
+        if let Err(err) = response_sender.send(AddTwoInts_Response { sum }) {
+            println!("Error while sending result: {err:?}");
+        }
+    }
+}
+
+fn main() -> Result<(), Error> {
+    let context = rclrs::Context::new(env::args())?;
+
+    let node = rclrs::create_node(&context, "minimal_deferred_service")?;
+    let (sender, receiver) = mpsc::channel();
+    let _worker = std::thread::spawn(move || worker_thread(receiver));
+
+    let _server = node
+        .create_deferred_service::<AddTwoInts, _>(
+            "add_two_ints",
+            move |request, response_sender| {
+                sender.send(Payload { request, response_sender }).ok();
+            }
+        );
+    rclrs::spin(node).map_err(|err| err.into())
+}
--- a/rclrs/src/node.rs
+++ b/rclrs/src/node.rs
@@ -14,7 +14,7 @@ pub use self::graph::*;
 use crate::rcl_bindings::*;
 use crate::{
     Client, ClientBase, Context, GuardCondition, ParameterOverrideMap, Publisher, QoSProfile,
-    RclrsError, Service, ServiceBase, Subscription, SubscriptionBase, SubscriptionCallback,
+    RclrsError, ResponseSender, Service, ServiceBase, Subscription, SubscriptionBase, SubscriptionCallback,
     ToResult,
 };
 
@@ -250,9 +250,11 @@ impl Node {
         Ok(publisher)
     }
 
-    /// Creates a [`Service`][1].
+    /// Creates a [`Service`] whose callback immediately responds to a
+    /// request. The return value of the callback provides the response.
     ///
-    /// [1]: crate::Service
+    /// See also: [`Self::create_deferred_service`]
+    //
     // TODO: make service's lifetime depend on node's lifetime
     pub fn create_service<T, F>(
         &self,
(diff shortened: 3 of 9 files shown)


esteve:
@mxgrey thanks for the PR. Could you fix the formatting issues? Thanks.

I think the best outcome would be in the end to have only one `create_service` which would convert synchronous handlers into asynchronous ones and make the whole thing asynchronous internally, similarly to how it's done in the client (e.g. a map with request ids and request handlers, or in this case, response ids)

mxgrey:
> convert synchronous handlers into asynchronous ones and make the whole thing asynchronous internally

I assume this is to reduce the complexity of having an explicit enum for the two different types of callback. So we'd funnel everything into one callback signature by just wrapping the "immediate" service type with a "deferred"-style wrapper. That seems very reasonable to me, I'll be happy to make that change.

> I think the best outcome would be in the end to have only one create_service

I agree with having one entry point for creating services. But I can see two potential APIs for that, and I'm not sure which is preferable:

### (1) Infer whether it's immediate or deferred
```
let immediate_service = node.create_service(
    "name",
    |header, request: MyServiceType_Request| -> MyServiceType_Response {
        /* ... do stuff ... */
    }
);

let deferred_service = node.create_service(
    "name",
    |request: MyServiceType_Request, sender: ResponseSender<_>| {
        /* ... do stuff ... */
    }
);
```

### (2) Explicit choice in the API
```
let immediate_service = node.create_service("name")
    .immediate(
        |header, request: MyServiceType_Request| -> MyServiceType_Response {
            /* ... do stuff ... */
        }
    );

let deferred_service = node.create_service("name")
    .deferred(
        |request: MyServiceType_Request, sender: ResponseSender<_>| {
            /* ... do stuff ... */
        }
    );
```

I think at first glance most people would assume that (1) is preferable, but the only way to implement an API like that in Rust is to use an obscure pattern that I'm going to refer to as "marker generics". The actual API implementation would look something like this:
```
impl Node {
    pub fn create_service<M, R: IntoService<M>>(
        name: &str,
        responder: R,
    ) -> Service<R::Service>;
}
```
and then we'll need to implement `IntoService` for the different signatures, something like this:
```
struct ImmediateMarker;
impl<Request: ServiceRequest> IntoService<(ImmediateMarker, Request)> for T
where
    T: Fn(&rmw_request_id_t, request: Request) -> Request::Response
{
    /* ... */
}

struct DeferredMarker;
impl<Request: ServiceRequest> IntoService<(DeferredMarker, Request)> for T
where
    T: Fn(request: Request, sender: ResponseSender<Request::Response>)
{
    /* ... */
}
```

This can absolutely work, and the user ergonomics is pretty decent, but the "marker generics" pattern needed for the implementation can be confusing to maintain. It's also very difficult for a user to understand by looking at the function signature, so it'll be important to have very clear examples of what's possible in the documentation. Lastly, compilation errors are very difficult to interpret when this pattern is used.

On the other hand, the API for option (2) isn't as pretty but the explicitness might make its usage more clear to users. It will also give much more helpful compilation errors if a user does anything wrong.

I don't actually have a strong preference between (1) and (2). I think the advantages and disadvantages to each are pretty well balanced, to the point that I'm personally okay with either. But we have to choose one or the other, so I'd appreciate input on which to draft first. We can always change our minds afterwards once we see the outcome.

esteve:
@mxgrey I like the first one, but with one change:

```rust

let immediate_service = node.create_service(
    "name",
    |header, request: MyServiceType_Request| -> MyServiceType_Response {
        /* ... do stuff ... */
    }
);

let deferred_service = node.create_service(
    "name",
    |header, request: MyServiceType_Request| -> Future<MyServiceType_Response> {
        /* ... do stuff ... */
    }
);
```

This would simplify the internal code so that the non-async handler can be wrapped into a `Future` and just handle it as that. And additionally, we could pass `async` blocks or functions as handlers.

esteve:
> This would simplify the internal code so that the non-async handler can be wrapped into a `Future` and just handle it as that. And additionally, we could pass `async` blocks or functions as handlers.

So, essentially this would only be only one API, and non-async handlers can be converted into async handlers implicitly.

mxgrey:
> This would simplify the internal code so that the non-async handler can be wrapped into a Future and just handle it as that.

This would work in C++, Python, or JavaScript, but in Rust [`Future` is a trait](https://doc.rust-lang.org/std/future/trait.Future.html), not a container or a wrapper. We could allow the user to return something like `impl Future<Output=Response>` and then box that to store it somewhere, but we would still need an async executor (e.g. tokio) to process the Futures because otherwise they will all just sit idle.

I absolutely want to support what you're describing in the long-run, but I would put it under a third category of `Async`, so ultimately we'd support three signatures:
1. Immediate
2. Deferred
3. Async

And I think the `Async` signature would have to be feature-gated by a choice of async runtime, e.g. we could have an `async_tokio` feature that gives access to a tokio-based multithreaded executor and then allows the `Async` service signature.

But I anticipate that may be a considerable amount of work, and in the meantime I have existing use cases where I'd like to handle multi-threading from outside of rclrs by punting the processing off to a threadpool, and that's something we can accomplish with this `ResponseSender` pattern without the need to implement multithreaded execution inside of rclrs.

mxgrey:
After giving this some more thought, feature-gating the Async signature based on the availability of a runtime sounds like it would be a terrible situation for both users and maintainers.

I can think of two ways to avoid that:
1. Make our own very simplistic async runtime (probably based on [this example](https://rust-lang.github.io/async-book/02_execution/04_executor.html)) as a default async runtime and then let users swap it out with a tokio based runtime via features or a downstream crate.
2. Have a `tokio_async` default feature which users can replace if they prefer a different runtime.

I still would like the three different function signatures because I need the deferred service approach for the way I want to use ros2-rust, but that signature can be wrapped into a function that returns a future, so it should work nicely with an async framework.

Let me know if you have a strong preference, otherwise I'll aim to do option (1).

esteve:
@mxgrey you mean something like a wrapper that `ros2-rust` targets, which can be backed by any of the actual async runtimes? If that's the case, yes I like that solution, I don't want `rclrs` to impose a specific async runtime, even if by de facto we all use tokio.

mxgrey:
Yes, that's the idea. We'll define some traits for an "executor" and then downstream crates could create implementations for those traits based on various third-party async runtimes, e.g. tokio and async-std.

But for this to be a good arrangement, I think we should implement our own **very simple** single-threaded async executor that allows users to get their nodes running without an additional dependency. The [example here](https://rust-lang.github.io/async-book/02_execution/04_executor.html) shows all the code we need to implement a simple single-threaded executor, so I think this is pretty reasonable for us to do.

I'm going to close this PR in favor of first implementing an async executor for rclrs and then building the async services on top of that.

PR #338, opened 2023-10-24
Add VS Code Setup Instructions

Carter12s:
Hopefully this is useful.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,5 +21,9 @@ RUN pip install --upgrade pytest
 # Install the colcon-cargo and colcon-ros-cargo plugins
 RUN pip install git+https://github.com/colcon/colcon-cargo.git git+https://github.com/colcon/colcon-ros-cargo.git
 
+# Default source the ros environment, this correctly sets up the environment
+# For vscode and rust-analyzer to work inside the docker image
+RUN echo "source /opt/ros/$ROS_DISTRO/setup.bash" >> /root/.bashrc
+
 RUN mkdir -p /workspace && echo "Did you forget to mount the repository into the Docker container?" > /workspace/HELLO.txt
 WORKDIR /workspace
--- a/docs/building.md
+++ b/docs/building.md
@@ -1,12 +1,10 @@
 # Building `ros2_rust` packages
-
 This is a more detailed guide on how to build ROS 2 packages written in Rust that expands on the minimal steps in the README.
 
 In this guide, the Humble distribution of ROS 2 is used, but newer distributions can be used by simply replacing 'humble' with the distribution name everywhere.
 
 
 ## Choosing a workspace directory and cloning `ros2_rust`
-
 A "workspace directory", or just "workspace", is simply a directory of your choosing that contains your `ros2_rust` checkout and potentially other ROS 2 packages. It will also usually be your working directory for building. There is only one limitation: It must not contain the ROS 2 installation, so it can't be `/`, for instance. Note that this has **nothing** to do with a [`cargo` workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html).
 
 If you don't have a workspace directory already, simply create an empty directory in a convenient location.
@@ -23,19 +21,16 @@ git clone https://github.com/ros2-rust/ros2_rust.git src/ros2_rust
 
 
 ## Environment setup
-
-Building `rclrs` requires a standard [ROS 2 installation](https://docs.ros.org/en/humble/Installation.html), and a few extensions.
-These extensions are: `colcon-cargo`, `colcon-ros-cargo`, `cargo-ament-build`. The first two are `colcon` plugins, and the third is a `cargo` plugin.
+Building `rclrs` requires a standard [ROS 2 installation](https://docs.ros.org/en/humble/Installation.html), and a few extensions. These extensions are: `colcon-cargo`, `colcon-ros-cargo`, `cargo-ament-build`. The first two are `colcon` plugins, and the third is a `cargo` plugin.
 
 The `libclang` library is also required for automatically generating FFI bindings with `bindgen`. See the [`bindgen` docs](https://rust-lang.github.io/rust-bindgen/requirements.html) on how to install it. As a side note, on Ubuntu the `clang` package is not required, only the `libclang-dev` package.
 
 The `python3-vcstool` package is used in [importing auxiliary repositories](#importing-repositories). It can also be installed through `pip` instead of `apt`.
 
-You can either install these dependencies on your computer, or use the [premade Docker image](#using-the-docker-image).
+You can either install these dependencies on your computer, or use the [premade Docker image](#option-2:-using-the-docker-image).
 
 
 ### Option 1: Installing the dependencies
-
 The exact steps may differ between platforms, but as an example, here is how you would install the dependencies on Ubuntu:
 
 <!--- These steps should be kept in sync with README.md --->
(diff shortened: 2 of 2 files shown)


docs/building.md
```diff
@@ -190,8 +174,50 @@ cargo build
 
 If you'd like to not have any of that "overriding dependencies through  `.cargo/config.toml`", you can do that too of course. You just need to use concrete paths for any dependency that isn't published on `crates.io`, such as message packages. For instance, you might have to write `std_msgs = {path = '/workspace/install/std_msgs/share/std_msgs/rust'}`.
 
+## VS Code Setup with .devcontainer
+A convienent (and reccomended) development setup is use the provided Dockerfile with VS Code's `Dev Containers` extension. To set this up:
```
jhdcs, line 178:
```suggestion
A convenient (and recommended) development setup is use the provided Dockerfile with VS Code's `Dev Containers` extension. To set this up:
```
A couple typos.

docs/building.md
```diff
@@ -190,8 +174,50 @@ cargo build
...
+
+- Follow the steps in [using the docker image](#option-2-using-the-docker-image) to build an imaged tagged with `ros2_rust_dev` and your preffered ROS2 version.
```
jhdcs, line 180:
```suggestion
- Follow the steps in [using the docker image](#option-2-using-the-docker-image) to build an imaged tagged with `ros2_rust_dev` and your preferred ROS2 version.
```
Another typo.

docs/building.md
```diff
@@ -190,8 +174,50 @@ cargo build
...
+- Install [VS Code](https://code.visualstudio.com/download) and the [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension.
+- In the root folder of your workspace (alongside `src/`) create a folder named .devcontainer with a file named devcontainer.json and the contents:
```
jhdcs, line 182:
```suggestion
- In the root folder of your workspace (alongside `src/`) create a folder named `.devcontainer` with a file named `devcontainer.json` and the contents:
```
Just highlighting the file names a bit more.

docs/building.md
```diff
@@ -190,8 +174,50 @@ cargo build
...
+        "vscode": {
+            "extensions": [
+                "rust-lang.rust-analyzer",
+                "vadimcn.vscode-lldb"
+            ]
+        }
+    }
+}
+```
+- In the root folder of your workspace (alongside `src/`) create a folder named .vscode with a file named settings.json and the contents:
```
jhdcs, line 199:
```suggestion
- In the root folder of your workspace (alongside `src/`) create a folder named `.vscode` with a file named `settings.json` and the contents:
```
Again, just highlighting file/folder names.

jhdcs (changes requested):
Just a couple typos that I noticed while going through this. Apart from that, this looks very well done!

@esteve a policy question - do we want to take a stance as to which IDE we recommend people use when working with `rclrs`? While I do enjoy VSCode for development, I could also see us wanting to remain a bit neutral in that regard. Not saying that this contribution can't be added, mind you, I'm just wondering if we want to say that we recommend people use VSCode, is all. I'll defer to your judgement on that.

Regardless, thank you VERY much for this @Carter12s! I think this will be very useful for people that want to use VSCode with ROS 2 Rust!

Carter12s:
> do we want to take a stance as to which IDE we recommend people use when working with `rclrs`?

Fair point, happy to strip that language from MR. In my opinion it is extremely important to provide a "Batteries Included" path for people getting started with a new complex system (like ROS2). I'm imagining the use case of someone new to both ROS and Rust thinking this could be a good place to start with both (probably a lot of grad students in this position as we speak).

I would say anyone who has enough experience to have an opinion against using VS Code probably doesn't need these instructions, and probably won't be offended by the suggestion. But, either way understand the neutral stance if desired.

Carter12s:
@jhdcs Ready for a second look, all suggestions addressed and caught a couple other missing words.

jhdcs (approved):
These changes look good to me! Thank you again!

Dockerfile
```diff
@@ -21,5 +21,9 @@ RUN pip install --upgrade pytest
 # Install the colcon-cargo and colcon-ros-cargo plugins
 RUN pip install git+https://github.com/colcon/colcon-cargo.git git+https://github.com/colcon/colcon-ros-cargo.git
 
+# Default source the ros environment, this correctly sets up the environment
+# For vscode and rust-analyzer to work inside the docker image
+RUN echo "source /opt/ros/$ROS_DISTRO/setup.bash" >> /root/.bashrc
```
luca-della-vedova, line 26: I personally disagree with the approach of adding `source` to the user's `bashrc`.

Apart from being (imho) a bad habit, it is convenient for "most" cases but can cause footguns in others. For example let's say a user wanted to do any change on the ROS 2 source code that requires them to compile it from source.

If they did a ROS 2 source build and naively sourced the `install/setup.bash` they would end up with two overlays. I believe multiple overlays can create a fair bit of trouble if ABI / API compatibility is not respected, as documented in this discussion https://github.com/colcon/colcon-core/issues/465 and in the extensive warnings printed at compile time.
Also, there is a minor footgun for the source build case in which, since a ROS installation is automatically sourced, if users forget the `source install/setup.bash` command, their commands will "just work" which might not be what they expect.

Carter12s: No that is a super fair point, and I felt dirty doing this.

Do you know of a workaround / alternative? Party of why I wrote these instructions was because of how hard I found it to get the basic Rust tooling that I'm used to having working with this project.

VS Code supports loading environment variables in a [few different ways](https://code.visualstudio.com/remote/advancedcontainers/environment-variables), but I don't think there is any good way to "export" the right variable definitions out of a terminal instance inside the container and then load them into VS Code. Even if you did this it would be really brittle (maybe a VS Code plugin could update automatically).

I can find[ stack-overflow questions](https://stackoverflow.com/questions/69848116/vscode-how-to-source-environment-variable-files-like-setup-bash) looking for a way of doing this, but no answers.

The crux of the issue is that `rust-analyzer` and programs launched via its launch configurations, don't provide a way to say "run this bash command ahead of time"...

I can keep digging... But this shouldn't be this hard.

Carter12s: So I went on a rabbit hole to try to find other ROS projects that have "solved this problem".

I found Nav2 which is doing a fancier (and probably better) version of sourcing in the .bashrc: https://github.com/ros-planning/navigation2/blob/main/Dockerfile#L153

I'm don't 100% understand what they are doing with their UNDERLAY_WS, but it seems like the better thing to do would be to invoke `colcon build` in the Dockerfile to get setup.bash to generate, and then source that instead of directly sourcing `iron/setup.bash`

Is that a good enough solution?

I'll be frank I'm coming from ROS1 and haven't touched ROS2 workspaces yet, so learning while doing here.

felixf4xu: In my opinion, it's fine to source ros setup.bash in the docker's .bashrc.

To my understanding, this is a docker setup, not a local host system setup. The user of the docker has less possibility to rebuild ROS2 from source.

roboticswithjulia:
Hello, there is a MR https://github.com/ros2-rust/ros2_rust/pull/435 that may be  a better way to help people to use docker and devcontainers, if I can have your feedback, maybe we can consider to update the new MR with useful info from this MR and proceed closing this one. Thank you!!

esteve:
@Carter12s sorry this slipped through the cracks and never got a proper review, I'm closing it, but thanks for the PR!

PR #342, opened 2023-11-07
Add parameter services

luca-della-vedova:
This PR builds on top of the basic parameter implementation and adds (to my knowledge) fully functional parameter services.
Note that parameter callbacks are still not implemented so, for example, users have no way to register a callback to react to parameter change events when using services to set values, even though the value itself will be updated and can be fetched through a `parameter.get()` call.
I didn't go into the rabbit hole of dealing with updates to the `use_sim_time` parameter to change the clock type at runtime between simulated and wall clock, but we should probably figure out how to do it soon, probably when we have such callbacks.

I tested this manually with all the use cases I could think of and it seems to work fine. 
I also added a fair bit of integration tests under the `rclrs_tests` and they all pass, the majority of this PR is actually test code.

## Overview

There was already a skeleton for parameter services with empty callbacks, this PR fills it up by creating the actual service implementations.
There are some loose ends that I believe could be cleaned up (I'd like to address at the very least all the `TODO(luca)` in the code, any hint is welcome!) but the bulk of the functionality should be there.

### Test it!

Since `use_sim_time` has been introduced a very simple test can be done by running any example and doing:

```
ros2 param list
# Should return `use_sim_time`
ros2 param set /minimal_subscriber use_sim_time 42
# Should fail
ros2 param set /minimal_subscriber use_sim_time true
# Should succeed and set the parameter to true
ros2 param get /minimal_subscriber use_sim_time
```

Otherwise you can declare your own node with parameters, or just look at the tests under `rclrs_tests/src/parameter_service_tests.rs`


luca-della-vedova:
Note that apart from the TODOs in the PR there is an additional gotcha.
For the sake of simplicity (and because parameters were not really implemented), when adding time and clocks in #325, I left out runtime update of `use_sim_time` (what is currently accomplished by a parameter callback in other client libraries). The parameter is only checked, and acted on, once at node creation time.
This PR would add an API for users to set parameters from the CLI `ros2 param set` which could cause confusing behaviors if the parameter is updated at runtime but the change is not reflected.

A true fix involves figuring out a way to trigger callbacks on parameter changes that could involve new public facing APIs and add considerable burden to the PR.
My recommendation would be to set the `use_sim_time` parameter as `ReadOnly` for now and revisit in the future how we can handle runtime updates.
My personal opinion is that changing the underlying clock base at runtime is a very niche (and in most cases very dangerous) approach so hopefully forbidding it explicitly in the API will offer one less footgun, until we figure out if / how we should support it.

rclrs/src/parameter/service.rs
```diff
@@ -1,89 +1,264 @@
...
+            move |_req_id: &rmw_request_id_t, req: GetParameterTypes_Request| {
+                let map = map.lock().unwrap();
+                let types = req
+                    .names
+                    .into_iter()
+                    .map(|name| {
+                        let name = name.to_cstr().to_str().ok()?;
+                        map.storage.get(name).map(|s| s.to_parameter_type())
+                    })
+                    .collect::<Option<_>>()
```
mxgrey, line 94: If I'm grokking this correctly, if the user included even **one** parameter name in the request that doesn't exist in the storage, we're going to end up collecting the **whole** sequence (including any valid parameter names) into a single `None` and then `unwrap_or_default()` below will produce an empty sequence.

According to [the documentation for GetParameterTypes](https://docs.ros2.org/galactic/api/rcl_interfaces/srv/GetParameterTypes.html) that would not be the correct behavior. Instead we should return a value of `PARAMETER_NOT_SET` in the same location in the sequence where the unknown parameter was placed. I think we can fix this by simply replacing this and the next line with
```
.map(|type_| type_.unwrap_or(ParameterType::PARAMETER_NOT_SET)
.collect();
```

rclrs/src/parameter/service.rs
```diff
@@ -1,89 +1,264 @@
...
+                            ParameterStorage::Undeclared(_) => ParameterDescriptor {
+                                name: name.into(),
+                                dynamic_typing: true,
+                                ..Default::default()
+                            },
+                        };
+                        descriptor.type_ = storage.to_parameter_type();
+                        Some(descriptor)
+                    })
+                    .collect::<Option<_>>()
```
mxgrey, line 76: Similar to my comment on `get_parameter_types`, I don't think this will give us the correct behavior which [expects the sequence in the response to always have the same length as the sequence in the request](https://docs.ros2.org/galactic/api/rcl_interfaces/srv/DescribeParameters.html).

I think we can map the `None` cases to a ParameterDescriptor which is default-initialized except for its name, which should match the request string in the same place in the sequence.

rclrs/src/parameter/service.rs
```diff
@@ -1,89 +1,264 @@
...
+                                        .clone()
+                                        .map(|v| v.into())
+                                        .unwrap_or_default(),
+                                ),
+                                DeclaredValue::ReadOnly(v) => Some(v.clone().into()),
+                            },
+                            ParameterStorage::Undeclared(value) => Some(value.clone().into()),
+                        }
+                    })
+                    .collect::<Option<_>>()
```
mxgrey, line 126: Same concern about collection here. Unknown parameter names should return a value of unset.

rclrs/src/parameter/value.rs
```diff
@@ -319,6 +324,107 @@ impl ParameterVariant for ParameterValue {
...
+                type_: ParameterType::PARAMETER_DOUBLE_ARRAY,
+                double_array_value: (*v).into(),
+                ..Default::default()
+            },
+            ParameterValue::StringArray(v) => RmwParameterValue {
+                type_: ParameterType::PARAMETER_STRING_ARRAY,
+                string_array_value: v
+                    .iter()
+                    .map(|v| v.clone().into())
+                    .collect::<Vec<_>>()
```
mxgrey, line 375: Since `Sequence` implements `FromIterator`, I think you can change these next two lines into simply
```
.collect();
```
That will avoid an unnecessary heap allocation to produce a `Vec` followed by another heap allocation to move everything over into the sequence.

rclrs/src/parameter/value.rs
```diff
@@ -319,6 +324,107 @@ impl ParameterVariant for ParameterValue {
...
+pub enum RmwParameterConversionError {
+    /// The parameter type was not valid.
+    InvalidParameterType,
+}
+
+impl TryFrom<RmwParameterValue> for ParameterValue {
+    type Error = RmwParameterConversionError;
+
+    fn try_from(param: RmwParameterValue) -> Result<Self, Self::Error> {
+        // TODO(luca) how to deal with PARAMETER_NOT_SET? Should we allow service calls to unset
```
mxgrey, line 394: Since we support "optional" parameters, I think it makes sense to allow parameter values to be unset by a service call. But I'm not sure if there's guidance from the ROS 2 documentation on what behavior is actually expected. Also supporting that would diminish some of the elegance we get with `TryFrom` here.

rclrs/src/parameter/value.rs
```diff
@@ -392,6 +498,35 @@ impl ParameterValue {
...
+            ParameterValue::ByteArray(_) => ParameterType::PARAMETER_BYTE_ARRAY,
+            ParameterValue::BoolArray(_) => ParameterType::PARAMETER_BOOL_ARRAY,
+            ParameterValue::IntegerArray(_) => ParameterType::PARAMETER_INTEGER_ARRAY,
+            ParameterValue::DoubleArray(_) => ParameterType::PARAMETER_DOUBLE_ARRAY,
+            ParameterValue::StringArray(_) => ParameterType::PARAMETER_STRING_ARRAY,
+        }
+    }
+
+    /// Returns the `ParameterKind` for the parameter as if it was a static parameter.
+    pub(crate) fn static_kind(&self) -> ParameterKind {
```
mxgrey, line 517: The name `static_kind` feels a bit odd to me. I would suggest renaming it to the `current_kind` or even just `kind` because this is a method on the `ParameterValue` enum, and there's no such thing as a "dynamic" value (the word "dynamic" only makes sense when talking about a parameter, not a specific value).

rclrs_tests/src/parameter_service_tests.rs
```diff
@@ -0,0 +1,618 @@
...
+        let types = names_and_types
+            .get("/names_types/node/get_parameter_types")
+            .unwrap();
+        assert!(types.contains(&"rcl_interfaces/srv/GetParameterTypes".to_string()));
+    };
+
+    std::thread::sleep(std::time::Duration::from_millis(100));
+
+    let service_names_and_types = node.node.get_service_names_and_types()?;
+    check_names_and_types(service_names_and_types);
```
mxgrey, line 120: I'm very confused about why we put all the tests into a closure up above and then call that closure only once here. I would understand using a closure if we called it multiple times, but we don't seem to do that.

It's not wrong, but it leaves me wondering if I'm misunderstanding something about the test.

rclrs_tests/src/parameter_service_tests.rs
```diff
@@ -0,0 +1,618 @@
...
+        // Getting both existing and non existing parameters should fail
+        let request = GetParameters_Request {
+            names: seq!["bool".into(), "non_existing".into()],
+        };
+        let client_finished = Arc::new(RwLock::new(false));
+        let call_done = client_finished.clone();
+        get_client
+            .async_send_request_with_callback(&request, move |response: GetParameters_Response| {
+                *call_done.write().unwrap() = true;
+                assert_eq!(response.values.len(), 0);
```
mxgrey, line 307: As mentioned in earlier comments, this test doesn't match up with the ROS 2 documented expectations of the service, which is that the response always has the exact same size sequence as the request.

rclrs_tests/src/parameter_service_tests.rs
```diff
@@ -0,0 +1,618 @@
...
+                    assert!(desc[3].dynamic_typing);
+                    assert!(!desc[3].read_only);
+                },
+            )
+            .unwrap();
+        try_until_timeout(|| *client_finished.read().unwrap())
+            .await
+            .unwrap();
+
+        // If a describe parameters request is sent with a non existing parameter, an empty
```
mxgrey, line 543: As mentioned in earlier review comments, this is not the documented behavior for ROS 2.

rclrs_tests/src/parameter_service_tests.rs
```diff
@@ -0,0 +1,618 @@
...
+            names: seq!["bool".into(), "non_existing".into()],
+        };
+        let client_finished = Arc::new(RwLock::new(false));
+        let call_done = client_finished.clone();
+        describe_client
+            .async_send_request_with_callback(
+                &request,
+                move |response: DescribeParameters_Response| {
+                    *call_done.write().unwrap() = true;
+                    assert_eq!(response.descriptors.len(), 0);
```
mxgrey, line 555: Not the expected behavior.

rclrs_tests/src/parameter_service_tests.rs
```diff
@@ -0,0 +1,618 @@
...
+            names: seq!["bool".into(), "non_existing".into()],
+        };
+        let client_finished = Arc::new(RwLock::new(false));
+        let call_done = client_finished.clone();
+        get_types_client
+            .async_send_request_with_callback(
+                &request,
+                move |response: GetParameterTypes_Response| {
+                    *call_done.write().unwrap() = true;
+                    assert_eq!(response.types.len(), 0);
```
mxgrey, line 604: Not the expected behavior.

rclrs_tests/src/parameter_service_tests.rs
```diff
@@ -0,0 +1,618 @@
...
+                    assert_eq!(response.types[2], ParameterType::PARAMETER_DOUBLE);
+                    assert_eq!(response.types[3], ParameterType::PARAMETER_STRING);
+                },
+            )
+            .unwrap();
+        try_until_timeout(|| *client_finished.read().unwrap())
+            .await
+            .unwrap();
+
+        // If a get parameter types request is sent with a non existing parameter, an empty
```
mxgrey, line 592: Not the expected behavior.

mxgrey (changes requested):
I think this is generally a very good implementation of parameter services.

I've left some comments, many of which aren't really actionable in this PR but more to stir a conversation and see if anyone else has thoughts about how to improve quality in the future.

Some comments do require action, though. The way the services respond to requests that include unknown parameter names does not match up with the ROS 2 documentation, which I link to in my inline comments. It should be very easy to correct that in the code, but it's a significant correctness detail that we shouldn't overlook.

rclrs/src/parameter/service.rs
luca-della-vedova: I guess this is one of the cases where documentation is different from actual behavior.
I agree that it is puzzling but all of these were actually a conscious choice. What I did was try these edge cases on `rclcpp` and check how it behaves. Since this is for `get_parameter_types` here is a snippet from my local testing under iron:

```
lucadv@ros:~$ ros2 service call /minimal_publisher/get_parameter_types rcl_interfaces/srv/GetParameterTypes "{names: [use_sim_time]}"       
requester: making request: rcl_interfaces.srv.GetParameterTypes_Request(names=['use_sim_time'])

response:
rcl_interfaces.srv.GetParameterTypes_Response(types=[1])

lucadv@ros:~$ ros2 service call /minimal_publisher/get_parameter_types rcl_interfaces/srv/GetParameterTypes "{names: [use_sim_time, hello]}"
requester: making request: rcl_interfaces.srv.GetParameterTypes_Request(names=['use_sim_time', 'hello'])

response:
rcl_interfaces.srv.GetParameterTypes_Response(types=[])
```

As you can see, the actual behavior of `rclcpp` is that it returns an empty array as soon as one of the names is invalid.
The implementation for get_parameter_types is [here](https://github.com/ros2/rclcpp/blob/1c350d0d7fb9c7158e0a39057112486ddbd38e9a/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L1007-L1030), you can see how in that implementation the message creation actually throws an exception if any parameter does not exist, and that exception will be caught downstream and send an empty response.
This same behavior is present in `rclpy`, [here](https://github.com/ros2/rclpy/blob/7a7f23e0d7e51dd244001ef606b97f1153e5a97e/rclpy/rclpy/node.py#L615-L646).
What I did _not_ notice however, until you pointed it out and I went to cross check, is that the behavior you mention is indeed what happens when nodes allow undeclared parameters, so I will fix this now and add a test case for it.

rclrs/src/parameter/service.rs
luca-della-vedova: Same answer as `get_parameter_types`! https://github.com/ros2-rust/ros2_rust/pull/342#discussion_r1529627501. Specifically this is what happens when I run on a `rclcpp` example:

```
lucadv@ros:~$ ros2 service call /minimal_publisher/describe_parameters rcl_interfaces/srv/DescribeParameters "{names: [use_sim_time]}"       
requester: making request: rcl_interfaces.srv.DescribeParameters_Request(names=['use_sim_time'])

response:
rcl_interfaces.srv.DescribeParameters_Response(descriptors=[rcl_interfaces.msg.ParameterDescriptor(name='use_sim_time', type=1, description='', additional_constraints='', read_only=False, dynamic_typing=False, floating_point_range=[], integer_range=[])])

lucadv@ros:~$ ros2 service call /minimal_publisher/describe_parameters rcl_interfaces/srv/DescribeParameters "{names: [use_sim_time, hello]}"
requester: making request: rcl_interfaces.srv.DescribeParameters_Request(names=['use_sim_time', 'hello'])

response:
rcl_interfaces.srv.DescribeParameters_Response(descriptors=[])
```

But again same that it should be behaving differently if undeclared parameters are allowed and I'll be fixing this.

rclrs/src/parameter/service.rs
luca-della-vedova: Same behavior as the others:

```
lucadv@ros:~$ ros2 service call /minimal_publisher/get_parameters rcl_interfaces/srv/GetParameters "{names: [use_sim_time]}"                 
requester: making request: rcl_interfaces.srv.GetParameters_Request(names=['use_sim_time'])

response:
rcl_interfaces.srv.GetParameters_Response(values=[rcl_interfaces.msg.ParameterValue(type=1, bool_value=False, integer_value=0, double_value=0.0, string_value='', byte_array_value=[], bool_array_value=[], integer_array_value=[], double_array_value=[], string_array_value=[])])

lucadv@ros:~$ ros2 service call /minimal_publisher/get_parameters rcl_interfaces/srv/GetParameters "{names: [use_sim_time, hello]}"
requester: making request: rcl_interfaces.srv.GetParameters_Request(names=['use_sim_time', 'hello'])

response:
rcl_interfaces.srv.GetParameters_Response(values=[])
```

rclrs/src/parameter/service.rs
mxgrey: Interesting... Iirc restricting to only declared parameters was a later add on of ROS 2, not in the initial design. I suspect when ability to restrict it was added, the person implementing it was not mindful of the documented behavior of the service. This is probably something to report to the other client library maintainers.

rclrs_tests/src/parameter_service_tests.rs
luca-della-vedova: I feel at some point of the writing the closure was reused, but I agree it doesn't make any sense now. https://github.com/ros2-rust/ros2_rust/pull/342/commits/a7d1befbe078418fca4852b200e3e1b6c4a62f9f

rclrs/src/parameter/value.rs
luca-della-vedova: https://github.com/ros2-rust/ros2_rust/pull/342/commits/a7d1befbe078418fca4852b200e3e1b6c4a62f9f

rclrs/src/parameter/value.rs
luca-della-vedova: Done https://github.com/ros2-rust/ros2_rust/pull/342/commits/a7d1befbe078418fca4852b200e3e1b6c4a62f9f

rclrs/src/parameter/range.rs
```diff
@@ -0,0 +1,204 @@
...
+                        step: range.step.unwrap_or(0).try_into().unwrap(),
+                    }]
+                }
+            })
+            .unwrap_or_default();
+        let float_range = self
+            .float
+            .as_ref()
+            .map(|range| {
+                // TODO(luca) Double check whether we should use MIN/MAX or INFINITY/NEG_INFINITY
```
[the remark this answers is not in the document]
luca-della-vedova: Done https://github.com/ros2-rust/ros2_rust/pull/342/commits/a7d1befbe078418fca4852b200e3e1b6c4a62f9f

rclrs/src/parameter.rs
```diff
@@ -605,9 +454,101 @@ enum ParameterStorage {
...
+    }
+
+    /// Stores the requested parameter in the map.
+    fn store_parameter(&mut self, name: Arc<str>, value: ParameterValue) {
+        match self.storage.entry(name) {
+            Entry::Occupied(mut entry) => match entry.get_mut() {
+                ParameterStorage::Declared(storage) => match &storage.value {
+                    DeclaredValue::Mandatory(p) => *p.write().unwrap() = value,
+                    DeclaredValue::Optional(p) => *p.write().unwrap() = Some(value),
+                    DeclaredValue::ReadOnly(_) => {}
```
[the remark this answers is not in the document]
luca-della-vedova: Yea I agree it should never happen but in case it does an `unreachable` panic is better than a silent failure, done https://github.com/ros2-rust/ros2_rust/pull/342/commits/a7d1befbe078418fca4852b200e3e1b6c4a62f9f

rclrs/src/parameter/service.rs
```diff
@@ -1,89 +1,264 @@
...
+        node: &Node,
+        parameter_map: Arc<Mutex<ParameterMap>>,
+    ) -> Result<Self, RclrsError> {
+        let fqn = node.fully_qualified_name();
+        // TODO(luca) make sure it is OK to have an Arc instead of a Weak here and cleanup on
+        // destruction is made for the parameter map.
+        let map = parameter_map.clone();
+        let describe_parameters_service = node.create_service(
+            &(fqn.clone() + "/describe_parameters"),
+            move |_req_id: &rmw_request_id_t, req: DescribeParameters_Request| {
```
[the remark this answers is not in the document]
luca-della-vedova: Very good point, refactored all of them https://github.com/ros2-rust/ros2_rust/pull/342/commits/a7d1befbe078418fca4852b200e3e1b6c4a62f9f

rclrs/src/parameter/service.rs
```diff
@@ -1,89 +1,264 @@
...
 
 impl ParameterService {
-    pub fn new(rcl_node_mtx: Arc<Mutex<rcl_node_t>>) -> Result<Self, RclrsError> {
-        let describe_parameters_service = Service::new(Arc::clone(&rcl_node_mtx),
-            "describe_parameters",
-            |req_id: &rmw_request_id_t, req: DescribeParameters_Request| {
-                DescribeParameters_Response {
-                    descriptors: seq![]
-                }
+    pub(crate) fn new(
```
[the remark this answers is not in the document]
luca-della-vedova: Refactored all of them https://github.com/ros2-rust/ros2_rust/pull/342/commits/a7d1befbe078418fca4852b200e3e1b6c4a62f9f

rclrs/src/parameter/service.rs
luca-della-vedova: For now I just addressed the difference in behavior between my implementation and the other client libraries in https://github.com/ros2-rust/ros2_rust/pull/342/commits/136047989bd259c46a5638ff41a91430b8f5835a.
Depending on how the discussion on other client libraries goes I'm happy to revert it but at least for now we behave the same way.
For what it's worth I backtracked to this PR https://github.com/ros2/rclcpp/pull/495 when the change was introduced. Before the change the behavior was what you mentioned, after the change it is the current one.

rclrs/Cargo.toml
```diff
@@ -34,6 +34,8 @@ rosidl_runtime_rs = "0.4"
 tempfile = "3.3.0"
 # Needed for publisher and subscriber tests
 test_msgs = {version = "*"}
+# Needed for parameter service tests
+tokio = { version = "*", features = ["rt", "time", "macros"] }
```
jhdcs, line 38: Are we all ok with adding a test dependency? Personally, I feel it's all right, just want to make sure the other maintainers are cool with it.

rclrs/src/parameter.rs
```diff
@@ -605,9 +454,101 @@ enum ParameterStorage {
...
+                ParameterKind::ByteArray => ParameterType::PARAMETER_BYTE_ARRAY,
+                ParameterKind::BoolArray => ParameterType::PARAMETER_BOOL_ARRAY,
+                ParameterKind::IntegerArray => ParameterType::PARAMETER_INTEGER_ARRAY,
+                ParameterKind::DoubleArray => ParameterType::PARAMETER_DOUBLE_ARRAY,
+                ParameterKind::StringArray => ParameterType::PARAMETER_STRING_ARRAY,
+                ParameterKind::Dynamic => match &s.value {
+                    DeclaredValue::Mandatory(v) => v.read().unwrap().rcl_parameter_type(),
+                    DeclaredValue::Optional(v) => v
+                        .read()
+                        .unwrap()
```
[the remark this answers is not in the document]
jhdcs: I think documenting the `unwraps` is a reasonable idea, though this may mean that we need to dig through our already-written code to make sure any other `unwraps` are in compliance with these requirements.

rclrs/src/parameter.rs
```diff
@@ -605,9 +454,101 @@ enum ParameterStorage {
...
+                        "Parameter was not declared and undeclared parameters are not allowed",
+                    );
+                }
+            }
+        }
+        Ok(value)
+    }
+
+    /// Stores the requested parameter in the map.
+    fn store_parameter(&mut self, name: Arc<str>, value: ParameterValue) {
```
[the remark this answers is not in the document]
jhdcs: Agreed. If and when we implement benchmark tests for `rclrs` we can decide if such an optimization is needed. For now, I think that getting parameter services up-and-running in a usable state takes priority.

rclrs/src/parameter/service.rs
```diff
@@ -1,89 +1,994 @@
...
+                crate::spin_once(inner_node.clone(), Some(std::time::Duration::ZERO)).ok();
+                crate::spin_once(client.clone(), Some(std::time::Duration::ZERO)).ok();
+                *inner_done.read().unwrap()
+            })
+            .await
+            .unwrap();
+        });
+
+        let res = tokio::task::spawn(async move {
+            // Desctibe all parameters
```
jhdcs, line 842:
```suggestion
            // Describe all parameters
```

jhdcs (changes requested):
Just a couple minor tweaks, but looks good otherwise!

rclrs/Cargo.toml
luca-della-vedova: I personally don't see any other way to do it, if we want to test clients and services we need some sort of async runtime. I am personally not very familiar with `tokio` so it wouldn't have been my first choice but I saw that it was the [framework of choice](https://github.com/ros2-rust/ros2_rust/blob/0e6939c0247751776be5c9fc95774ffcd9fd897e/examples/minimal_client_service/Cargo.toml#L21) for the services example so I stuck to it for consistency.

mxgrey: I don't see any reason to be shy about test dependencies if they're able to add any amount of value to our ability to have worthwhile tests. Especially a dependency as well vetted as tokio.

I could understand hesitation if it were a dependency on the actual package because that could limit what platforms are able to use rclrs, but test dependencies don't have any negative downstream impacts.

Guelakais:
I need this feature! I am currently working on your parameter implementation and I have already noticed that ros2 does not show me the parameters, but I can adjust them with a launch file, as can be seen in an application example, for which you need a corresponding [publisher, subscriber, and a corresponding launch](https://gitlab.com/ros21923912/simple_ros2_node/-/blob/more_simple_nodes/docs/using_parameters.md?ref_type=heads) file. Overall, I find the integration of ros2 parameters in ros2 rust quite successful and almost as good as with the [generate_parameters](https://github.com/PickNikRobotics/generate_parameter_library) library.

rclrs/src/parameter/service.rs
luca-della-vedova: Changed the behavior and tests to return `PARAMETER_NOT_SET`. https://github.com/ros2-rust/ros2_rust/pull/342/commits/3274f42ce33ae094bd4b2020a03f72dec75d716b.
The same applies for all other services but I'll just bring it up here that the closure still returns an `Option` and the `unwrap_or_default` that results in an empty response is still present.
The failure scenario where an empty array will be returned is when conversion to string for the requested parameter name failed. I admit not really knowing how to handle this, since we can't convert the parameter to a string we also can't safely return a `PARAMETER_NOT_SET` response for it since that requires... returning its name in a string format.
I'm happy to change this if any other suggestion comes up.

luca-della-vedova:
I also addressed the loose end of having an external API that makes it easier to change `use_sim_time` but its changes being actually ignored by making the parameter explicitly read only and adding a TODO if we want to explore changing it at runtime in the future (which is quite a can of worms that I'm happy to push down a bit imho) https://github.com/ros2-rust/ros2_rust/pull/342/commits/00c7951778b4feddbe8329afd07dfbbaebcecbd8

rclrs/src/parameter/service.rs
```diff
@@ -1,89 +1,941 @@
...
-use crate::rcl_bindings::rcl_node_t;
+use super::ParameterMap;
+use crate::parameter::{DeclaredValue, ParameterKind, ParameterStorage};
+use crate::{rmw_request_id_t, Node, RclrsError, Service};
 
+// The variables only exist to keep a strong reference to the services and are technically unused.
+// What is used is the Weak that is stored in the node, and is upgraded when spinning.
 pub struct ParameterService {
+    #[allow(dead_code)]
     describe_parameters_service: Arc<Service<DescribeParameters>>,
```
maspe36, line 16: I'm not really understanding this, shouldn't the executor eventually need handles to these services?

rclrs/src/node/builder.rs
```diff
@@ -308,6 +308,7 @@ impl NodeBuilder {
             parameter,
         });
         node.time_source.attach_node(&node);
+        node.parameter.create_services(&node)?;
```
maspe36, line 311: Can we add a node option to disable the parameter services like rclcpp has? 
https://github.com/ros2/rclcpp/blob/de666d2bf41806d150744d7c5dcd7a07ad3ffcd0/rclcpp/include/rclcpp/node_options.hpp#L221

Useful for projects with large node graphs as these services do add a lot of network traffic that can affect discovery performance
https://community.rti.com/static/documentation/performance/benchmarks/srcDoc/pro/7.3/core/discovery/discovery_performance.html#time-to-complete-discovery-spdp-and-spdp2

maspe36 (commented):
A few very small things, thanks for your patience on the review!

rclrs/src/parameter.rs
```diff
@@ -1,19 +1,21 @@
...
+mod range;
+mod service;
 mod value;
 
 pub(crate) use override_map::*;
+pub use range::*;
+use service::*;
 pub use value::*;
 
-use crate::{call_string_getter_with_rcl_node, rcl_bindings::*, RclrsError};
```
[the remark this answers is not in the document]
luca-della-vedova: Interesting, I merge `main` but fmt did nothing:

```
ros2_rust/rclrs$ cargo fmt
Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel.
```

Then I tried `cargo +nightly fmt` and indeed it did reformat a the imports.
Since this was not caught by CI, should we also make sure to set this change on the CI side as well? It was green so I didn't notice, it might require installing the nightly toolchain though.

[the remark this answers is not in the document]
luca-della-vedova: Anyway https://github.com/ros2-rust/ros2_rust/pull/342/commits/d73d07c597141f2ac7f4198e920e21b8a7fa1eaf

rclrs/src/parameter/service.rs
luca-della-vedova: I'm a bit fuzzy on the details since it's been a few months since I wrote this but I believe that the following happens:

* You call [create_service](https://github.com/ros2-rust/ros2_rust/blob/875bda3855d748e6169457592af198211f5d2af5/rclrs/src/node.rs#L274-L291), it returns a strong reference but holds a weak reference internally.
* The executor's spin_once call creates a new [WaitSet](https://github.com/ros2-rust/ros2_rust/blob/875bda3855d748e6169457592af198211f5d2af5/rclrs/src/executor.rs#L50) that waits and collects ready entities.
* The `new_for_node` wait set function, collects the subscriptions by using the [live_*](https://github.com/ros2-rust/ros2_rust/blob/875bda3855d748e6169457592af198211f5d2af5/rclrs/src/node.rs#L319-L345) functions, that iterate through all Weak objects and try to upgrade them to a strong reference, returning all valid pointers.

The result I believe, is that the caller of `create_service` _must_ hold on to a strong reference otherwise that service will be dropped. However, the node itself will store its own `Weak` reference and the strong reference will actually never be used at all, but we must still be sure to keep the service alive for the necessary time.

rclrs/src/node/builder.rs
maspe36: Yeah this is what other client libraries default to

rclrs/src/parameter/service.rs
maspe36: Ahh, I see! Thanks for the clarification. 

Is this TODO still accurate then? Seems like because we will drop these strong references when the node is dropped that our lifetimes are aligned.
https://github.com/ros2-rust/ros2_rust/blob/875bda3855d748e6169457592af198211f5d2af5/rclrs/src/node.rs#L273

maspe36 (approved):

rclrs/src/parameter/service.rs
luca-della-vedova: I haven't introduced that TODO but for my understanding it would still be accurate.
Since the strong reference is returned by the `create_*` call but the node only holds a weak one, what could actually happen is something like this (pathological case):

```
let mut sub;
{
    let node = create_node([...]);
    sub = node.create_subscription([...]);
    // Node is dropped!
}
// sub still has an owning reference, what now?
```

jhdcs (approved):

maspe36 (approved):

jhdcs:
Thank you very much for all your hard work! I'm glad to see ROS 2 Rust finally get support for this!!!

PR #343, opened 2023-11-07
Version 0.4.0

esteve:

PR #344, opened 2023-11-07
Revert "Version 0.4.0"

esteve:
Reverts ros2-rust/ros2_rust#343


esteve:
Had to revert because rclrs depends on rosgraph_msgs, but it's not vendored, so it can't be published to crates.io

jhdcs (approved):

PR #349, opened 2023-11-11
Add wchar support

maspe36:
Added `wchar` support to `rosidl_runtime_rs`. 

From the [ROS 2 IDL - Interface Definition and Language Mapping](https://design.ros2.org/articles/idl_interface_definition.html) spec

> <h4 id="7414424-wide-char-type">7.4.1.4.4.2.4 Wide Char Type</h4>
> 
> 
> 
> IDL type | Value range
> -- | --
> wchar | A 16-bit wide character
> 

This maps nicely to rust's [u16](https://doc.rust-lang.org/std/primitive.u16.html) type.

You can't actually use `wchar` in a .msg file (see [humble docs for supported .msg file types](https://docs.ros.org/en/humble/Concepts/Basic/About-Interfaces.html#field-types)), but its perfectly valid and supported by [rosidl_parser](https://github.com/ros2/rosidl/blob/iron/rosidl_parser/rosidl_parser/definition.py#L43), which means its valid in a .idl file and we should support it.

Also added an example .idl file that we can generate rust code for.

Even though we can generate code for the .idl file, it does not mean that we can use all of the idl features present in the file. The commented out lines are things we don't have support for that actually break the build.

In another PR I am planning to figure out a better way to write tests besides an example package, probably something similar to `rclrs_tests`.

### DIFF
--- a/rosidl_generator_rs/rosidl_generator_rs/__init__.py
+++ b/rosidl_generator_rs/rosidl_generator_rs/__init__.py
@@ -102,9 +102,9 @@ def generate_rs(generator_arguments_file, typesupport_impls):
         'constant_value_to_rs': constant_value_to_rs,
         'value_to_rs': value_to_rs,
         'convert_camel_case_to_lower_case_underscore':
-            rosidl_pycommon.convert_camel_case_to_lower_case_underscore,
+        rosidl_pycommon.convert_camel_case_to_lower_case_underscore,
         'convert_lower_case_underscore_to_camel_case':
-            convert_lower_case_underscore_to_camel_case,
+        convert_lower_case_underscore_to_camel_case,
         'msg_specs': [],
         'srv_specs': [],
         'package_name': args['package_name'],
@@ -168,7 +168,6 @@ def generate_rs(generator_arguments_file, typesupport_impls):
 
     return 0
 
-
 def get_rs_name(name):
     keywords = [
         # strict keywords
@@ -183,7 +182,6 @@ def get_rs_name(name):
     ]
     # If the field name is a reserved keyword in Rust append an underscore
     return name if not name in keywords else name + '_'
-
 
 def escape_string(s):
     s = s.replace('\\', '\\\\')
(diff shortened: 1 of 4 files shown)


rosidl_generator_rs/rosidl_generator_rs/__init__.py
```diff
@@ -219,6 +213,7 @@ def primitive_value_to_rs(type_, value):
     if type_.type in [
             'byte',
             'char',
+            'wchar',
```
maspe36, line 216: Honestly, not sure whats going on with this function. My local tests worked just fine without this change. I'm planning to go a bit deeper in a following PR

maspe36:
Hm, `rclrs_example_msgs` is failing tests right now. Its one of the autogenerated `ament_lint_common` ones, `flake8_rosidl_generated_py`. 

Looking at the autogenerated code in `ws/build/rclrs_example_msgs/rosidl_generator_py/rclrs_example_msgs/msgs/_my_message.py`

```python3
@wchar_value.setter
def wchar_value(self, value):
    if self._check_fields:
        from collections import UserString 
        assert \
            False
    self._wchar_value = value
```

It seems like its not getting an assert. I'll need to take a look at how `rosidl_parser` has its tests for `MyMessage.idl` setup. Probably missed something.

If anyone has any ideas that would also be appreciated :)

maspe36:
Seems like a problem in [rosidl_generator_py](https://github.com/ros2/rosidl_python/blob/058080fe0eec35c75e9954a5fa90105c9fdcafcf/rosidl_generator_py/resource/_msg.py.em). `rosidl_parser` doesn't actually run linter tests on its input files, so they pass successfully. 

I created a small example package with just this .idl file and it also fails. This seems to have nothing to do with ros2_rust or the way I've enabled wchar support. 

Removed the linter tests from `rclrs_example_msgs`. If that's not okay for now I can pull this .idl file example into my next PR (and only turn off linter tests for a specific idl message pacakge) but the wchar feature will be untested (even in examples) until then.

maspe36:
@esteve @nnmm @jhdcs Can I get some eyes on this?

jhdcs (approved):
My apologies, I misunderstood your last comment to mean that this PR was still in development!

The changes appear to be fine to me. I'm pleasantly surprised it doesn't seem to take a gigantic amount of code to support a new type!

maspe36:
@esteve I believe this is good to merge. Is there something else you'd like me to do beforehand?

esteve:
@maspe36 looks good to me, sorry for the slow response and thanks for the ping

esteve (approved):

PR #350, opened 2023-11-14
[docs] Add <Arc> to node field in RepublisherNode

miferco97:
Add the ```Arc<rclcrs::Node>``` type in the ```RepublisherNode``` in the example of the documentation. Otherwise you get this error: 

image


esteve (approved):
@miferco97 thanks!

PR #356, opened 2023-12-21
Add Arc::clone to the Readme tutorial to fix compile error

marcbone:
This PR fixes a compile error in the tutorial that happened because you cannot move the "node" member out of the republisher


esteve (approved):
@marcbone thanks!

PR #360, opened 2024-01-19
Add <Arc> to publisher field to fix the compile error

ksron:
Add `Arc<>` type to the `publisher` field in `RepublisherNode` to fix the compile error:

image


esteve (approved):
@ksron thanks!

PR #361, opened 2024-02-01
Add default implementation and builder pattern for QoS

luca-della-vedova:
This PR adds a default implementation for `QoSProfile` as well as a builder pattern based on the `rclcpp` API.
I split it into a first commit that should be fully uncontroversial and a second commit that contains the builder API for `Duration` based items, which should also be OK I believe.

I intentionally left out all the methods in the other APIs that just set a profile to the requested structure, since while in C++ people might have to do something like:

```
const auto reliability = QoSReliabilityPolicy::Reliable;
auto profile = QoSProfile::SystemDefault().reliability(reliability);
```

In Rust the API is (imho) not so necessary since they can just use struct initialization, such as:

```
let reliability = QoSReliabilityPolicy::Reliable;
let profile = QoSProfile {
   reliability,
   ..QOS_SYSTEM_DEFAULT
};
```

### DIFF
--- a/rclrs/src/qos.rs
+++ b/rclrs/src/qos.rs
@@ -179,6 +179,12 @@ pub struct QoSProfile {
     pub avoid_ros_namespace_conventions: bool,
 }
 
+impl Default for QoSProfile {
+    fn default() -> Self {
+        QOS_PROFILE_DEFAULT
+    }
+}
+
 impl From<QoSProfile> for rmw_qos_profile_t {
     fn from(qos: QoSProfile) -> Self {
         Self {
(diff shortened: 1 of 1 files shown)


rclrs/src/qos.rs
```diff
@@ -179,6 +179,12 @@ pub struct QoSProfile {
     pub avoid_ros_namespace_conventions: bool,
 }
 
+impl Default for QoSProfile {
+    fn default() -> Self {
```
jhdcs, line 183: Should we make a note here that this is setting the QOS profile to the RCL default? Just to avoid any confusion?

jhdcs (approved):
These changes all look good to me! Only one minor comment/question about adding a comment to the default function, but I don't think it's enough to hold up the PR if we don't want to do it.

maspe36 (approved):

esteve (dismissed):
@luca-della-vedova thanks. I agree with @jhdcs about adding a note about the RCL default, could you add that? Thanks

rclrs/src/qos.rs
luca-della-vedova: Done! https://github.com/ros2-rust/ros2_rust/pull/361/commits/6ac6caa3655a4458739d7d1457c20200bfa627ca

esteve: @luca-della-vedova thanks 🙂

esteve (approved):

PR #362, opened 2024-02-04
Unicode handling in String and WString

nwn:
According to https://design.ros2.org/articles/wide_strings.html, the `string` and `wstring` types must use the UTF-8 and UTF-16 encodings (not necessarily enforced), cannot contain null bytes or words (enforced), and, when bounded, are measured in terms of bytes or words.

Moreover, though the rosidl_runtime_c `U16String` type uses `uint_least16_t`, Rust guarantees the existence of a precise `u16` type, so we should use that instead of `ushort`, which isn't guaranteed to be the same as `uint_least16_t` anyways. (Rust doesn't support platforms where `uint_least16_t != uint16_t`.)

### DIFF
--- a/rosidl_runtime_rs/src/string.rs
+++ b/rosidl_runtime_rs/src/string.rs
@@ -10,7 +10,7 @@ mod serde;
 use crate::sequence::Sequence;
 use crate::traits::SequenceAlloc;
 
-/// A zero-terminated string of 8-bit characters.
+/// A zero-terminated UTF-8 string.
 ///
 /// The layout of this type is the same as `rosidl_runtime_c__String`. See the
 /// [`Message`](crate::Message) trait for background information on this topic.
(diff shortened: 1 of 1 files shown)


maspe36 (approved):
Hey sorry for the long delay @nwn, these changes LGTM. 

> Rust doesn't support platforms where uint_least16_t != uint16_t.

Out of curiosity, I was trying to track down this claim, but I'm coming up empty. Do you have a source for this?

nwn:
No worries @maspe36, thanks for taking a look.   

> > Rust doesn't support platforms where `uint_least16_t != uint16_t`.
>
> Out of curiosity, I was trying to track down this claim, but I'm coming up empty. Do you have a source for this?

I'm actually failing to find a direct source for that claim as well, though I thought I had one. I think it can still be inferred from the [definitions](https://en.cppreference.com/w/cpp/types/integer) of `uint16_t` and `uint_least16_t`. If a platform defines `uint16_t`, then by definition `uint_least16_t` must be the same size. So by providing an exact `u16`, Rust would require this.

PR #370, opened 2024-03-12
Allow ros2_rust to be built within a distro's colcon workspace

mxgrey:
I attempted to build the ros2_rust repo directly inside the colcon workspace of a ROS 2 distro and found that it doesn't work because of some `rosidl_generator_rs` behavior that wanted to discover all the RMW implementations in the environment, and that just doesn't quite work with colcon's approach to isolated builds.

However, after investigating it carefully I realized the cmake code that wanted to find all the RMW implementations wasn't actually doing anything helpful. I'm guessing it's dead code that had a specific purpose many years ago, but some redesign since then made it unnecessary.

This PR reworks the dependencies so that there's no need to create a separate colcon workspace to build `ros2_rust`.


jhdcs (approved):
I'm not seeing any major issues with the changes. If anything, the code looks cleaner with all the extra calls removed.

jhdcs (approved):

Issue #369, opened 2024-03-12
Investigate the rcl struct drop order


maspe36:
As discussed in the [March 11th monthly meetup](https://discourse.ros.org/t/ros-2-rust-meeting-march-11th/36523), we have some occasional unit test failures. These failures have seemingly been narrowed down to rclrs_tests and appear to be related to DDS usage. See https://github.com/ros2-rust/ros2_rust/pull/364 for more details.

With that in mind, it is imperative that we take multiple approaches to thoroughly investigate and analyze the root causes of these failures.

For this issue, we agreed that it would be beneficial to look at the drop order for our structs. Since we interface with C, it is possible that we are not properly dropping owned resources (i.e. not in the right order). This is a rather subtle potential problem and will require lots of attention to detail. See [Destructors](https://doc.rust-lang.org/reference/destructors.html) from the rust language reference for more details.

maspe36:
While trying to produce an occasionally observed segfault in our unit tests, I managed to get `graph_tests.rs` to actually panic. This seems like I managed to drop node while calling `NodeBuilder::build`? 

```
>>> [rcutils|error_handling.c:108] rcutils_set_error_state()
This error state is being overwritten:

  'map is not initialized, at /home/sam/ros2_iron/src/ros2/rcutils/src/hash_map.c:535'

with this new error message:

  'Unable to fini type cache for node., at /home/sam/ros2_iron/src/ros2/rcl/rcl/src/rcl/node.c:399'

rcutils_reset_error() should be called after error handling to avoid this.
<<<

>>> [rcutils|error_handling.c:108] rcutils_set_error_state()
This error state is being overwritten:

  'Unable to fini type cache for node., at /home/sam/ros2_iron/src/ros2/rcl/rcl/src/rcl/node.c:399'

with this new error message:

  'node argument is null, at /home/sam/ros2_iron/src/ros2/rmw_fastrtps/rmw_fastrtps_cpp/src/rmw_node.cpp:80'

rcutils_reset_error() should be called after error handling to avoid this.
<<<

>>> [rcutils|error_handling.c:108] rcutils_set_error_state()
This error state is being overwritten:

  'node argument is null, at /home/sam/ros2_iron/src/ros2/rmw_fastrtps/rmw_fastrtps_cpp/src/rmw_node.cpp:80, at /home/sam/ros2_iron/src/ros2/rcl/rcl/src/rcl/node.c:404'

with this new error message:

[...]

Issue #367, opened 2024-03-12
Check if our CI configuration is potentially causing our intermittent failures


maspe36:
As discussed in the [March 11th monthly meetup](https://discourse.ros.org/t/ros-2-rust-meeting-march-11th/36523), we have some occasional unit test failures. These failures have seemingly been narrowed down to `rclrs_tests` and appear to be related to DDS usage. See https://github.com/ros2-rust/ros2_rust/pull/364 for more details.

With that in mind, it is imperative that we take multiple approaches to thoroughly investigate and analyze the root causes of these failures.

For this issue, we agreed that it would be beneficial to do some research on using shared memory in a unit test that runs in CI. We assume that shared memory requires some form of process isolation, so presumably someone else (another client library) has faced and solved these issues. 

Note from RTI on this https://community.rti.com/forum-topic/process-isolation-when-testing-dds-code

maspe36:
I didn't get around to digging too deep here, but I have been unable to reproduce the intermittent failures locally since #372. Closing for now.

Issue #366, opened 2024-03-12
Attempt to reproduce intermittent unit test failures with a different DDS vendor


maspe36:
As discussed in the [March 11th monthly meetup](https://discourse.ros.org/t/ros-2-rust-meeting-march-11th/36523), we have some occasional unit test failures. These failures have seemingly been narrowed down to `rclrs_tests` and appear to be related to DDS usage. See https://github.com/ros2-rust/ros2_rust/pull/364 for more details.

With that in mind, it is imperative that we take multiple approaches to thoroughly investigate and analyze the root causes of these failures.

For this issue, we agreed that it would be beneficial to see if we can reproduce this issue with another DDS vendor. It is not out of the question that our usage is manifesting a bug in a specific DDS implementation. Lets try as many tier 1 DDS vendors as we can get our hands on.

maspe36:
Ever since https://github.com/ros2-rust/ros2_rust/pull/372 got merged, I haven't been seeing the intermittent failures in graph_tests with Eclipse Cyclone, FastDDS, or RTI Connext. 

Going to close this for now.

Issue #365, opened 2024-03-12
Investigate Release Builds with Debug Symbols under GDB for potential issues with rclrs_tests usage of DDS


maspe36:
As discussed in the [March 11th monthly meetup](https://discourse.ros.org/t/ros-2-rust-meeting-march-11th/36523), we have some occasional unit test failures. These failures have seemingly been narrowed down to `rclrs_tests` and appear to be related to DDS usage. See https://github.com/ros2-rust/ros2_rust/pull/364 for more details.

With that in mind, it is imperative that we take multiple approaches to thoroughly investigate and analyze the root causes of these failures.

For this issue, we agreed that it would be beneficial to attach GDB to some release builds that contain debug symbols, in a fresh ros2 workspace. As I understand it, the hope was we could get a nicer stack trace for our intermittent failures than what we saw in #364.

maspe36:
@mxgrey I'm not sure how deep you dug here, but since #372 I haven't been able to reproduce this locally. 

Going to close this for now.

mxgrey:
I actually wasn't able to get gdb to work correctly on any test cases because the rapid creation and destruction of threads was leaving gdb without a stack to latch onto.

But it's reassuring that you're no longer reproducing it, and it's a good indication that this was a drop order issue on our end all along.

I'm skeptical that we're totally out of the woods yet. There may be potential drop orders that just don't happen to show up in our tests. So I'll still plan on auditing the rcl handlers more closely.

PR #372, opened 2024-03-18
Move rcl structs to end of Node declaration

luca-della-vedova:
Cherry-picked from #342, should help towards #369.

CC @maspe36

### DIFF
--- a/rclrs/src/node.rs
+++ b/rclrs/src/node.rs
@@ -65,14 +65,17 @@ unsafe impl Send for rcl_node_t {}
 /// [3]: crate::NodeBuilder::new
 /// [4]: crate::NodeBuilder::namespace
 pub struct Node {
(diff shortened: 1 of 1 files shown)


jhdcs (approved):
Simple change, and good catch! Hopefully this adds a bit of stability while we continue analyzing our current drop issues...

Thank you very much!

maspe36:
I think with this we can close out https://github.com/ros2-rust/ros2_rust/issues/369, https://github.com/ros2-rust/ros2_rust/issues/366, https://github.com/ros2-rust/ros2_rust/issues/365, and https://github.com/ros2-rust/ros2_rust/issues/367. 

If we see an intermittent failure again we can create more appropriate issues. Thanks for your work @luca-della-vedova!

Issue #334, opened 2023-10-23
Fix documentation now that rclrs is on crates.io


Carter12s:
Was doing a pass through the documentation to familiarize myself with the crate. Found multiple "*Note: once `rclrs` is published on crates.io" blurbs in docs/building.md

I do believe rclrs is now on crates.io and this documentation should be updated. Not familiar enough with the crate at the moment to make this changes, so flagging this issue, but I intend to take a stab at it myself later.

esteve:
@Carter12s thanks for filing this ticket. I believe #379 should address this issue, could you have a look at it when you have a moment? Thanks.

PR #379, opened 2024-03-20
Remove references to rclrs not being on crates.io

esteve:
Fixes https://github.com/ros2-rust/ros2_rust/issues/334

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -19,8 +19,6 @@ mkdir src
 git clone https://github.com/ros2-rust/ros2_rust.git src/ros2_rust
 ```
 
-*Note: Once `rclrs` is published on crates.io, it's not technically needed anymore to clone the `ros2_rust` repo, and this section will be modified to reflect that.*
-
 
 ## Environment setup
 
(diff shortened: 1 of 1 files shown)


maspe36 (approved):
So it should be possible to build an rclrs node without having the entirety of ros2_rust in your workspace or an underlay?

esteve:
@maspe36 I'm afraid I'm a bit lost here, is your comment about https://github.com/ros2-rust/ros2_rust/pull/380 ?

maspe36:
I am specifically referring to this removed line
> *Note: Once `rclrs` is published on crates.io, it's not technically needed anymore to clone the `ros2_rust` repo, and this section will be modified to reflect that.*

It sounds like once I run these commands,

```
cargo install --debug cargo-ament-build 
pip install git+https://github.com/colcon/colcon-cargo.git
pip install git+https://github.com/colcon/colcon-ros-cargo.git
```

I could write a node in rust that pulls rclrs from crates.io when I `colcon build`. Is that understanding correct?

mxgrey:
I'd love to be wrong about this but I think for now you'd at least need to have `rosild_generator_rs` in your colcon workspace with the message packages that you need. If we can get that package on the build farm then I think the usage you're describing would work.

Issue #371, opened 2024-03-14
Failed to build depends on `builtin_interfaces`


sprhawk:
https://github.com/ros2-rust/ros2_rust/blob/27bc35584badf0406927b5d6c88625de2ef52580/rclrs/Cargo.toml#L29

I got failure building on `builtin_interfaces`. I [removed it ](https://github.com/iotpi/ros2_rust/commit/1399660123fac828cc82c1fd8fc5ee7408fbd2cd) then it is built successfully.

I'm not sure why it is a dependency of rclrs.

sprhawk:
I read the [building.md](https://github.com/ros2-rust/ros2_rust/blob/main/docs/building.md) again

maspe36:
> I got failure building on builtin_interfaces.

Can you share this failure and maybe your workspace folder structure? I have not seen this error myself

sprhawk:
> > I got failure building on builtin_interfaces.
> 
> Can you share this failure and maybe your workspace folder structure? I have not seen this error myself

I built whole ROS2 myself in Debian Bookworm

I tried later, more libs like `libstatistics_collector` didn't get rust subdir built.

mxgrey:
The vendored version of `builtin_interfaces` that you're using in your diff was a workaround from early in the rclrs development that shouldn't be needed anymore.

It sounds like you're trying to build the `ros2_rust` repo in a single workspace with all the other critical packages of ROS 2. If that's the case you'll need #370 which I opened recently to make that possible. I recommend checking out that branch and doing a clean build (run `$ rm -rf build install` before building).

sprhawk:
> The vendored version of `builtin_interfaces` that you're using in your diff was a workaround from early in the development of rclrs that shouldn't be needed anymore.
> 
> It sounds like you're trying to build the `ros2_rust` repo in a single workspace with all the other critical packages of ROS 2. If that's the case you'll need #370 which I opened recently to make that possible. I recommend checking out that branch and doing a clean build (run `$ rm -rf build install` before building).

Thanks I tried your branch but not clean all build and install, I'll try it again.

maspe36:
I am not seeing the original problem, but the second problem you describe,

> I tried later, more libs like libstatistics_collector didn't get rust subdir built.

I do see. I don't always see it, but when it pops up in my workspace I need to do a clean build to resolve it. Next time my workspace is borked I'll take a closer look.

mxgrey:
I think once #380 is merged we can close this issue.

[...]

PR #380, opened 2024-03-20
Declare rust_packages only when installing Rust IDL bindings

mxgrey:
As discussed [here](https://github.com/ros2-rust/ros2_rust/issues/371#issuecomment-2008696967) there was a strange error that made it impossible to use ros2_rust in an underlay workspace. Packages that have no association to Rust were declaring themselves in the `rust_packages` ament index. This was causing downstream colcon workspaces to redirect their `.cargo/config.toml` file towards non-existent directories. When cargo is directed towards non-existent directories, it considers the situation a build error and exits immediately.

After some challenging investigation, I narrowed the problem down to this: Packages that contain **any** message definitions were invoking the `rosidl_generator_rs`'s extension to the ROS IDL generation pipeline. That extension included the line `ament_index_register_resource("rust_packages")`. For a normal message package that would be fine, but some packages have message definitions that are purely internal to the package for testing purposes and not installed (using the `SKIP_INSTALL` option of `rosidl_generate_interfaces`). Since the messages are not installed in these cases, we also do not install the Rust bindings, which makes perfect sense. However, if we declare the package under the `rust_packages` index, then colcon-ros-cargo will expect a Rust package even though one does not exist.

This PR solves the problem in an extremely simple way: The `rosidl_generator_rs` extension will only declare a package into the `rust_packages` index if the message package is being installed. I've done an A/B test on this change and confirmed that building an rclrs node in a downstream workspace is impossible without this, but works perfectly fine with the changes in this PR.

### DIFF
--- a/rosidl_generator_rs/cmake/rosidl_generator_rs_generate_interfaces.cmake
+++ b/rosidl_generator_rs/cmake/rosidl_generator_rs_generate_interfaces.cmake
@@ -115,9 +115,6 @@ file(MAKE_DIRECTORY "${_output_path}")
 
 set(_target_suffix "__rs")
 
-ament_index_register_resource("rust_packages")
-
-
 # needed to avoid multiple calls to the Rust generator trick copied from
 # https://github.com/ros2/rosidl/blob/master/rosidl_generator_py/cmake/rosidl_generator_py_generate_interfaces.cmake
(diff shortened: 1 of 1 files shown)


esteve (approved):
@mxgrey thanks!

Issue #368, opened 2024-03-12
Move rclrs_test into rclrs


maspe36:
As @esteve explained in our [March 11th monthly meeting](https://discourse.ros.org/t/ros-2-rust-meeting-march-11th/36523), `rclrs_tests` does not need to be a separate package. It was originally created as a separate package before there was a pattern of vendoring message packages. 

Now that we are able to do this, there is no merit to `rclrs_tests` being its own package, the contents of this package are intended to be unit tests for `rclrs`.

PR #383, opened 2024-03-24
Move the tests in the `rclrs_tests` crate into `rclrs`.

maspe36:
Addresses https://github.com/ros2-rust/ros2_rust/issues/368

`rclrs_tests` does not need to be a separate crate. According to Esteve from the March 11th meeting, it was originally created as a separate crate before there was a pattern of vendoring message packages.

I looked at each test and moved them to their appropriate source file. For the helpers, I followed a test organization pattern I've seen in [axum](https://github.com/tokio-rs/axum/tree/main/axum/src/test_helpers).


jhdcs (approved):
The changes look good to me! I especially like the addition of the test helpers. Makes things a bit more organized / less boilerplate.

PR #384, opened 2024-03-30
Moved the `rclrs_example_msgs` package to the examples folder. This package was only ever used by examples.

maspe36:
Continuing to clean up the root level folders. This space comes at a premium and we should try to reduce the amount of noise.


esteve (approved):
@maspe36 thanks!

esteve:
@maspe36 what do you think of eventually replacing this package with upstream's https://github.com/ros2/test_interface_files/tree/rolling ? Maybe even vendorize it. Both packages seem quite similar and it'd mean not maintaining our own package.

maspe36:
We could even take it a step further and move the examples out of the repo. Long term if we can get both `rosidl_generator_rs` and `rosidl_runtime_rs`upstreamed into [rosidl](https://github.com/ros2/rosidl), the only thing left in this repo would be the examples and `rclrs`. 

rclcpp and rclpy examples are already in a different [repo](https://github.com/ros2/examples), and if we move our examples into a different repo, this would leave us with just `rclrs`, which mirrors the structure of say [rclcpp](https://github.com/ros2/rclcpp).

esteve:
@maspe36 we don't need to move `rosidl_generator_rs` and `rosidl_runtime_rs` to another organization on Github for them to be included in the buildfarm. However, moving them to their own repositories within the `ros2-rust` would enable us to release them independently. We should probably do that now anyway.

PR #387, opened 2024-04-05
Refactoring Of Dependency usage

Guelakais:
I just tidied up the imports a little. I have also built and tested everything with the changes. No problems.


mxgrey:
I personally prefer the style that you've migrated to in this PR, but I think this kind of change needs to be discussed in the working group before moving forward. I'll bring this up at the next meeting which is this coming Monday.

If the working group approves of this style, then I think this PR should be updated to add a `.rustfmt.toml` file to the root directory with a [`imports_granularity = Crate`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#crate) setting. That way this style will be enforced by the CI and we won't have to periodically clean up style entropy down the road.

PR #388, opened 2024-04-09
Fix Mistakes in "Refactoring Of Dependency usage #387" (#387-fixup)

Guelakais:
* This commit fixes mistakes introduced in the previous "Refactoring Of Dependency usage" pull request (#387).
* I apologize for any inconvenience caused.
* The main addition remains the `.rustfmt.toml` file, which is now correctly included in all cargo projects. This enables streamlined code formatting using `cargo-fmt` inside a cargo project.
The Changes inside the `.rustfmt.toml` are:
```
imports_granularity = "Crate"
array_width = 60
attr_fn_like_width = 70
brace_style = "SameLineWhere"
combine_control_expr = false
comment_width = 80
wrap_comments = true
```


Guelakais:
By the way, watching the test process is quite exciting. How can you do the same thing that your test process does locally?

mxgrey:
Sorry to be picky, but I have a few concerns to consider:
* `rustfmt` will search recursively through parent directories to find a `.rustfmt.toml` file, so we don't need to add the file to each package. We can just have one `.rustfmt.toml` file at the root of the repository. I just checked locally that this works as expected. We should prefer one style configuration file for the whole project so that if there is ever a decision to modify the style in the future, we can do that easily.
* In the working group there was a consensus on merging imports, but that was the only style change discussed. I don't think it's a good idea to introduce other style changes without checking on how the group feels. It will be a month until the next meeting, but we could drop a message in the [matrix chat](https://matrix.to/#/!AiYXdTWExqamTdatBl:matrix.org?via=matrix.org&via=heyquark.com&via=c-base.org).
* After testing locally, I discovered that almost all of the style settings are currently `unstable`, which means they require the nightly channel in order for `rustfmt` to actually use them. This means we won't actually be able to enforce these style settings through CI for now, which is disappointing. There's nothing we can do about this point, I'm just stating an observation.

esteve:
Superseded by #389

PR #389, opened 2024-04-09
Fixup of #388

Guelakais:


esteve:
@Guelakais thanks for following up on our feedback. Could you in the future address the feedback on the same PR that you submitted instead of creating new ones? It's the GitHub workflow for tracking changes and feedback, otherwise it'll be much more difficult to track feedback. Thanks.

Guelakais:
I had accidentally applied the changes to my own main-branch and found it easier to delete the entire fork and set it up again than to remove commits. I will try to do this better in the future.

esteve (approved):
I somewhat disagree with @mxgrey about having a `.rustfmt.toml` file at the repository base (https://github.com/ros2-rust/ros2_rust/pull/388#issuecomment-2044698809). For now it's ok, but we'll eventually move out some of the projects to their own repository so we can release them independently, so we'll need a `.rustfmt.toml` file in each one.

In any case, I'm approving these changes, thanks @Guelakais

mxgrey:
@esteve Wouldn't it make more sense to put the file once in each root directory of each ros2-rust related repo? Putting it multiple times in the same repo just means we need to remember more places hidden within each repo to update if the target style ever changes.

Also I think the pieces we're thinking about moving out into their own repo(s) so far are Python packages which wouldn't use the `.rustfmt.toml` file anyway.

esteve:
@mxgrey yeah, that makes sense, we can just add a `.rustfmt.toml` file at the root when we create the new repos.

esteve:
@Guelakais thanks!

Issue #385, opened 2024-03-31
`self.node.get_clock().now().to_ros_msg().unwrap()` doesn't work anymore


Guelakais:
```
Some errors have detailed explanations: E0433, E0560.
For more information about an error, try `rustc --explain E0433`.
error: could not compile `lanelet_rs` (bin "lanelet_publisher") due to 3 previous errors; 1 warning emitted
error[E0308]: mismatched types
   --> src/lanelet_subscriber.rs:73:28
    |
73  |                     stamp: self.node.get_clock().now().to_ros_msg().unwrap()
    |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `builtin_interfaces::msg::Time`, found `Time`
    |
    = note: `Time` and `builtin_interfaces::msg::Time` have similar names, but are actually distinct types
note: `Time` is defined in crate `rclrs`
   --> /ros_ws/install/rclrs/share/rclrs/rust/src/vendor/builtin_interfaces/msg.rs:223:1
    |
223 | pub struct Time {
    | ^^^^^^^^^^^^^^^
note: `builtin_interfaces::msg::Time` is defined in crate `builtin_interfaces`
   --> /ros_ws/install/builtin_interfaces/share/builtin_interfaces/rust/src/msg.rs:185:1
    |
185 | pub struct Time {
    | ^^^^^^^^^^^^^^^

error[E0599]: no method named `sec` found for struct `Time` in the current scope
  --> src/lanelet_path_cross_product.rs:78:34
   |
78 |                         sec: now.sec(),
   |                                  ^^^ method not found in `Time`

error[E0599]: no method named `nanosec` found for struct `Time` in the current scope
  --> src/lanelet_path_cross_product.rs:79:38
```
How do you imagine how the timestamp should be filled properly and why does this error first occur today? What have you done?

Guelakais:
Ok, I have a workaround for now. So you can fill the timestamp of a ros2 header correctly:
```
Time {
          sec: (self.node.get_clock().now().nsec /10_i64.pow(9)) as i32,
          nanosec: (self.node.get_clock().now().nsec %10_i64.pow(9)) as u32,
          }
```
[...]

PR #392, opened 2024-04-18
Time to time msg bug

Guelakais:
Mentioning (#385)
So what I basically tried was to adapt the rosidl_generator_rs so that it generates publishable rust crates from ros2 message types, as you can see [here](https://github.com/Guelakais/ros2_rust/tree/rosidl_generator_advancement). that worked reasonably well and since the build worked much better than before, I thought I had solved the problem.

I also read up on the licences of these guys beforehand, so I thought I was on the safe side in this case. Unfortunately, some of these types cannot be deleted from crates.io without further ado. You have to yank them, whereby crates.io first checks whether this dependency is used anywhere. This check is currently running.


esteve:
@Guelakais thanks for working on this. Could you update the description and the title to reflect what these changes do / fix? We follow this convention for any PR that is submit, it helps us track issues and PRs better. Thanks.

esteve:
I've noticed you've published a bunch of crates on crates.io that contain the messages, it'd have been nice to notify us beforehand, there's a reason we haven't published them ourselves. This can cause breakages in existing code and conflicting dependencies, please delete them from crates.io

Can you add any of us as maintainers on crates.io so that we can fix this situation if you can't? Thanks

esteve:
Don't worry, I know the intention was good. Let me see if it's possible to delete crates from crates.io, in the meantime, could you add me as a maintainer?

esteve:
crates.io has the policy of not removing crates, they can only be yanked, but the crates are going to be there forever. We'll need to rethink how messages are included during the build to avoid crates.io potentially interfering in the future.

esteve:
@Guelakais can you add me as a maintainer to the other crates on crates.io? You've only added me to action_msgs, geometry_msgs and lanelet_msgs. I'm going to update their descriptions so that people don't ask us about them.

PR #393, opened 2024-04-19
Rosidl generator advancement

Guelakais:
As I have already announced in another [pull request](https://github.com/ros2-rust/ros2_rust/pull/392), I have been working with your package generator and have upgraded it so that it can generate almost publishable packages. The only thing you have to change manually are the respective version numbers of the packages in the Cargo.toml files. Automating this is also not trivial, which is why I haven't included it yet. In the end it's up to you. Personally, I saw publishing the packages on crates.io as the next logical step if you don't want to go the same way as [r2r](https://github.com/sequenceplanner/r2r), for example, which delivers all possible interface dependencies directly with the main crate. This is also a possibility, even if it would of course bloat ros2 rust. I can definitely understand your point of view. Simply loading the entire ros2 type system onto crates can definitely lead to problems, especially if anyone does it.
At the same time, I didn't just beef up the generator, I streamlined it altogether. As far as I know, the most blatant changes to the part only affected 2 files and 4 lines of code in the end.


esteve (changes requested):
@Guelakais fix the permissions on the files, you've added the exectuable bit to all of them and it's not easy not see what the changes are.

esteve:
@Guelakais we don't want to publish messages to crates.io, anyone could hijack the packages and introduce bugs via the dependencies, crates.io does not control who can or cannot publish a crate. Moreover, a message package can reference different ROS distributions (with different fields, for example), that's why publishing messages to crates.io is problematic, because there's only one distribution available there.

Issue #395, opened 2024-04-29
Crate imports_granularity requires nightly


maspe36:
This was identified by @luca-della-vedova [here](https://github.com/ros2-rust/ros2_rust/pull/342/files#r1582706665). 

Currently, our CI doesn't actually check that our imports are formatted with crate level import granularity. Additionally, this functionality is an unstable feature (https://github.com/rust-lang/rustfmt/issues/4991) so it's only supported in nightly. 

Running locally without a nightly toolchain
```
ros2_rust/rclrs$ cargo fmt
Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel.
```

The way I see it, we have two options.
1. We manually check imports and try to catch this in PRs
2. We start using a nightly toolchain in our CI. We can mitigate some of the churn for nightly here by selecting a version ourselves.

Guelakais:
Isn't this something that should be mentioned directly in the readmy or .rustfmt.toml as a comment? The Cargo manifest format allows comments starting with `#`. Isn't this something that should be mentioned directly in the readmy or .rustfmt.toml as a comment? The Cargo manifest format allows comments starting with `#`. I could do a PR to adapt the README.md.

esteve:
@maspe36 I prefer to only target stable for builds, however we could target nightly for formatting, but I wonder how much of a moving target it is.

PR #396, opened 2024-04-30
Use nightly for style check

mxgrey:
Related to https://github.com/ros2-rust/ros2_rust/issues/395

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,2 +1,3 @@
+# This requires the nightly release of Rust, so when formating the repo, use
+# $ cargo +nightly fmt
 imports_granularity = "Crate"
-


maspe36 (approved):
Can you add a comment in the rustfmt file mentioning that it requires nightly? 

Or perhaps we should have a tracking issue like [this](https://github.com/LibertyDSNP/frequency/issues/1895)?

esteve (dismissed):
@mxgrey thanks! I agree with @maspe36 , a comment about style checking requiring nightly would be nice

esteve (approved):
@mxgrey LGTM, thanks!

mxgrey:
Comment added in 0aaf01d

I think opening an issue is fine, but realistically I'm not expecting this situation to change in the foreseeable future. It seems that a surprising number of rustfmt settings have been stuck on nightly for many years.

maspe36 (approved):

PR #403, opened 2024-05-17
Compile on targets where c_char ≠ i8

AdminXVII:
Fix function which took a i8 where c_char is correct, which fixes compilation where those two types aren't equivalent (ex: aarch64)


esteve (approved):
@AdminXVII LGTM, thanks!

Issue #391, opened 2024-04-15
QoS for Clients and Services


samouwow:
The API for creating new clients and services ([create_client()](https://docs.rs/rclrs/latest/rclrs/struct.Node.html#method.create_client) and [create_service()](https://docs.rs/rclrs/latest/rclrs/struct.Node.html#method.create_service)) doesn't allow for specifying non-default QoS.

From what I can tell, adding the ability to specify QoS should be as simple as passing a QosProfile structure through to the client_options / service_options parameter of the rcl_init functions.

I'm happy to open a PR to add this feature but before I do, I wanted to ask:

1. Is there a reason create_client() and create_service() don't currently allow for a QoS profile?
2. It seems the cleanest way to add this feature will be to modify the existing node function signatures, is there as standard way to make breaking changes like this? (E.g. just go ahead, make a new function and deprecate the old one?)

Guelakais:
I had the same problem. You can simply create your own QoSprofile or modify an existing QosProfile. This then works as follows:
```
fn new(context: &Context) -> Result<Self, RclrsError> {
        let node = create_node(context, "simple_publisher").unwrap();
        let _publisher = node
            .create_publisher("publish_hello", QOS_PROFILE_DEFAULT.keep_last(5))
            .unwrap();
        Ok(Self { node, _publisher })
    }
```
In this example, the QoSprofile que depth was adjusted using the keeplast() method. This is exactly how you can adjust any value of an existing QoSprofile. Unfortunately, I only found the methods in [github](https://github.com/ros2-rust/ros2_rust/blob/main/rclrs/src/qos.rs). If you have any more questions about the Qosprofile implementation of rclrs, just ask me. I am currently working with it.

mxgrey:
I think custom QoS is discouraged for clients and services because there's a significant risk that they'll misbehave if you give them anything other than reliable quality. 

That being said I think it could be reasonable to customize some things like history depth. I'm not sure what the default service setting is for depth, but I can imagine a very busy service could get overwhelmed if that value isn't high enough.

imdexsam:
@Guelakais thanks for the example, are you suggesting I use a standard publisher instead of a client as a workaround for custom QoS?

[...]

PR #404, opened 2024-05-20
QoS for Clients and Services

samouwow:
Addresses #391

### DIFF
--- a/examples/minimal_client_service/src/minimal_client.rs
+++ b/examples/minimal_client_service/src/minimal_client.rs
@@ -7,7 +7,10 @@ fn main() -> Result<(), Error> {
 
     let node = rclrs::create_node(&context, "minimal_client")?;
 
-    let client = node.create_client::<example_interfaces::srv::AddTwoInts>("add_two_ints")?;
+    let client = node.create_client::<example_interfaces::srv::AddTwoInts>(
+        "add_two_ints",
+        rclrs::QOS_PROFILE_SERVICES_DEFAULT,
+    )?;
 
     let request = example_interfaces::srv::AddTwoInts_Request { a: 41, b: 1 };
 
--- a/examples/minimal_client_service/src/minimal_client_async.rs
+++ b/examples/minimal_client_service/src/minimal_client_async.rs
@@ -8,7 +8,10 @@ async fn main() -> Result<(), Error> {
 
     let node = rclrs::create_node(&context, "minimal_client")?;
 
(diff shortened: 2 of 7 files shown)


esteve:
@samouwow thanks for your PR. However, I'm hesitant to accept this change, neither `rclcpp` or `rclpy` accept arbitrary QoS settings for clients and services, among some of the reasons is the issue of interoperability. Publishers and subscribers accept different values for their QoS, but require the other end to have compatible settings (e.g. a publisher's reliability QoS is set as fire and forget, whereas a subscription is created as reliable wouldn't work). Adding this change to the API may break compatilibity with the rest of client libraries.

samouwow:
@esteve thanks for the quick response.

As best I can tell both [rclpy](https://docs.ros2.org/latest/api/rclpy/api/node.html#rclpy.node.Node.create_client) and [rclcpp](https://docs.ros.org/en/rolling/p/rclcpp/generated/classrclcpp_1_1Node.html#_CPPv4I0EN6rclcpp4Node13create_clientEN6rclcpp6ClientI8ServiceTE9SharedPtrERKNSt6stringERKN6rclcpp3QoSEN6rclcpp13CallbackGroup9SharedPtrE) have an API to provide a custom QoS when creating clients and servers, this PR would simply bring rclrs in line with those other libraries.

I will concede both of those client libraries have the QoS as default values, would you feel more comfortable if rclrs had one method for creating clients/servers with QoS and one without?

E.g. `create_client()` and `create_client_qos()`?

esteve:
@samouwow my bad, you're right, I thought neither APIs had that option. Adding a secondary API that takes QoS argument would work, or perhaps just keep create_client as a legacy API and have a Client builder instead would feel more idiomatic

@jhdcs @maspe36 thoughts?

jhdcs:
I like the idea of a client builder. Rust does like its builder patterns, after all.

maspe36:
I also think we should look into creating a client builder instead of proliferating `create_<type>_<extras>` functions.

esteve:
@samouwow thanks for all the work. Would you be interested in adding builders for `Client`s and `Service`s? That way we could add QoS settings and many other options when either structs are created.

mxgrey:
I've opened #427 which introduces an option builder pattern to allow services and clients to override specific fields in the default client/service QoS. It applies the same option builder pattern to subscriptions and publishers, so all the different primitive creation functions are consistent.

I believe that addresses the last round of feedback on this PR.

mxgrey:
We now have the options builder for services and clients merged, which allows users to set their QoS. I believe it would be reasonable to close this PR now.

Issue #406, opened 2024-06-12
Panic occurs when publish sequence types for Rust 1.78


Oscarchoi:
I found that, in **Rust 1.78**, when trying to publish data of sequence types, such as **std_msgs::msg::UInt8MultiArray** or messages utilizing **sensor_msgs::msg::PointCloud2**, a panic occurs with the message: 
`unsafe precondition(s) violated: ptr::write_bytes requires that the destination pointer is aligned and non-null.`

I tested `ros2-rust 0.4.1` on ubuntu 22.04.1 + ros2 humble and checked on ubuntu 20.04.6 + ros2 foxy, too.
The source code I tested and the backtrace message are as follows:
<details>
  <summary>source</summary>

```bash
use rclrs::{Context, Node};

fn main() {
    let context = Context::new([]).expect("failed to create context");

    let node = Node::new(&context, "rust_ros2_node").expect("failed to create node");

    let publisher = node
        .create_publisher::<std_msgs::msg::UInt8MultiArray>("u8_array", rclrs::QOS_PROFILE_DEFAULT)
        .expect("failed to create publisher");

    let layout = std_msgs::msg::MultiArrayLayout {
        dim: [
            std_msgs::msg::MultiArrayDimension {
                label: String::from("x"),
                size: 1,
                stride: 1,
            },
            std_msgs::msg::MultiArrayDimension {
                label: String::from("y"),
                size: 1,
                stride: 1,
            },
            std_msgs::msg::MultiArrayDimension {
                label: String::from("z"),
                size: 1,
                stride: 1,
            },
        ]
        .to_vec(),
        data_offset: 0,
    };

    let msg = std_msgs::msg::UInt8MultiArray {
        layout,
        data: vec![0u8, 1, 2],
    };
    publisher.publish(msg).expect("failed to publish message");

    println!("Published");
}

```

</details>

<details>
  <summary>backtrace</summary>

```bash
$ ~/Workspace/rclrs-tutorial/sequence$ cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/sequence`
thread 'main' panicked at library/core/src/panicking.rs:156:5:
unsafe precondition(s) violated: ptr::write_bytes requires that the destination pointer is aligned and non-null
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/std/src/panicking.rs:645:5
[...]

PR #407, opened 2024-06-20
Fix panic on sequence_init() when size is 0

Oscarchoi:
resolves: #406

In Rust 1.78, a panic occurs when publishing sequence-type messages containing array types, which could be fixed by adding null pointer handling.


jhdcs (approved):
The change appears to add an appropriate guard to prevent a panic. I'm not seeing any immediate issues from the change, since the code looks like it otherwise can handle a null pointer, but a second reviewer would be appreciated.

esteve (approved):
@Oscarchoi LGTM, thanks!

PR #409, opened 2024-06-24
Better error handling and better Comments

Guelakais:
A lot of error handling in the code is solved via if-queries, although you could also do it via match cases match cases are more idiomatic. There are still parts of the code that I don't quite understand. I have tried to comment everything I understand as much as possible.

### DIFF
--- a/rclrs/src/client.rs
+++ b/rclrs/src/client.rs
@@ -314,6 +314,7 @@ impl<T> ClientBase for Client<T>
 where
     T: rosidl_runtime_rs::Service,
 {
+    /// Returns a reference to the `ClientHandle` associated with this client.
     fn handle(&self) -> &ClientHandle {
         &self.handle
     }
--- a/rclrs/src/client.rs
+++ b/rclrs/src/client.rs
@@ -315,7 +315,20 @@ where
     fn handle(&self) -> &ClientHandle {
         &self.handle
     }
-
+    /// Executes the client by taking a response from the underlying `rcl_client_t` instance
+    /// and handling it appropriately.
+    ///
+    /// This method attempts to take a response from the client using `take_response`. If a response
+    /// is successfully taken, it is processed by either calling the associated callback or sending
+    /// the response to the associated future.
+    ///
+    /// If the `take_response` method returns a `ClientTakeFailed` error, it is treated as a
+    /// spurious wakeup and is not considered an error.
+    ///
+    /// # Returns
+    ///
+    /// `Ok(())` if the response was successfully processed or if a spurious wakeup occurred.
+    /// `Err(RclrsError)` if an error occurred while taking or processing the response.
     fn execute(&self) -> Result<(), RclrsError> {
         let (res, req_id) = match self.take_response() {
             Ok((res, req_id)) => (res, req_id),
--- a/rclrs/src/client.rs
+++ b/rclrs/src/client.rs
@@ -41,6 +41,16 @@ impl ClientHandle {
 }
 
 impl Drop for ClientHandle {
+    /// Cleans up the ROS client when the `ClientHandle` instance is dropped.
+    ///
+    /// This implementation finalizes the `rcl_client_t` instance by calling `rcl_client_fini`.
+    /// It also acquires the entity lifecycle mutex to protect against the risk of global variables
+    /// in the rmw implementation being unsafely modified during cleanup.
+    ///
+    /// # Safety
+    ///
+    /// This function calls an unsafe function (`rcl_client_fini`) from the ROS client library.
+    /// The entity lifecycle mutex is locked to ensure thread safety during the cleanup process.
     fn drop(&mut self) {
         let rcl_client = self.rcl_client.get_mut().unwrap();
         let mut rcl_node = self.node_handle.rcl_node.lock().unwrap();
--- a/rclrs/src/client.rs
+++ b/rclrs/src/client.rs
@@ -118,9 +118,6 @@ where
     pub(crate) fn new(node_handle: Arc<NodeHandle>, topic: &str) -> Result<Self, RclrsError>
     // This uses pub(crate) visibility to avoid instantiating this struct outside
     // [`Node::create_client`], see the struct's documentation for the rationale
-    pub(crate) fn new(node_handle: Arc<NodeHandle>, topic: &str) -> Result<Self, RclrsError>
-    // This uses pub(crate) visibility to avoid instantiating this struct outside
-    // [`Node::create_client`], see the struct's documentation for the rationale
     where
         T: rosidl_runtime_rs::Service,
     {
--- a/rclrs/src/client.rs
+++ b/rclrs/src/client.rs
@@ -98,6 +98,26 @@ where
     T: rosidl_runtime_rs::Service,
 {
     /// Creates a new client.
+    ///
+    /// # Arguments
+    ///
+    /// * `node_handle` - An `Arc` reference to the `NodeHandle` associated with this client.
+    /// * `topic` - The name of the topic to which the client will send requests.
+    ///
+    /// # Returns
+    ///
+    /// A `Result` containing the newly created `Client` instance or an `RclrsError` if an error occurred
+    /// during initialization.
+    ///
+    /// # Errors
+    ///
+    /// This function may return an error if:
+    ///
+    /// - The `topic` string contains a null character.
+    /// - The initialization of the underlying `rcl_client_t` instance fails.
+    pub(crate) fn new(node_handle: Arc<NodeHandle>, topic: &str) -> Result<Self, RclrsError>
+    // This uses pub(crate) visibility to avoid instantiating this struct outside
+    // [`Node::create_client`], see the struct's documentation for the rationale
     pub(crate) fn new(node_handle: Arc<NodeHandle>, topic: &str) -> Result<Self, RclrsError>
     // This uses pub(crate) visibility to avoid instantiating this struct outside
     // [`Node::create_client`], see the struct's documentation for the rationale
--- a/rclrs/src/client.rs
+++ b/rclrs/src/client.rs
@@ -30,6 +30,11 @@ pub struct ClientHandle {
 }
 
 impl ClientHandle {
+    /// Locks the `rcl_client_t` instance and returns a mutable reference to it.
+    ///
+    /// # Returns
+    ///
+    /// A `MutexGuard` holding a mutable reference to the `rcl_client_t` instance.
     pub(crate) fn lock(&self) -> MutexGuard<rcl_client_t> {
         self.rcl_client.lock().unwrap()
     }
--- a/rclrs/src/clock.rs
+++ b/rclrs/src/clock.rs
@@ -15,6 +15,7 @@ pub enum ClockType {
 }
 
 impl From<ClockType> for rcl_clock_type_t {
+    /// Converts a `ClockType` enum variant to the corresponding `rcl_clock_type_t` variant.
     fn from(clock_type: ClockType) -> Self {
         match clock_type {
             ClockType::RosTime => rcl_clock_type_t::RCL_ROS_TIME,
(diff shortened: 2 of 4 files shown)


esteve (changes requested):
@Guelakais thanks for your PR. I don't like doing matches on booleans, it actually makes the code less readable. Could you just keep the comments? Thanks.

rosidl_runtime_rs/src/sequence.rs
```diff
@@ -445,7 +466,25 @@ where
 }
 
 // ========================= impl for SequenceIterator =========================
-
+/// Implements the `Clone` trait for the `Sequence` struct.
```
esteve, line 469: This comment doesn't match the code, it implements the `Iterator` trait (see https://doc.rust-lang.org/std/iter/trait.Iterator.html#tymethod.next)

esteve (changes requested):

Guelakais:
> @Guelakais thanks for your PR. I don't like doing matches on booleans, it actually makes the code less readable. Could you just keep the comments? Thanks.

I don't like that either. However, I am more of the opinion that an error-enum should be introduced anyway, which covers error handling better. The matches somehow serve as an aid to my thinking. A suggestion: If I haven't figured it out by Wednesday evening, I'll rebuild the entire PR clean and with comments only. Until then, we can also see whether the other PRs are running better.

Guelakais:
> @Guelakais I'm not going to approve a PR that has code similar to this:
> 
> ```rust
>     fn clone(&self) -> Self {
>         let mut seq = Self::default();
>         match T::sequence_copy(self, &mut seq) {
>             true => seq,
>             _ => panic!("Cloning Sequence failed"),
>         }
>     }
> ```
> 
> it just does not make sense, plus there's no way to change the definition of this function because it's part of the `Clone` trait, you can't change it and expect to continue working. All the functions you've changed are functions defined in their respective traits, so you can't return an error-enum if it's not part of the function signature in the trait.

Ok,

Guelakais:
I have finally learnt how to remove git commits from my history. That makes a lot of things easier.

Guelakais (commented):
I looked at it and edited it. The changes are currently being imported.

rosidl_runtime_rs/src/sequence.rs
esteve: This still does not match the code, it says "Implements the `Clone` trait for the `Sequence` struct.", but the traits that's bein implemented is `Iterator` (see https://github.com/ros2-rust/ros2_rust/pull/409/files#diff-4577ff1d223537bc123ec0d170a0a48ac7621b9a5e0e4fa1fae114da05edff4dR467)

esteve (changes requested):

Guelakais (commented):
I've removed the corresponding commit

esteve:
@Guelakais I'm moving this PR back to draft, let us know when you're done making changes and we'll review it.

Guelakais (commented):
could also be worse. It's ok.

Guelakais:
Yes, no, it's fine. I'm done with that so far.

rclrs/src/client.rs
```diff
@@ -30,12 +30,27 @@ pub struct ClientHandle {
...
         self.rcl_client.lock().unwrap()
     }
 }
 
 impl Drop for ClientHandle {
+    /// Cleans up the ROS client when the `ClientHandle` instance is dropped.
+    ///
+    /// This implementation finalizes the `rcl_client_t` instance by calling `rcl_client_fini`.
+    /// It also acquires the entity lifecycle mutex to protect against the risk of global variables
+    /// in the rmw implementation being unsafely modified during cleanup.
```
maspe36, line 48: This section is redundant with the safety section.

rclrs/src/client.rs
```diff
@@ -83,6 +98,23 @@ where
...
+    ///
+    /// # Errors
+    ///
+    /// This function may return an error if:
+    ///
+    /// - The `topic` string contains a null character.
+    /// - The initialization of the underlying `rcl_client_t` instance fails.
     pub(crate) fn new(node_handle: Arc<NodeHandle>, topic: &str) -> Result<Self, RclrsError>
     // This uses pub(crate) visibility to avoid instantiating this struct outside
     // [`Node::create_client`], see the struct's documentation for the rationale
```
maspe36, line 120: Just noticed this, this is actually just a plain comment — it won't be part of the generated docs.

rclrs/src/client.rs
```diff
@@ -279,10 +311,24 @@ impl<T> ClientBase for Client<T>
...
-
+    /// Executes the client by taking a response from the underlying `rcl_client_t` instance
+    /// and handling it appropriately.
+    ///
+    /// This method attempts to take a response from the client using `take_response`. If a response
+    /// is successfully taken, it is processed by either calling the associated callback or sending
+    /// the response to the associated future.
+    ///
+    /// If the `take_response` method returns a `ClientTakeFailed` error, it is treated as a
+    /// spurious wakeup and is not considered an error.
```
maspe36, line 326: Instead of enumerating each possible result in the docs of many functions, consider just linking to `RclrsError` and in a future PR we can improve the docs for that. 

Here is some inspiration from tokio 
https://doc.rust-lang.org/nightly/std/io/enum.ErrorKind.html

rclrs/src/client.rs
```diff
@@ -279,10 +311,24 @@ impl<T> ClientBase for Client<T>
...
     fn handle(&self) -> &ClientHandle {
         &self.handle
     }
-
+    /// Executes the client by taking a response from the underlying `rcl_client_t` instance
+    /// and handling it appropriately.
+    ///
+    /// This method attempts to take a response from the client using `take_response`. If a response
+    /// is successfully taken, it is processed by either calling the associated callback or sending
+    /// the response to the associated future.
```
maspe36, line 323: These sections could be combined and condensed. The docs for `take_response` already detail quite a bit about whats going on under the hood with the rcl handle. 

This should say something along the lines of... 

Get a response from the rcl handle via take_response and execute the associate callback or future.

rclrs/src/clock.rs
```diff
@@ -45,19 +46,37 @@ impl Clock {
...
+    ///
+    /// A tuple of Clock and ClockSource
     pub fn steady() -> Self {
         Self::make(ClockType::SteadyTime)
     }
 
     /// Creates a new Clock with `ClockType::RosTime` and a matching `ClockSource` that can be used
     /// to update it
+    ///
+    /// # Returns
```
maspe36, line 61: There really is no need for the `# Returns` section. Rustdoc already does a great job showcasing the function signature which will include the return type (with a link to it!).

Sometimes the function signature + name is sufficient documentation and anything we add is just noise.

rclrs/src/clock.rs
```diff
@@ -45,19 +46,37 @@ impl Clock {
...
+    /// A tuple of Clock and ClockSource
     pub fn with_source() -> (Self, ClockSource) {
         let clock = Self::make(ClockType::RosTime);
         let clock_source = ClockSource::new(clock.rcl_clock.clone());
         (clock, clock_source)
     }
 
     /// Creates a new clock of the given `ClockType`.
+    ///
+    /// # Arguments
```
maspe36, line 72: Same thing as the `# Returns` section. Rustdoc already does a good job showcasing the function signature with links to the referenced types. No need to add this manually, its a tad noisy

rosidl_runtime_rs/src/sequence.rs
```diff
@@ -92,6 +92,16 @@ pub struct SequenceIterator<T: SequenceAlloc> {
 // ========================= impl for Sequence =========================
 
 impl<T: SequenceAlloc> Clone for Sequence<T> {
+    /// Creates a deep copy of the sequence.
+    ///
+    /// # Panics
```
maspe36, line 97: This `# Panics` and the `# Safety` sections are good adds! This is exactly what the rust book encourages us to do.
https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html

rosidl_runtime_rs/src/sequence.rs
```diff
@@ -489,11 +537,40 @@ impl Display for SequenceExceedsBoundsError {
         )
     }
 }
-
+/// Implements the `std::error::Error` trait for the `SequenceExceedsBoundsError` struct.
```
maspe36, line 540: Please remove this comment, it duplicates the information in the line of code and doesn't provide any additional content.

rosidl_runtime_rs/src/sequence.rs
```diff
@@ -463,14 +511,14 @@ impl<T: SequenceAlloc> Iterator for SequenceIterator<T> {
         self.idx += 1;
         Some(elem)
     }
-
+    /// returns the remaining length of the sequence
```
maspe36, line 514: This comment doesn't really explain much. Here is some inspiration from the iterator trait
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint

Also, this is not something to be addressed in this PR but I wonder why Sequence doesn't actually implement iterator.

rosidl_runtime_rs/src/sequence.rs
```diff
@@ -463,14 +511,14 @@ impl<T: SequenceAlloc> Iterator for SequenceIterator<T> {
...
     fn size_hint(&self) -> (usize, Option<usize>) {
         let len = (self.seq.size + 1) - self.idx;
         (len, Some(len))
     }
 }
-
 impl<T: SequenceAlloc> ExactSizeIterator for SequenceIterator<T> {
+    /// Returns the length of the sequence as unsigned integer.
```
maspe36, line 521: If you want to document this function, then the comment should expand on more than just the function signature.

Checkout the rustdocs for [std::Vec::len](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.len)

rosidl_runtime_rs/src/sequence.rs
```diff
@@ -489,11 +537,40 @@ impl Display for SequenceExceedsBoundsError {
...
+/// `unsafe` because they call the corresponding extern "C" functions, which may have undefined
+/// behavior if their preconditions are not met. However, the macro assumes that there are no
+/// special preconditions for these functions.
+///
+/// # Arguments
+///
+/// * `$rust_type:ty` - The Rust type for which the `SequenceAlloc` trait should be implemented.
+/// * `$init_func:ident` - The name of the extern "C" function used to initialize a sequence.
+/// * `$fini_func:ident` - The name of the extern "C" function used to finalize a sequence.
+/// * `$copy_func:ident` - The name of the extern "C" function used to copy a sequence.
```
maspe36, line 570: Even with `cargo doc --document-private-items` (or `--all`), this won't be picked up as a docstring by rust doc. So consider just making it a regular comment.

rosidl_runtime_rs/src/sequence.rs
```diff
@@ -505,6 +582,21 @@ macro_rules! impl_sequence_alloc_for_primitive_type {
         }
 
         impl SequenceAlloc for $rust_type {
+            /// Initializes a sequence with a given size.
```
maspe36, line 585: None of the docstrings in here will be picked up by rustdoc. I'd recommend trimming them down (the high level description and safety sections mainly) and turning them into regular comments.

maspe36 (commented):
Please follow the rust doc convention for linking to items by name
https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

---

In general, I think our documentation should follow this high level guideline

```
Short summary of the intention of the function/struct/modules

On success, does this.

If there is a failure, here is what we do.
```

More tokio inspriation
https://docs.rs/tokio/latest/tokio/io/trait.AsyncBufRead.html#tymethod.poll_fill_buf

esteve:
@Guelakais are you still interested in getting this PR merged? Could you please address @maspe36 's feedback? Thanks.

Issue #412, opened 2024-08-02
`impl std::error::Error` for parameter errors


alluring-mushroom:
Hello :smiley: 

I have been testing this excellent project for work.
In using [eyre](https://docs.rs/eyre/latest/eyre/index.html) for much of my error handling, I have found that the errors implemented in [paramaters.rs](https://github.com/ros2-rust/ros2_rust/blob/e485b1c90893bedab32ee9e54ff5e47c4f6bd004/rclrs/src/parameter.rs) do not implement [std::error::Error](https://doc.rust-lang.org/std/error/trait.Error.html), such as `DeclarationError`:
https://github.com/ros2-rust/ros2_rust/blob/e485b1c90893bedab32ee9e54ff5e47c4f6bd004/rclrs/src/parameter.rs#L629

Would you accept a PR containing these?

I assume you do not wish to depend on [thiserror](https://docs.rs/thiserror/latest/thiserror/) as you haven't used it for [RclrsError](https://github.com/ros2-rust/ros2_rust/blob/e485b1c90893bedab32ee9e54ff5e47c4f6bd004/rclrs/src/error.rs#L76). Is there any advice on the [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) impls for any of these? I think many of the comments should suffice, such this one:
https://github.com/ros2-rust/ros2_rust/blob/e485b1c90893bedab32ee9e54ff5e47c4f6bd004/rclrs/src/parameter.rs#L630

Thank you for your time.

jhdcs:
Hello!

Glad you are enjoying `rclrs` so far, and I hope it proves useful for your work! And thank you for the suggestion - yes, I think we probably would accept a PR for implementing `std::error::Error`.

I think that using the the comments would be a good start for the Display implementations for the errors. However, as you're doing so, think about whether that `Display` implementation would help you figure out what might be going wrong with your program. If you think what the comments say aren't enough to help steer an end user (such as yourself or your work colleagues), it may mean that we need to expand what's `Display`ed during an error.

If anything is unclear, or confusing, please feel free to reach out to us! We will do our best to help. And once again, thank you for the suggestion, and offering to help fix it! It means a **lot** to us!

jhdcs:
Discussions about a PR are usually best to do on the PR in question, so I think these questions would be better answered over there.

luca-della-vedova:
This has been fixed by https://github.com/ros2-rust/ros2_rust/pull/413 and can be closed

PR #413, opened 2024-08-14
Add std::error::Error impls to Error enums of `parameter` module

alluring-mushroom:
Adds error impls to the error types used by rclrs parameters.

See #412.

### DIFF
--- a/rclrs/src/parameter.rs
+++ b/rclrs/src/parameter.rs
@@ -624,6 +624,18 @@ pub enum ParameterValueError {
     ReadOnly,
 }
 
+impl std::fmt::Display for ParameterValueError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ParameterValueError::OutOfRange => write!(f, "parameter value was out of the parameter's range"),
+            ParameterValueError::TypeMismatch => write!(f, "parameter was stored in a static type and an operation on a different type was attempted"),
+            ParameterValueError::ReadOnly => write!(f, "a write on a read-only parameter was attempted"),
+        }
+    }
+}
+
+impl std::error::Error for ParameterValueError {}
+
 /// Error that can be generated when doing operations on parameters.
 #[derive(Debug)]
 pub enum DeclarationError {
@@ -643,6 +655,27 @@ pub enum DeclarationError {
     /// An invalid range was provided to a parameter declaration (i.e. lower bound > higher bound).
     InvalidRange,
 }
+
+impl std::fmt::Display for DeclarationError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            DeclarationError::AlreadyDeclared => write!(f, "parameter was already declared, but a new declaration was attempted"),
+            DeclarationError::NoValueAvailable => {
+                write!(f, "parameter was declared as non-optional but no value was available, either through a user specified default, a command-line override, or a previously set value")
+            },
+            DeclarationError::OverrideValueTypeMismatch => {
+                write!(f, "the override value that was provided has the wrong type")
+            },
+            DeclarationError::PriorValueTypeMismatch => {
+                write!(f, "the value that the parameter was already set to has the wrong type")
+            },
+            DeclarationError::InitialValueOutOfRange => write!(f, "the initial value that was selected is out of range"),
+            DeclarationError::InvalidRange => write!(f, "an invalid range was provided to a parameter declaration (i.e. lower bound > higher bound)"),
+        }
+    }
+}
+
+impl std::error::Error for DeclarationError {}
 
 impl<'a> Parameters<'a> {
     /// Tries to read a parameter of the requested type.
--- a/rclrs/src/parameter/value.rs
+++ b/rclrs/src/parameter/value.rs
@@ -376,10 +376,23 @@ impl From<ParameterValue> for RmwParameterValue {
 
 /// An error that occured when trying to convert a parameter from an
 /// `rcl_interfaces::msg::ParameterValue`
+#[derive(Debug)]
 pub enum RmwParameterConversionError {
     /// The parameter type was not valid.
     InvalidParameterType,
 }
+
+impl std::fmt::Display for RmwParameterConversionError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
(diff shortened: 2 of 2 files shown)


jhdcs:
Answering your questions that you have listed here: https://github.com/ros2-rust/ros2_rust/issues/412

"It occured to me that good error messages that are displayed to an end user for parameters should likely include the parameter in question: its name and also possibly the description used in ParameterBuilder. This would require altering the Error struct to store the ParameterBuilder, or at least the relevant information. How do you feel about that?"

I think that adding in the name and description of the parameter would be a good idea. Also, if applicable, reporting the value might be useful. Especially for an "Out Of Range" or "Type Mismatch" error.

"Do you think it is a good idea to include ros2 links in these messages?"

I think that would be good, though we would have to keep an eye on the links to avoid link rot. For example, we would probably have to modify the links for each new version of ROS 2 that's supported. Not necessarily a dealbreaker, but something to keep in mind.

"I'm not sure about the change in rclrs/src/parameter/value.rs , as that code was fairly dense. I wasn't immediatly sure if RmwParameterConversionError would be user facing and would need an Error impl, originally I had only found the Errors in rclrs/src/parameter.rs"

I'll need to do some digging on this one to figure out what's going on. Usually `rmw` stuff isn't user-facing - it deals with the underlying communication methods between nodes. However, it's not fully my area of expertise, so hopefully one of the other maintainers/contributors can chime in to explain what's happening...

luca-della-vedova:
Just for curiosity, what would be the feeling on introducing a dependency to make this a bit less verbose such as [thiserror](https://docs.rs/thiserror/latest/thiserror/)?
But in general thanks for the contribution! I agree with @jhdcs that we could improve the errors to include a value, but that could be tricky when we want to print a generic type and don't want to make the error itself generic, so I understand if it is out of scope of this PR

alluring-mushroom:
I think the best gain for `thiserror` here is that the macro describing the variant is next to the variant, and this keeps more things clear. I will leave that up to everyone here though.

I have struggled a bit with including the value. As you say, it requires the error to be generic, but I have run into another issue. 
Many of the dependencies of the parameter types aren't `Debug`. Thus, many places need the derive. This includes vendored code. I could manually implement `Debug`, but that felt sketchy.
When I work on this again in the next couple of days, I'll push code and add references to this/another comment.

I can see two alternatives to include the data in a message.
1. Include a `String` in the variant, and stringify the name description and value when the error is generated. This means no generics
2. Include the name, description and value in the error, and combine in the display impl. Don't use the Parameter types directly. Requires generics, but can use references. Also allows error handling code to use these (maybe unnecessary?)

I figure I might just finish the PR with nice messages so we have Error impls, and then start a new one with one of these three solutions to improve the error messages.

luca-della-vedova:
Yap fully agree also with the split into another PR! I personally feel the cost of an additional (small) dependency is worth the clarity of less typing and, as you said, keeping the display message in the same place where the error is defined but I'll leave one of the main maintainers to comment on that

jhdcs:
While I usually think it's a good idea to try and limit the number of dependencies for a core library, I think that `thiserror` is a small enough size with a large enough benefit that it might be warranted for inclusion. If it makes contributing quality code easier for people, I think we can certainly take a look at adding it into the project.

How much work would need to be done to transform our current method of creating errors into using `thiserror`? Because if we do include it, we should probably use it library-wide for consistency.

alluring-mushroom:
This is what a single error conversion looks like, for one of the ones I added:
[diff](https://github.com/alluring-mushroom/ros2_rust/blob/412_error_impl_thiserror/rclrs/src/parameter.rs)
```diff
--- a/rclrs/src/parameter.rs
+++ b/rclrs/src/parameter.rs
@@ -18,6 +18,8 @@ use std::{
     sync::{Arc, Mutex, RwLock, Weak},
 };

+use thiserror::Error;
+
 // This module implements the core logic of parameters in rclrs.
 // The implementation is fairly different from the existing ROS 2 client libraries. A detailed
 // explanation of the core differences and why they have been implemented is available at:
@@ -614,28 +616,21 @@ pub struct Parameters<'a> {
 }

 /// Describes errors that can be generated when trying to set a parameter's value.
-#[derive(Debug)]
+#[derive(Error, Debug)]
 pub enum ParameterValueError {
+    #[error("parameter value was out of the parameter's range")]
     /// Parameter value was out of the parameter's range.
     OutOfRange,
     /// Parameter was stored in a static type and an operation on a different type was attempted.
+    #[error(
+        "parameter is stored in a static type and an operation on a different type was attempted"
+    )]
     TypeMismatch,
     /// A write on a read-only parameter was attempted.
+    #[error("a write on a read-only parameter was attempted")]
     ReadOnly,
 }

-impl std::fmt::Display for ParameterValueError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            ParameterValueError::OutOfRange => write!(f, "parameter value was out of the parameter's range"),
-            ParameterValueError::TypeMismatch => write!(f, "parameter was stored in a static type and an operation on a different type was attempted"),
-            ParameterValueError::ReadOnly => write!(f, "a write on a read-only parameter was attempted"),
-        }
-    }
-}
-
-impl std::error::Error for ParameterValueError {}
-
 /// Error that can be generated when doing operations on parameters.
 #[derive(Debug)]
 pub enum DeclarationError {
```

Unfortunately I learned that derive macros cannot generate doc comments, so we will still need those, but at least they are next to each other.

There aren't that many errors that we would need to convert I think:
```bash
rg "impl.*Error for"
rclrs/src/parameter/value.rs
395:impl std::error::Error for RmwParameterConversionError {}

rclrs/src/dynamic_message/error.rs
45:impl Error for DynamicMessageError {

rclrs/src/error.rs
74:impl Error for RclErrorMsg {}
76:impl Error for RclrsError {
317:impl Error for RclReturnCode {}

rclrs/src/parameter.rs
637:impl std::error::Error for ParameterValueError {}
678:impl std::error::Error for DeclarationError {}

rosidl_runtime_rs/src/sequence.rs
493:impl std::error::Error for SequenceExceedsBoundsError {}

rosidl_runtime_rs/src/string.rs
505:impl std::error::Error for StringExceedsBoundsError {}
```

I've taken this out of draft mode, as I figured I would merge this change before submitting another PR for the `thiserror` change, but I can include it all in this one if you like

luca-della-vedova (approved):
Imho this is fairly straightforward and uncontroversial and I can see the argument in favor of leaving the cleanup to a separate PR and just have a consistent implementation first

esteve (approved):
@alluring-mushroom thanks!

Issue #411, opened 2024-07-22
Documentation incorrect regarding BoundedSequence.is_empty()


imdex-matthew:
Documentation incorrectly indicates that you can check a BoundedSequence is empty.

/// // The default bounded sequence is empty
/// assert!(BoundedSequence::<i32, 5>::default().is_empty());

This results in the following error:

thread '...' panicked at library/core/src/panicking.rs:220:5:
unsafe precondition(s) violated: slice::from_raw_parts requires the pointer to be aligned and non-null, and the total size of the slice not to exceed `isize::MAX`

maspe36:
Hey @imdex-matthew, I took a loot at this, and it actually wasn't incorrect documentation. This was a bug in our implementation. 

Now that #416 is merged, this should be fixed

PR #416, opened 2024-09-22
Add in missing nullptr check when calling `std::slice::from_raw_parts`

maspe36:
Seems like back in #407 we added a conditional to not call [write_bytes()](https://doc.rust-lang.org/core/ptr/fn.write_bytes.html) if the internal data buffer of a Sequence was a nullptr. 

It is true that it is unsafe to pass a nullptr to `write_bytes` 
> the pointer must be non-null and properly aligned.

But this breaks our Sequence abstraction because now an empty sequence has a nullptr for its buffer. So later, when we may call [from_raw_parts()](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety), we panic if our Sequence is empty.

> data must be non-null and aligned even for zero-length slices.

So this PR is a small hack for #411, but doesn't fundamentally address the issue. 

---

Taking a deeper look at the BoundedSequence and  Sequence structs, I think we should refactor them to use something like [RawVec](https://doc.rust-lang.org/stable/src/alloc/raw_vec.rs.html#69) (The RawVec impl in the std lib actually uses [Unique](https://doc.rust-lang.org/stable/src/core/ptr/unique.rs.html#36) which is currently unstable :disappointed: ). We know the type that a Sequence or BoundedSequence is constructed with, so there should be no reason our internal buffer isn't non-zero and aligned. 

`Vec::<i32>::default().is_empty()` doesn't panic so `Sequence::<i32>::default().is_empty()` shouldn't panic.

### DIFF
--- a/rosidl_runtime_rs/src/sequence.rs
+++ b/rosidl_runtime_rs/src/sequence.rs
@@ -261,18 +261,26 @@ where
     ///
     /// Equivalent to `&seq[..]`.
     pub fn as_slice(&self) -> &[T] {
-        // SAFETY: self.data points to self.size consecutive, initialized elements and
-        // isn't modified externally.
-        unsafe { std::slice::from_raw_parts(self.data, self.size) }
+        if self.data.is_null() {
+            &[]
+        } else {
+            // SAFETY: self.data is not null and points to self.size consecutive,
+            // initialized elements and isn't modified externally.
+            unsafe { std::slice::from_raw_parts(self.data, self.size) }
+        }
     }
 
     /// Extracts a mutable slice containing the entire sequence.
     ///
     /// Equivalent to `&mut seq[..]`.
     pub fn as_mut_slice(&mut self) -> &mut [T] {
-        // SAFETY: self.data points to self.size consecutive, initialized elements and
-        // isn't modified externally.
-        unsafe { std::slice::from_raw_parts_mut(self.data, self.size) }
+        if self.data.is_null() {
+            &mut []
+        } else {
+            // SAFETY: self.data is not null and points to self.size consecutive,
+            // initialized elements and isn't modified externally.
+            unsafe { std::slice::from_raw_parts_mut(self.data, self.size) }
+        }
     }
 }
 


rosidl_runtime_rs/src/sequence.rs
```diff
@@ -261,18 +261,26 @@ where
...
 
     /// Extracts a mutable slice containing the entire sequence.
     ///
     /// Equivalent to `&mut seq[..]`.
     pub fn as_mut_slice(&mut self) -> &mut [T] {
-        // SAFETY: self.data points to self.size consecutive, initialized elements and
-        // isn't modified externally.
-        unsafe { std::slice::from_raw_parts_mut(self.data, self.size) }
+        if self.data.is_null() {
+            &mut []
```
maspe36, line 278: Actually, now that I'm looking at this, I'm not sure this would work. If someone started pushing to this mutable slice `self.data` would never get updated, so this likely would compile but exhibit incorrect runtime behavior.

rosidl_runtime_rs/src/sequence.rs
```diff
@@ -261,18 +261,26 @@ where
     ///
     /// Equivalent to `&seq[..]`.
     pub fn as_slice(&self) -> &[T] {
-        // SAFETY: self.data points to self.size consecutive, initialized elements and
-        // isn't modified externally.
-        unsafe { std::slice::from_raw_parts(self.data, self.size) }
+        if self.data.is_null() {
+            &[]
```
maspe36, line 265: Taken from this example in the stdlib

https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#ffi-handling-null-pointers

esteve (dismissed):
@maspe36 good catch! Thanks.

rosidl_runtime_rs/src/sequence.rs
jhdcs: Is it possible to write a unit test for this scenario? Just to make sure that it's working, and also as a warning sign in case future changes cause it to break?

maspe36: > Is it possible to write a unit test for this scenario? 

Good point, added a unit test for the base cases. But also, just digging around the slice API, doesn't seem like this is really a concern to mutate the slice as its empty, and the only mutation that can be done will be with the fixed size. 

i.e.
```
let mut slice = Sequence::<i32>::default().as_mut_slice();

// I can't do anything useful with slice now
// slice[0] = 42 will segfault because the len of the slice is zero.
// I actually don't think it matters, its just not really idiomatic rust
```

jhdcs (approved):
I think this should probably work. After thinking about it a bit more, I don't think the mutating slice thing should be an issue since I _think_ the borrow checker should protect against that. Right?

maspe36:
Yup, @jhdcs that was the same realization I came to. I've been writing too much C++ :)

Issue #398, opened 2024-05-13
Upgrade to Rust 1.78


maspe36:
As of Rust 1.78, u128 is now FFI safe. 

Doing this upgrade will help us resolve https://github.com/ros2-rust/ros2_rust/issues/320 once and for all. As part of this upgrade, we should also remove the `#![allow(improper_ctypes)]` and `#![allow(improper_ctypes_definitions)]` lines from our rcl bindings as we no longer need to ignore these lints.

maspe36:
That rust version is supposed to be the minimum supported version of rust that rclrs would work with. However, our [CI does not use this same version](https://github.com/ros2-rust/ros2_rust/blob/main/.github/workflows/rust.yml#L54) so there is a chance that its actually misleading now. 

I think with this upgrade to at least Rust 1.78 we'd want to upgrade rclrs to 1.78 in the cargo.toml as well since we would depend on u128 FFI safe functions.

luca-della-vedova:
Something that we should probably think about is what ROS 2 / Ubuntu version we target and what Rust compiler is shipped with it.
The `rustc` shipped in Noble is [1.75](https://packages.ubuntu.com/noble/rustc), so having anything higher than that as a minimum supported version _might_ put us at risk of not being able to build the packages on a stock Ubuntu distro that only relies on Debian packages and doesn't boostrap rustc from rustup.

mxgrey:
@luca-della-vedova That's a good point. Since Noble is an LTS we should expect that the build farm will be targeting it for quite a while now, and any build farm integration that we hope to accomplish before 26.04 will probably be forced to rely on the version of Rust that shipped with Noble.

It seems the only motivation to go as high as 1.78 is this FFI safety for u128, but even with version 1.78, it's not guaranteed to be safe across all targets. Maybe it's worth putting some time into researching whether there's a more sound way to achieve FFI safety. Especially something that can work with version 1.75.

maspe36:
Yeah, I think tying our minimum Rust version to what is shipped with Noble is reasonable if it helps reduce friction to get into the build farm.

> boostrap rustc from rustup.

FWIW, `rustup` is also [shipped in 24.04](https://packages.ubuntu.com/noble/rustup). All hope is not lost for moving beyond 1.75. Perhaps as the project matures and we are integrated into the build farm we can look at how to unblock Rust upgrades. The hope is still to have a single version of rclrs target many different ROS distros.

mxgrey:
[...]

PR #420, opened 2024-10-03
Use latest stable Rust CI

mxgrey:
This PR will test whether using the latest stable version of Rust will result in cause CI to fail as suspected in https://github.com/ros2-rust/ros2_rust/pull/419#issuecomment-2391502532

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -51,7 +51,7 @@ jobs:
         use-ros2-testing: ${{ matrix.ros_distribution == 'rolling' }}
 
     - name: Setup Rust
-      uses: dtolnay/rust-toolchain@1.74.0
+      uses: dtolnay/rust-toolchain@stable
       with:
         components: clippy, rustfmt
 


mxgrey:
Even after #419 was merged, the CI for this PR still showed some clippy errors. I've updated and repurposed this PR to get the CI green for the latest stable version of Rust.

@esteve I just need to check on how you'd like to resolve [this comment](https://github.com/ros2-rust/ros2_rust/pull/419#issuecomment-2391761725)

> I'd rather have an explicit version of the Rust toolchain there (instead of stable)

versus [this comment](https://github.com/ros2-rust/ros2_rust/pull/419#issuecomment-2392412852)

> we should do that here as well, with the latest stable Rust toolchain to catch any issues with the compiler

I've noticed that [this line](https://github.com/ros2-rust/ros2_rust/blob/ad9667f18151a5f35d02de8ad9889c5e26cefd04/.github/workflows/rust.yml#L9) already has the CI run every day, which is probably more frequently than what's really needed, but the daily runs don't cost us anything so :shrug: . Are you okay with using `@stable` as the PR already does, or do you want me to put in a fixed toolchain version?

For what its worth we do express a minimum supported Rust version in the [Cargo.toml of rclrs](https://github.com/ros2-rust/ros2_rust/blob/ad9667f18151a5f35d02de8ad9889c5e26cefd04/rclrs/Cargo.toml#L9) (which needed to be updated for this PR to pass since we've been using a language feature that stabilized in 1.70). Maybe that's sufficient to express what version we're supporting?

rclrs/Cargo.toml
```diff
@@ -6,7 +6,7 @@ authors = ["Esteve Fernandez <esteve@apache.org>", "Nikolai Morin <nnmmgit@gmail
 edition = "2021"
 license = "Apache-2.0"
 description = "A ROS 2 client library for developing robotics applications in Rust"
-rust-version = "1.63"
+rust-version = "1.70"
```
maspe36, line 9: We actually depend on u128 FFI functions (and suppress warnings currently), so it'd be nice if our minimum was actually 1.78 so we could remove that.

https://github.com/ros2-rust/ros2_rust/issues/398

esteve:
@mxgrey thanks for the changes.

> @esteve I just need to check on how you'd like to resolve https://github.com/ros2-rust/ros2_rust/pull/419#issuecomment-2391761725

I had forgotten that we already run CI periodically and not only per PR. I meant running the periodic workflow with `@stable`, but the per-PR CI with a fixed version. Mainly so that we can update the minimum required version in `rclrs/Cargo.toml` whenever a PR fails.

mxgrey:
It might be confusing if the CI is inconsistent between the PRs and the daily runs. If someone submits a PR to fix the stable CI, we won't get a positive indication that it was successful until the day after the PR is merged.

When I have a ROS project where I want to support multiple ROS distros on one branch, I make a CI matrix that runs all the tests across multiple distros. What if we do that here: Have a CI matrix (for both PRs and daily CI) that runs the tests against the oldest supported Rust version as well as stable?

esteve:
Yeah, you're right, it can be confusing. In that case, what you propose sounds more like what I had in mind, basically, I want us to be able to detect when to update the version in `rclrs/Cargo.toml`

rclrs/Cargo.toml
mxgrey: I've bumped the minimum version to 1.78 because I agree that it's good to declare that the minimum version we support is the one where u128 is ABI safe.

But sadly that's not enough to let us remove the lint because [the lint is still being applied for the foreseeable future](https://github.com/rust-lang/lang-team/issues/255), I guess because it's still not guaranteed to be ABI safe across all build targets.

mxgrey:
@esteve Sadly I tried out the matrix idea and discovered that variables cannot be used in the `uses:` input for GitHub actions [[docs here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#env-context)]. This seems to be a fundamental limitation in how GitHub actions are implemented: They want to be able to resolve all reusable actions before forming any of the context data.

I've reluctantly resorted to just copy/pasting the workflow file with the only difference between the two being `@stable` vs `@1.78`.

It may be possible to improve this situation by creating a workflow template and reusing it between the two files, but [based on the documentation](https://docs.github.com/en/actions/sharing-automations/creating-workflow-templates-for-your-organization#creating-a-workflow-template) the template needs to be written at the organization level in the `.github` repo, so someone will need to create a `.github` repo for `ros2-rust` before we can take that approach.

esteve (approved):
@mxgrey thanks!

PR #421, opened 2024-10-14
Async Execution

mxgrey:
This PR is not quite ready for review but I'm opening it as a draft to facilitate discussion.

This is a working implementation of async execution for rclrs. Here's a high level list of the capabilities this provides: 
* callbacks for subscriptions, services, and clients can all support async functions (as well as still supporting sync function).
* downstream users can create custom executors and users do not need to change their code
  * custom executors can use any async framework of their choosing
  * custom executors can support multi-threading and parallel execution of nodes
* services and clients can now infer what kind of arguments a user's callback has (i.e. do they want the RequestId, ServiceInfo or only the message)
* support for tracking graph change events, especially when a service becomes available for a client

Here's a checklist of remaining TODOs:
- [ ] Create doctests and more thorough documentation for new features
- [ ] Migrate existing examples to the new API
- [ ] Create new examples that showcase the new capabilities

I'll update this PR description in the future with a more complete explanation of what's happening here, but I wanted to open this draft to get the conversation going.


mxgrey:
I think we can just close this since the spun off PRs have diverged enough from this original PR that there's not much point in keeping it open.

For an overview of the final product that I envision for async execution, I recommend looking at #446

PR #422, opened 2024-10-21
Add rosout logging to rclrs

geoff-imdex:
The purpose of this PR is to add logging macros that replicate the logging interface provided by the `rclcpp` and `rclpy` interfaces, while aiming to provide an idiomatic rust interface.

This work is based on the initial work from [here](https://github.com/sequenceplanner/r2r/blob/89cec03d07a1496a225751159cbc7bfb529d9dd1/r2r/src/utils.rs) and [here](https://github.com/mvukov/rules_ros2/pull/371).

Examples of the interfaces provided:
```
log_debug!(&node.name(), "Simple message");
log_debug!(&node.name(), "Simple message {some_variable}");
log_fatal!(&node.name(), "Simple message from {}", node.name());
log_warn!(&node.name(), LogConditions::once(), "Only log this the first time");
log_error!(&node.name(),
           LogConditions::skip_first_throttle(Duration::from_millis(1000)),
           "Noisy error that we expect the first time");

log_info!(&node.name(), LogConditions { occurs: LoggingOccurrence::Always,
                                        publish_interval: Duration::from_millis(1000),
                                        log_if_true: count % 10 == 0, },
          "Manually constructed LogConditions");
```

The macros make use of the `LogConditions` struct which provides the control over the logging, e.g. throttling, occurrence (once, skip first), etc.
The most common control constructs can be instantiated using the provided `builder` functions, e.g.
```
/// Only log the message on the first occurrence
LogConditions::once() 

/// Only log the message at the specified duration
LogConditions::throttle(Duration)

/// Only log the message if the condition is true.
LogConditions::log_if_true(<some expression>)
```

## Known Issues
~~1. The is currently an issue on node closure (as we process the `rcl_node_fini` call) where the "impl" pointer appears to be "null" when we try too fini the rosout publisher (in `rcl_logging_rosout_fini_publisher_for_node`).  The rest of the node closure occurs successfully.  Note: the rosout usage can be disabled with `--ros-args --disable-rosout-logs` and the error will not occur.~~
This issue has been addressed in the latest commit.  I believe that the issue was caused by the `rcl_node` variable in builder.rs going out of scope during the "drop" sequence. The fix is to allocate the `rcl_node` on the heap (using Box).
(As a side note: technically, we may have the same issue with the Context object).
2. We use static variables to facilitate the "once", "skip_first", and throttling functionality, however, I believe it would be preferred 
to use the `lazy_static`.  At this moment, our rust compiler is 1.79 and therefore the lazy_static is not regarded as stable.


esteve:
@geoff-imdex thank you so much for your PR. Given that the code is based on the work done at `r2r` and at `rules_ros2`, it'd be good to have some feedback from the authors of `r2r`, and maybe we could a shared crate with this code so as to not duplicate work. Legally speaking we can copy that code (both `r2r` and `ros2-rust` are license as Apache 2.0), but it's always nice to ask.

@m-dahl @kristoferB is it ok with you guys if we copy this code from `r2r` into `ros2-rust`? We'd of course attribute credit. Moreover, would you be interested in spinning off the logging code of `r2r` into a shared crate that both projects can use?

rclrs/src/node.rs
```diff
@@ -71,10 +71,15 @@ pub struct Node {
...
+/// Note: we capture the rcl_node_t returned from rcl_get_zero_initialized_node()
+/// to guarantee that the node handle exists until we drop the NodeHandle
+/// instance. This addresses an issue where previously the address of the variable
+/// in the builder.rs was being used, and whose lifespan was (just) shorter than the
+/// NodeHandle instance.
 ///
 /// [1]: <https://doc.rust-lang.org/reference/destructors.html>
 pub(crate) struct NodeHandle {
-    pub(crate) rcl_node: Mutex<rcl_node_t>,
+    pub(crate) rcl_node: Mutex<Box<rcl_node_t>>,
```
mxgrey, line 82: Can I check why this was put into a Box? The contents of `rcl_node` are just pointers so there shouldn't be any issues with moving the address of the `rcl_node` (and its parent struct is managed by an Arc, so its address shouldn't change anyway).

mxgrey:
In rclrs there's a [Logger](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/logger.hpp) class that bundles up factors like name of the logger and severity. It also allows hierarchical structures where you can create a Logger which is a child of another Logger and it will inherit settings like a name prefix and severity level (which can be overridden).

In this PR the name and severity seem to be loose values. In fact I don't see where severity gets filtered at all (e.g. we wouldn't want DEBUG severity being printed out unless the user asks for DEBUG level logging).

Would we be interested in having a `Logger` struct similar to rclcpp's? I could draft that up.

I do like the `LogConditions` of this implementation, though. That's a very nice mechanism.

rclrs/src/logging.rs
```diff
@@ -0,0 +1,342 @@
...
+        line_number: line as usize,
+    };
+    let format = CString::new("%s").unwrap();
+    let logger_name = CString::new(logger_name)
+        .expect("Logger name is a valid c style string, e.g. check for extraneous null characters");
+    let message = CString::new(msg)
+        .expect("Valid c style string required, e.g. check for extraneous null characters");
+    let severity = severity.to_native();
+
+    let _guard = LOG_GUARD.lock().unwrap();
```
mxgrey, line 42: Since this is rclrs should use [`ENTITY_LIFECYCLE_MUTEX`](https://github.com/ros2-rust/ros2_rust/blob/f45a66f47dc727e3ccb13037a6c57923af1446c7/rclrs/src/context.rs#L23), which is the mutex that rclrs uses to ensure that global variables within the rcl layer and lower are protected.

rcl logging is notoriously not thread safe, so it's important to make sure we're locking appropriately.

rclrs/src/context.rs
```diff
@@ -142,6 +150,24 @@ impl Context {
...
+            let ret = {
+                let _lifecycle_lock = ENTITY_LIFECYCLE_MUTEX.lock().unwrap();
+                // Note: shadowing the allocator above.  The rcutils_get_default_allocator provides us with mechanisms for allocating and freeing
+                // memory, e.g. calloc/free. As these are function pointers and will be the same each time we call the method, it is safe to
+                // perform this shadowing
+                // Alternate is to call rcl_init_options_get_allocator however, we've freed the allocator above and restructuring the call
+                // sequence is unnecessarily complex
+                let allocator = rcutils_get_default_allocator();
+
+                rcl_logging_configure(&rcl_context.global_arguments, &allocator).ok()
```
mxgrey, line 168: This will get called every time a Context is created, but it's meant to be called only once. It's possible for users to create multiple Contexts in one application (we do this a lot in tests), so we'd be making redundant calls.

Similar to https://github.com/ros2/rclcpp/pull/998 we should guard this from being called multiple times. We also need to make sure that `rcl_logging_fini` is only called after all Contexts are destructed.

This is a tricky problem to resolve in a sound way. I might suggest something like this:
```
struct LoggingConfiguration {
    logging: Mutex<Weak<LoggingLifecycle>>
}

impl LoggingConfiguration {
    fn configure(args: &rcl_arguments_t) -> Result<Arc<LoggingLifecycle>, RclError> {
        static CONFIGURATION: LazyLock<Self> = LazyLock::new(|| Self { logging: Mutex::new(Weak::new()) });
        let mut logging = CONFIGURATION.logging.lock().unwrap();
        if let Some(arc_logging) = logging.upgrade() {
            return Ok(arc_logging);
        }
        let arc_logging = Arc::new(LoggingLifecycle::new(args)?);
        *logging = arc_logging.downgrade();
        Ok(arc_logging);
    }
}

struct LoggingLifecycle;
impl LoggingLifecycle {
    fn new(args: &rcl_arguments_t) -> Result<Self, RclError> {
        let _lifecycle_lock = ENTITY_LIFECYCLE_MUTEX.lock().unwrap();
        let allocator = rcutils_get_default_allocator();
        rcl_logging_configure(args, &allocator).ok()?;
        Self
    }
}

impl Drop for LoggingLifecycle {
    fn drop(&mut self) {
        let _lifecycle_lock = ENTITY_LIFECYCLE_MUTEX.lock().unwrap();
        rcl_logging_fini();
    }
}
```
Then we would add an `Arc<LoggingLifecycle>` to the `ContextHandle`. Once all `ContextHandle`s are gone, the `LoggingLifecycle` will drop and call `rcl_logging_fini`. If a new context is made after that, `rcl_logging_configure` will be called again.

m-dahl:
> @geoff-imdex thank you so much for your PR. Given that the code is based on the work done at `r2r` and at `rules_ros2`, it'd be good to have some feedback from the authors of `r2r`, and maybe we could a shared crate with this code so as to not duplicate work. Legally speaking we can copy that code (both `r2r` and `ros2-rust` are license as Apache 2.0), but it's always nice to ask.
> 
> @m-dahl @kristoferB is it ok with you guys if we copy this code from `r2r` into `ros2-rust`? We'd of course attribute credit. Moreover, would you be interested in spinning off the logging code of `r2r` into a shared crate that both projects can use?

Thanks for the mention, happy to see it used and improved. In the long term I would be in favor of a shared crate based on this work (perhaps it could be a starting point to for more code sharing), but right now I don't have much time to work on ros stuff so for now I think you should just go ahead and do what makes most sense for this crate.

rclrs/src/node.rs
geoff-imdex: Good question - without this change it appears as though the node handle goes out of scope before we get chance to clear the logging publisher (in humble this occurs in the rcl_node_fini() call).  The behaviour without the change is that you will get a segfault on node closure because the object pointed to by the rcl_node_t is no longer valid.  Took a bit of debugging to get to the bottom of it - and honestly, I don't particularly like the change with the extra de-refs.
Looking at Jazzy, I believe that the issue will no longer occur due the logging publisher being managed outside the rcl_node_init/fini calls.
For reference the failure occurs [here](https://github.com/ros2/rcl/blob/6d52f2b54980140476222b420467e7fbd8575dd9/rcl/src/rcl/node.c#L378)

rclrs/src/logging.rs
```diff
@@ -0,0 +1,344 @@
...
+/// e.g. log_debug!, etc, however, it is relatively straight forward to call the macro directly
+/// if you really want to.
+///
+/// # Examples
+///
+/// ```
+/// use rclrs::{log_debug, log_error, log_fatal, log_info, log_warn, LogConditions, LoggingOccurrence};
+/// use std::sync::Mutex;
+/// use std::time::Duration;
+/// use std::env;
```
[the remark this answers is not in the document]
geoff-imdex: Nice!  Thanks for the tip.

rclrs/src/logging.rs
geoff-imdex: Good comment.  The thought had crossed my mind on this one - I thought it more akin to the mutex in rclpp logging, but I'm happy to change it to the `ENTITY_LIFECYCLE_MUTEX`

rclrs/src/node.rs
mxgrey: Thanks for the explanation. I think we do want to keep supporting humble for now, so I'll try to reproduce the seg fault in humble and see if there may be alternative solution.

I'm still pretty lost on how boxing the `rcl_node_t` would affect its lifetime, but if I can reproduce the problem then I should be able to make sense of it.

mxgrey: Oh actually if rcl itself is storing the `rcl_node_t*` value somewhere and using it without consideration for its lifetime then I can see how this would happen. I'll check if that's the case, and maybe we can set things up so that we don't initialize the `rcl_node_t` until it's already inside the `NodeHandle`.

geoff-imdex:
> In rclcpp there's a [Logger](https://github.com/ros2/rclcpp/blob/rolling/rclcpp/include/rclcpp/logger.hpp) class that bundles up parameters like the name of the logger and severity filter. It also allows hierarchical structures where you can create a Logger which is a child of another Logger and it will inherit settings like a name prefix and severity level (which can be overridden).
> 
As you can tell, I avoided this extra complexity with this implementation as I wasn't sure (naivety/ignorance) what benefit/use it would bring and was a little outside the scope of what I needed.

> In this PR the name and severity seem to be loose values. In fact I don't see where severity gets filtered at all (e.g. we wouldn't want DEBUG severity being printed out unless the user asks for DEBUG level logging).
> 
It's not obvious, but the filtering occurs in the rcutils_log function. I'm sure you're across this already, but just to be sure: the rcutils logging code basically has a look-up table that tracks the logger name and what logging level is enabled for it.  It does mean that we do a certain amount of function calls for levels that aren't enabled, however, they're never published.  The rclcpp implementation uses some very smart macro construction to include logs at compile time, but I'm not sure we would (or could?) want to do the same thing.

> Would we be interested in having a `Logger` struct similar to rclcpp's? I could draft that up.
> 
> I do like the `LogConditions` of this implementation, though. That's a very nice mechanism.
Thanks!

rclrs/src/node.rs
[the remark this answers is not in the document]
geoff-imdex: I've replied [above](https://github.com/ros2-rust/ros2_rust/pull/422#discussion_r1822560373) but as a minimal example of how I was able to cause the segfault:
```
use rclrs::{
    log_debug, log_error, log_fatal, log_info, log_warn, LogConditions, LoggingOccurrence, Node,
};
static NODE_NAME: &str = "test_node";

// Minimal test code
use std::env;
use std::thread::sleep;
fn main() -> Result<(), rclrs::RclrsError> {
    log_warn!(NODE_NAME, "Starting");
    let context = rclrs::Context::new(env::args()).unwrap();
    let node = rclrs::Node::new(&context, NODE_NAME).unwrap();
    sleep(Duration::from_secs(1));
 
    log_warn!(NODE_NAME, "Finishing");
    Ok(())
}
```
Note: some of the use statements are unnecessary in the code - I've just hacked up an existing test app.  We also had a similar test in our CI that showed the same behaviour.  
Happy to go in to more detail if needed.

geoff-imdex:
@mxgrey , @maspe36 , really appreciate you taking the time to review the code and provide some helpful feedback.  I'll try and incorporate the suggestions over the next few days or so.
As to the segfault we observed locally, happy to provide more details as needed - would be interesting to note if you still the same behaviour with a minimal test app (create node + context, and shutdown).

mxgrey:
> the rcutils logging code basically has a look-up table that tracks the logger name and what logging level is enabled for it.

I've never dug into how logging works at the rcl layer, so I actually didn't know this detail of the implementation. Thanks for highlighting this.

That actually makes me feel much more strongly that we should have a firm `Logger` struct rather than using `&str`. The current API in the PR would allow rclrs users to pass arbitrary names into the macros, and it sounds like that could lead to misbehavior in the rcl layer. If the macros require a `Logger` object to be passed in then we can enforce that the user is always using a valid logger that rcl will know about.

I have some ideas for how to give rclrs a `Logger` struct that aligns with rclcpp's while also covering the capabilities of `LogCondition`, but implementing the ideas will take about the same amount of time as explaining them so I'll open a PR for that which targets your branch if you don't mind @geoff-imdex .

rclrs/src/node.rs
geoff-imdex: Yep, that's it in a nutshell - the rosout code keeps a record of the node used to create it (this [line](https://github.com/ros2/rcl/blob/6d52f2b54980140476222b420467e7fbd8575dd9/rcl/src/rcl/logging_rosout.c#L189) specifically) and this pointed to the stack variable in NodeHandle ([here](https://github.com/ros2-rust/ros2_rust/blob/f45a66f47dc727e3ccb13037a6c57923af1446c7/rclrs/src/node/builder.rs#L280)).  During the `rcl_node_fini` we call `rcl_logging_rosout_fini_publisher_for_node` which tries to reference the stack pointer, however, this area of memory it points to has now been mangled, i.e. the pointer still points to the same memory location but the values at that spot are garbage.

geoff-imdex:
> That actually makes me feel much more strongly that we should have a firm `Logger` struct rather than using `&str`. The current API in the PR would allow rclrs users to pass arbitrary names into the macros which might not be names of actual loggers that rcl recognizes, and it sounds like that could lead to misbehavior in the rcl layer. If we have the macros require a `Logger` object to be passed in then we can enforce that the user is always using a valid logger that rcl will know about.
> 
Not disagreeing with the suggestion to create a logging struct, but it is worth highlighting that the rcl logging interface explicitly allows passing in arbitrary log names, e.g. the logging macros throughout the rcl code pass in a simple string.  Perhaps a better way to phrase - the logging system supports existing without requiring a node to be associated - in this scenario the rcl layer uses the default log values (INFO level and logging to stdout/stderr).  It is only when we want to publish to `/rosout` that the logging is associated with a node (via the implementation in logging_rosout.c).
To be clear, I think the suggestion to create a logging struct will add a lot of value for the nodes that use the logging interface.

> I have some ideas for how to give rclrs a `Logger` struct that aligns with rclcpp's while also covering the capabilities of `LogCondition`, but implementing the ideas will take about the same amount of time as explaining them so I'll open a PR for that which targets your branch if you don't mind @geoff-imdex .
Please feel free! And thank you!

rclrs/src/context.rs
mxgrey: I took the liberty of opening a PR for this: https://github.com/geoff-imdex/ros2_rust/pull/6

It's _almost_ exactly what I recommended above.

rclrs/src/logging.rs
mxgrey: Taken care of by https://github.com/geoff-imdex/ros2_rust/pull/6

rclrs/src/node.rs
mxgrey: I've gone ahead and put together an alternative fix here: https://github.com/geoff-imdex/ros2_rust/pull/5

mxgrey:
> it is worth highlighting that the rcl logging interface explicitly allows passing in arbitrary log names

That's good to know. In that case I'll tweak the macro so that its first argument can be anything that implements `Borrow<str>`. Then we can implement `Borrow<str>` for `Node` as well as the new `Logger` struct that I'll introduce, and users will be able to pass in any of those things, as well as `&str` and `&String` and basically anything stringy.

mxgrey:
For visibility I'll mention here that I've opened a PR targeting the branch of this one that introduces a `Logger` struct, adds the ability to set severity levels for the loggers, and streamlines the syntax of the logging macros, especially for setting conditions: https://github.com/geoff-imdex/ros2_rust/pull/7

geoff-imdex:
@mxgrey , the iron builds are throwing some linting errors which I'll fix up.  As the changes are based from your fork, my plan is to merge in your PRs (starting with the `unboxed_logging` fix) and then address the linting issues in dev branches from this fork.  Hopefully this process works OK for you, but please let me know if that's not the case and we can work out a way forward.

The logging params PR is really nice.  As a rust newbie there's some really nice techniques in there that I'll be sure to use in the future.

luca-della-vedova:
CI failures look real, specifically using `LazyCell` that was only stabilized in Rust 1.80 while our MSRV is lower

geoff-imdex:
> CI failures look real, specifically using `LazyCell` that was only stabilized in Rust 1.80 while our MSRV is lower

@mxgrey has done some great work to fix the build issues (among many other improvements), unfortunately, I've not had time to review and merge in.  Hoping to get to that shortly.

mxgrey:
The latest and greatest version can be found here: https://github.com/geoff-imdex/ros2_rust/pull/11

The CI is green over there since I've replaced every use of LazyLock with OnceLock which was stabilized in 1.70.

mxgrey:
@esteve @maspe36 @geoff-imdex @luca-della-vedova 

Following up on the discussion in the working group meeting, would there be any preference for or against renaming `interval` to `cooldown` to make the behavior more clear?

The concern was that `interval` could be interpreted as "keep publishing this message with the given period" instead of "don't republish this message until the given period has passed", so the word `cooldown` might help make that more clear.

geoff-imdex:
> @esteve @maspe36 @geoff-imdex @luca-della-vedova
> 
> Following up on the discussion in the working group meeting, would there be any preference for or against renaming `interval` to `cooldown` to make the behavior more clear?
> 
> The concern was that `interval` could be interpreted as "keep publishing this message with the given period" instead of "don't republish this message until the given period has passed", so the word `cooldown` might help make that more clear.

No strong opinions on this from my perspective.  I had thought I'd followed the rcutils macro wording, but looking again I can see they use `throttle` - so potentially a small argument for consistency, but `cooldown` works just as well.

mxgrey:
The reason I moved away from `throttle` is because in my mind the word "throttle" implies we should provide a rate (frequency) but we actually provide a duration (cooldown interval).

That being said it's true that `throttle` is the terms used in ROS 1 and rclcpp, so I don't mind going back to throttle for the sake of consistency with other client libraries. I'd propose a casual vote via reaction emojis on this comment:

:heart: `throttle`
:rocket: `cooldown`

mxgrey:
All the intended features for logging are finished, along with tests. I've also updated the API based on the working group discussion last week.

This PR should be ready for maintainer reviews now.

jhdcs (changes requested):
Just a single typo, otherwise this looks good to me! Thank you so much for this!

luca-della-vedova (commented):
Can you merge `main` into this PR? There are a lot of spurious changes from other PRs that make it a bit difficult to review.
Another high level thing I would like to see is updating the examples to make sure of logging since we have it (i.e. [here](https://github.com/ros2-rust/ros2_rust/blob/066dd7c6e70d515268945ab6190793ef605fd887/examples/rust_pubsub/src/simple_subscriber.rs#L34-L41))
I'll try to give it a more thorough review when the diff is smaller

rclrs/src/logging/logger.rs
```diff
@@ -0,0 +1,101 @@
...
+///
+/// Note that if you are passing the Logger of a Node into one of the logging macros,
+/// then you can choose to simply pass in `&node` instead of `node.logger()`.
+///
+/// [1]: crate::log
+/// [2]: crate::Node::logger
+pub struct Logger {
+    name: Box<str>,
+    c_name: CString,
+}
```
luca-della-vedova, line 27: I believe it was mentioned in the WG meeting that it can be fairly useful to be able to clone the Logger so it can be used from outside the node that defined it. In that case should we derive `Clone`?

mxgrey: Great point, and I can change `Box<str>` to `Arc<str>` and `CString` to `Arc<CString>` for better efficiency.

rclrs/src/logging.rs
```diff
@@ -0,0 +1,604 @@
...
+            }
+
+            // If we have a throttle duration then check if we're inside or outside
+            // of that interval.
+            let throttle = params.get_throttle();
+            if throttle > std::time::Duration::ZERO {
+                static LAST_LOG_TIME: OnceLock<Mutex<SystemTime>> = OnceLock::new();
+                let last_log_time = LAST_LOG_TIME.get_or_init(|| {
+                    Mutex::new(std::time::SystemTime::now())
+                });
```
luca-della-vedova, line 134: I'm not too sure about this implementation, from what I see it hardcodes the time to a `std::time::SystemTime` but in ROS 2 the clock could be a ROS time.
The `rclcpp` implementation also seems to allow specifying arbitrary clocks and has [unit tests](https://github.com/ros2/rclcpp/blob/88ebea94e9c79499c34f825cf6565827c5ba958e/rclcpp/test/rclcpp/test_logging.cpp#L178-L230) to verify that passing a ROS clock correctly throttles it.
I would suggest either making the clock to be used configurable or maybe easier just defaulting to the node's clock, so at least when running on simulated time we will behave consistently regardless of the speed of the simulation.

mxgrey: Interesting, the idea of syncing it against ROS time didn't even occur to me. That strikes me as a weird use case, but I'm not opposed to it.

We can't rely on the node clock without making it mandatory to pass in a reference to the node. That could make logging unusable in code regions where it shouldn't have to be.

I don't mind making this configurable by adding a `.throttle_clock( _ )` option to `LogParams` that takes in an enum like

```
pub enum ThrottleClock {
    Clock(Clock),
    SystemTime,
    SteadyTime,
}
```

and this would default to `SteadyTime` (which I think is what 99.99% of use cases would want).

mxgrey: Added in https://github.com/geoff-imdex/ros2_rust/pull/13 and just waiting to be merged in that PR

rclrs/src/logging/logger.rs
mxgrey: Added in https://github.com/geoff-imdex/ros2_rust/pull/13 and just waiting to be merged in that PR

rclrs/src/logging.rs
luca-della-vedova: I admit I have never used the throttling API but when I'm in doubt on "what is the usage in the ecosystem" I usually just throw a [global github search](https://github.com/search?q=RCLCPP_DEBUG_THROTTLE&type=code) and glance at the code out there.

It is probably mandated by the fact that the macro in C++ requires you to pass a clock and it is probably easier to pass the node clock rather than create a new steady_clock just for that purpose, but it seems the majority of applications do indeed pass in the node clock (which would be running in either system time or simulated time depending on parameters).
I also wonder, what's the reason for redefining a `ThrottleClock` enum rather than using [Clock](https://github.com/ros2-rust/ros2_rust/blob/066dd7c6e70d515268945ab6190793ef605fd887/rclrs/src/clock.rs#L43-L58)? It also offers steady time / system time enum variants

mxgrey: >  it seems the majority of applications do indeed pass in the node clock

Are you suggesting that we make it mandatory, (e.g. instead of `.throttle(duration)` we only offer `.throttle(clock, duration)`)? If so, what do you perceive as the benefit of mandating an explicit clock instead of using steady clock by default and letting the user override that?

In my mind ROS time exists primarily to make sure that system behavior can be consistent across real world usage and simulated usage. Log throttling doesn't (..or at least shouldn't..) have an effect on the behavior of the system beyond preventing log spam. It's very hard for me to imagine a case where using ROS time is preferable for reducing log spam, so I would much prefer minimizing API noise and providing a sensible default which will probably always be what users actually want.

In any case the throttle time source type [is an optional parameter in `rclpy`'s logging API](https://github.com/ros2/rclpy/blob/34f9e13e72ea4528bd96f1d739bc7a8711cc923b/rclpy/rclpy/impl/rcutils_logger.py#L350-L354) with a default of steady time, which is exactly what I've done in https://github.com/geoff-imdex/ros2_rust/pull/13, so we'll be matching the API of rclpy even if we don't match the API of rclcpp.

> what's the reason for redefining a ThrottleClock enum rather than using [Clock](https://github.com/ros2-rust/ros2_rust/blob/066dd7c6e70d515268945ab6190793ef605fd887/rclrs/src/clock.rs#L43-L58)? It also offers steady time / system time enum variants

The difference is `ThrottleClock` borrows a clock so there's no need to clone a clock. Cloning a clock would involve incrementing and decrementing an atomic int, which isn't a terrible amount of overhead but it does involve cache syncs which should be avoided whenever possible for hot paths.

luca-della-vedova: > Are you suggesting that we make it mandatory, (e.g. instead of .throttle(duration) we only offer .throttle(clock, duration))? If so, what do you perceive as the benefit of mandating an explicit clock instead of using steady clock by default and letting the user override that?

Not really, I see the benefit of the sensible default, which is why I mentioned that probably people pass in the node clock just because the API forces them to and that's the most easily available clock. I'm fine with the current API.

> The difference is ThrottleClock borrows a clock so there's no need to clone a clock. Cloning a clock would involve incrementing and decrementing an atomic int, which isn't a terrible amount of overhead but it does involve cache syncs which should be avoided whenever possible for hot paths.

I see, my main point was about the underlying implementation as well, I am not quite sure but would it make sense to use a `Clock` with `ClockKind::SystemTime` / `ClockKind::SteadyTime` for the underlying implementation in the macro, rather than relying on Rust builtin methods?
I can see how the Rust method is probably cleaner and faster since it doesn't need to build an rcl clock or go through bindings. I however do not honestly know if the behavior of the two is guaranteed to be the same or not, and whether that might be an issue (I suppose both will do the same underlying system calls to get a steady / system time so it's probably OK?).
I'm also slightly put off by having different ways to deal with time internally (rcl / rust std), but I suppose this being purely an implementation detail and not exposed to users it is probably not a big deal?

mxgrey: > I can see how the Rust method is probably cleaner and faster since it doesn't need to build an rcl clock or go through bindings.

Yes, all of this, plus the fact that going through the rcl bindings also forces a heap allocation (creating the `Arc<Mutex<rcl_clock_t>>`) which is something to avoid for hot paths, and the whole point of throttling a log is because that log exists in a hot path.

I do agree that there's a somewhat silly amount of duplication of clock abstractions going on here, but here's the way I would look at it:
* `ThrottleClock` allows a choice of clock to be made at the point in the code where the logging is happening
* `ThrottleClock::Clock` allows a choice of clock to be made elsewhere, and the logging will just abide by that choice

> I suppose both will do the same underlying system calls to get a steady / system time so it's probably OK?

I would expect so as well. I've very surprised if they're using different syscalls.

> I'm also slightly put off by having different ways to deal with time internally (rcl / rust std)

Yeah I don't love it myself. It mostly comes from `rclrs` trying to accommodate the full breadth of `rcl`'s clock interface. Personally I wouldn't mind if we narrow down on what we support from the API of `rcl_clock_t`.. maybe even not use `rcl_clock_t` at all and just implement the whole abstract clock concept on the Rust side without any rcl baggage..? But that's definitely out of scope for this PR.

luca-della-vedova:
The diff seems to still be messed up? I wonder if updating upstream main and rebasing is the only way to solve it?

luca-della-vedova:
I also mentioned in my [previous comment](https://github.com/ros2-rust/ros2_rust/pull/422#pullrequestreview-2444795236) that it would be nice to use it in the examples rather than what we do currently which is just doing `println`.

Anyway I gave it a spin by applying the following patch locally:

```
$ git diff
diff --git a/examples/minimal_pub_sub/src/minimal_publisher.rs b/examples/minimal_pub_sub/src/minimal_publisher.rs
index 7200869..5931327 100644
--- a/examples/minimal_pub_sub/src/minimal_publisher.rs
+++ b/examples/minimal_pub_sub/src/minimal_publisher.rs
@@ -1,6 +1,7 @@
 use std::env;
 
 use anyhow::{Error, Result};
+use rclrs::{log, ToLogParams};
 
@@ -16,7 +17,7 @@ fn main() -> Result<(), Error> {
 
     while context.ok() {
         message.data = format!("Hello, world! {}", publish_count);
-        println!("Publishing: [{}]", message.data);
+        log!(node.info().skip_first().throttle(std::time::Duration::from_millis(1000)), "Publishing: [{}]", message.data);
         publisher.publish(&message)?;
         publish_count += 1;
         std::thread::sleep(std::time::Duration::from_millis(500));
```

Then running the publisher:

```
ros2 run examples_rclrs_minimal_pub_sub minimal_publisher
```

It seems to do the right thing

```
[INFO] [1732159768.496156067] [minimal_publisher]: Publishing: [Hello, world! 4]
[INFO] [1732159769.496589258] [minimal_publisher]: Publishing: [Hello, world! 6]
```

Do you think it's worth changing the examples here or should we just do it in a followup PR to avoid making the diff even larger? I'm happy either way but I'd rather this TODO doesn't get lost

rclrs/src/logging/log_params.rs
```diff
@@ -0,0 +1,293 @@
...
+    /// The severity of the logging instance.
+    severity: LogSeverity,
+    /// Specify when a log message should be published (See[`LoggingOccurrence`] above)
+    occurs: LogOccurrence,
+    /// Specify a publication throttling interval for the message.  A value of ZERO (0) indicates that the
+    /// message should not be throttled. Otherwise, the message will only be published once the specified
+    /// interval has elapsed. This field is typically used to limit the output from high-frequency messages,
+    /// e.g. if `log!(logger.throttle(Duration::from_secs(1)), "message");` is called every 10ms, it will
+    /// nevertheless only be published once per second.
+    throttle: Duration,
```
luca-della-vedova, line 18: Nit, can this be an `Option<Duration>` instead to be a bit more idiomatic about whether a throttling setting exists or not?

rclrs/src/parameter.rs
```diff
@@ -1366,7 +1369,7 @@ mod tests {
     #[test]
     fn test_optional_parameter_apis() {
         let ctx = Context::new([]).unwrap();
-        let node = create_node(&ctx, "param_test_node").unwrap();
+        let node = create_node(&ctx, &format!("param_test_node_{}", line!())).unwrap();
```
luca-della-vedova, line 1372: Are those changes to make sure we don't have name clashes when running tests?

mxgrey: Yep, the tests were producing lots of spam warnings because of clashing node names while tests ran in parallel.

rclrs/src/logging/log_params.rs
mxgrey: I did consider this, but I decided to stick with simply `Duration` for the following reasons:

* This field is private, so whether it's `Duration` or `Option<Duration>` is an implementation detail that should be considered invisible to users.
* Users never need to know to set `Duration::new(0, 0)` to get no throttle because no throttle is implied by default. Users only need to specify positive throttle. Therefore as far as users are concerned, there is no special rule that they need to know about.
* We need to check for zero or negative throttle anyway, even if we have an `Option<Duration>` because a negative duration would be invalid and a zero duration should have us skip throttling, just like we currently do, so there's no advantage for the implementation.
* A zero duration throttle is mathematically equivalent to no throttle, so we're not introducing a nonsense meaning for the value, in contrast to a simulator using 0 mass to mean an object is unmovable, where 0 is being given the opposite of its actual physical meaning.
* This is a very very minor, but `Option<Duration>` has a slightly larger memory footprint than `Duration`.

I'm open to reconsidering, but overall I would still lean towards just using `Duration`.

mxgrey:
> Do you think it's worth changing the examples here or should we just do it in a followup PR to avoid making the diff even larger?

I'd like to update the examples in a follow-up PR so we don't keep inflating the size of this one. I think the existing tests give enough demonstration for the reviewers of what the logging API looks like and evidence that it works as intended. Examples can be something we give more thought to after the implementation is in.

I'll open an issue for this so we don't lose sight of it.

rclrs/src/node.rs
```diff
@@ -66,23 +66,48 @@ pub struct Node {
...
+/// to guarantee that the node handle exists until we drop the NodeHandle
+/// instance. This addresses an issue where previously the address of the variable
+/// in the builder.rs was being used, and whose lifespan was (just) shorter than the
+/// NodeHandle instance.
 ///
 /// [1]: <https://doc.rust-lang.org/reference/destructors.html>
 pub(crate) struct NodeHandle {
     pub(crate) rcl_node: Mutex<rcl_node_t>,
     pub(crate) context_handle: Arc<ContextHandle>,
+    /// In the humbe distro, rcl is sensitive to the address of the rcl_node_t
```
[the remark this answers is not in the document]
mxgrey: Whoops, fixed by [622e887](https://github.com/ros2-rust/ros2_rust/pull/422/commits/622e887c7112463c68aef05d12c3b08be68c42ef#diff-edcb3d03aa3bd1a7a070881f3faa29fc9b06812e2fc26eaeb32e2d431067f268R85)

rclrs/src/parameter.rs
geoff-imdex: For some extra context, pre-Jazzy, the rcl node was responsible for creating the rosout publisher and the rosout logging code would warn if it had already created a publisher for a node with the same name.
In the Jazzy (and rolling) releases, the creation of the rosout publisher is no longer the responsibility of the rcl node and so the issue doesn't occur.
It does also mean that we will need to have Jazzy (and on) specific code for explicitly creating the rosout publisher.  Following the rclcpp/rclpy implementations, this can easily be added to the node builder code.  I'm happy to create a ticket for this so the work doesn't get lost if that helps?

jhdcs (approved):

PR #423, opened 2024-10-27
Update vendored interface packages

nwn:
In preparation for `rclrs` supporting actions (#410), it will need new dependencies on certain interface packages: specifically, `action_msgs` and its dependency `unique_identifier_msgs`. Since `rclrs` is a crate on [crates.io](https://crates.io/) and the interface packages are not, they need to be vendored into the `rclrs` code. This PR does so.

In the first commit, the `rclrs/vendor_interfaces.py` script is updated to vendor the above message packages in addition to the existing ones. In the process, the set of vendored packages is collapsed into a single list rather than being repeated throughout the script. Support is also added for packages with an `action.rs` file, though this is not currently used.

The second commit silences the `clippy::derive_partial_eq_without_eq` and `clippy::upper_case_acronyms` lints in the generated code for interface packages. This was already being done by the vendored packages in `rclrs`, but was moved to `rosidl_generator_rs` for the sake of downstream users who may also enforce `clippy` on them.

The third commit is the result of running the updated script on the `install` output of the desired packages with the latest version of the `rosidl_generator_rs` generator. In addition to introducing the new vendored packages, this resulted in some trivial changes to the existing ones: replacing `std::os::raw::c_void` with its alias `std::ffi::c_void` and a whitespace change.


maspe36 (approved):
LGTM

Issue #424, opened 2024-10-29
No rust crate created for service_msgs during build


azerupi:
Hi,
I'm trying to build ros2_rust on jazzy. I understand this is not fully supported yet?

When building I can see that all `*_msgs` packages get a rust crate except for the `service_msgs` package. This then fails the build because it is required by `test_msgs` in `rclrs`.

Does anyone know if this is a jazzy issue or if it is something else?

```
colcon build
Starting >>> builtin_interfaces
Starting >>> unique_identifier_msgs
Starting >>> rosidl_runtime_rs
Starting >>> test_interface_files
Starting >>> latency_dashboard
Finished <<< rosidl_runtime_rs [0.15s]
Finished <<< test_interface_files [0.15s]
Starting >>> rosidl_generator_rs
Finished <<< latency_dashboard [0.17s]
Finished <<< rosidl_generator_rs [0.44s]
Finished <<< unique_identifier_msgs [0.62s]
Finished <<< builtin_interfaces [0.67s]
Starting >>> service_msgs
Finished <<< service_msgs [0.56s]
Starting >>> action_msgs
Starting >>> type_description_interfaces
Finished <<< action_msgs [1.00s]
Starting >>> rosidl_default_generators
Starting >>> rosidl_default_runtime
Finished <<< rosidl_default_generators [0.19s]
Finished <<< rosidl_default_runtime [0.20s]
Starting >>> std_msgs
Starting >>> px4_msgs
Starting >>> rcl_interfaces
Starting >>> rosgraph_msgs
Starting >>> test_msgs
Starting >>> example_interfaces
Starting >>> rclrs_example_msgs
Starting >>> std_srvs
Starting >>> lifecycle_msgs
Starting >>> statistics_msgs
Finished <<< type_description_interfaces [1.24s]
Finished <<< rosgraph_msgs [0.94s]
Finished <<< std_srvs [1.00s]
Finished <<< statistics_msgs [1.04s]
Finished <<< rclrs_example_msgs [1.10s]
Finished <<< lifecycle_msgs [1.49s]
Finished <<< test_msgs [1.92s]
Finished <<< rcl_interfaces [1.96s]
Starting >>> rclrs
Starting >>> composition_interfaces
Finished <<< std_msgs [2.51s]
Starting >>> geometry_msgs
Starting >>> actionlib_msgs
Finished <<< example_interfaces [2.57s]
Finished <<< composition_interfaces [0.80s]
[...]

PR #425, opened 2024-10-30
Add a .repos file for jazzy

mxgrey:
To address #424 this PR adds a .repos file for jazzy.

The new repos file is exactly the same as the file for rolling, but the branches have been switched to jazzy. I used this on 24.04 with the jazzy distro sourced and the build and test went fine.

### DIFF
--- a/ros2_rust_jazzy.repos
+++ b/ros2_rust_jazzy.repos
@@ -1,0 +1,29 @@
+repositories:
+  ros2/common_interfaces:
+    type: git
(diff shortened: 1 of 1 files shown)


mxgrey:
We may want to consider moving the repos files into their own folder so they don't clutter the root directory if we need to keep adding more.

OTOH we should be able to remove all of them once message generation makes it onto the build farm.

esteve (approved):
@mxgrey thanks!

PR #440, opened 2024-11-29
Basic Timer implementation

agalbachicar:
# Description

This is a basic Timer implementation. It is heavily based on what other client libraries do, i.e. use RCL bindings and provide a safe API against the bindings.

## Details

It is by no means a final version but it is a good first approach to something close to the final version. We will happily change everything you tell us to change :)

It is important to look at the `TimerCallback` type trait definition.

We tried to keep mutable and non-mutable functions within `Timer` but then when integrating it against `WaitSet`, `Node` and `Executor` we had to make it all "non-mutable". 

This has been done in collaboration (peer programming) with @tulku , @JesusSilvaUtrera, @jballoffet :confetti_ball: 

## How to test it?

- [ ] Add end to end example 

We wrote an end to end example, we will soon add it so it can be show how to do it.
We introduced some unit tests, we can certainly extend them as soon as we define the final API.

# Notes

We observed some non-trivial delays in the callback execution. Certainly the overhead between the bindings and what we are doing is non-negligible. We would appreciate some input here.

# Identified discussion points

- i64 for durations, shall we use proper `Duration` types?
- Using the `Clock::with_source()` makes the test in `Timer` crash with SIGSEGV. See the comment and test in timer.rs.

### DIFF
--- a/rclrs/src/clock.rs
+++ b/rclrs/src/clock.rs
@@ -28,8 +28,7 @@ impl From<ClockType> for rcl_clock_type_t {
 #[derive(Clone, Debug)]
 pub struct Clock {
     kind: ClockType,
-    // TODO(ekumen): Fix the extra pub here.
-    pub rcl_clock: Arc<Mutex<rcl_clock_t>>,
+    rcl_clock: Arc<Mutex<rcl_clock_t>>,
     // TODO(luca) Implement jump callbacks
 }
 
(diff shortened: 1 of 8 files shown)


jhdcs:
In reference to your discussion points, I think that having timers use `Duration` would be valuable.

agalbachicar:
And also here, thanks a lot for such a feedback, @mxgrey ! I'll review it with my peers and get back as soon as possible.

romainreignier:
Just a friendly reminder that it would be nice to have timers in ros2_rust.

I have seen in https://github.com/agalbachicar/ros2_rust/pull/5 that @mxgrey has done a rewrite.
Could we try to upstream this new version or is this in conflict with the async refactor and needs further API changes?

mxgrey:
I think it would be reasonable to close this PR in favor of #480 

The original authors for this PR are credited in the new one.

agalbachicar:
Thanks! Closing then.

PR #441, opened 2024-12-02
Update for clippy 1.83

mxgrey:
Stable CI for new PRs are failing right now because of an update in clippy 1.83 that emits a warning when an explicit lifetime can be elided.

This PR updates based on the clippy recommendation, so Stable CI should go green now.


jhdcs (approved):
Quick and simple fix. Thank you!

PR #442, opened 2024-12-04
chore: remove references to installing cargo-ament-build via cargo

esteve:
Now that [cargo-ament-build](https://pypi.org/project/cargo-ament-build/) is avaialble via PyPI as a Python wheel and that [cargo-ros-cargo](https://pypi.org/project/colcon-ros-cargo/) depends on it, we can install both in one go.

### DIFF
--- a/.github/workflows/rust-minimal.yml
+++ b/.github/workflows/rust-minimal.yml
@@ -76,10 +76,6 @@ jobs:
         cd -
         done
 
-    - name: Install cargo-ament-build
-      run: |
-        cargo install --debug cargo-ament-build
-
     - name: Build and test
       id: build
       uses: ros-tooling/action-ros-ci@v0.3
--- a/.github/workflows/rust-stable.yml
+++ b/.github/workflows/rust-stable.yml
@@ -76,10 +76,6 @@ jobs:
         cd -
         done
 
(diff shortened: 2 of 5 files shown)


jhdcs (approved):

luca-della-vedova:
Ah this all happened while I was sleeping :sweat_smile: 

I guess this is OK for now but I would have argued that we will actually need to do the opposite soon, since installing `pip` packages in Jazzy / 24.04 is [very painful](https://github.com/ros2-rust/ros2_rust/pull/400), and it is unclear whether we will be able to package `cargo-ament-build` as a debian package.
I suspect the final intermediate solution will be to revert this and actually install `cargo-ament-build` from `cargo` and let `apt` manage the colcon extensions instead

esteve:
I don't think that would be necessary, the `cargo-ament-build` Python wheel could be distributed as a Debian package as is, as a precompiled binary and we wouldn't need to worry about the intermediate solution you describe.

For me, the ideal solution would be to mimic what the [ruff](https://packages.debian.org/sid/ruff) Debian package does.

mxgrey:
If the precompiled binary can be distributed as-is through debian then that's definitely the best case scenario.

PR #444, opened 2024-12-09
Fix buffer overflow for bounded sequence example

mxgrey:
The example message `VariousTypes` has a `string<=3` with a default value containing three characters which each have a length of two bytes in utf-8. This means the real length of the default value is 6 bytes even though the byte limit is `<=3`.

This PR changes the default value to be `"aou"` to be three bytes long. The other bounded string is a wide string (utf-16), so it can remain as-is.

To observe the problem, use `$ cargo run` in the `message_demo` example to see it return an error code if you use the `main` branch. Then use the branch of this PR and see that it does not return an error code.


esteve:
@mxgrey thanks!

esteve (approved):

PR #450, opened 2024-12-18
Use 1.75-compatible lockfiles when running Minimal CI

### LINKED ISSUE #449: Rust Minimal Workflow Failing
Our [latest Rust Minimal](https://github.com/ros2-rust/ros2_rust/actions/runs/12365587755) workflow failed with the following.

```
   Starting >>> rclrs
  --- output: rclrs
  error: package `home v0.5.11` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.75.0
  Either upgrade to rustc 1.81 or newer, or use
  cargo update home@0.5.11 --precise ver
  where `ver` is the latest version of `home` supporting rustc 1.75.0
  ---
  --- stderr: rclrs
  error: package `home v0.5.11` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.75.0
  Either upgrade to rustc 1.81 or newer, or use
  cargo update home@0.5.11 --precise ver
  where `ver` is the latest version of `home` supporting rustc 1.75.0
  ---
  Failed   <<< rclrs [1.32s, exited with code 1]
  ```
  
 Changelog for home can be found [here](https://github.com/rust-lang/cargo/blob/home-0.5.11/crates/home/CHANGELOG.md)

mxgrey:
This is an implementation of the suggestion that I made [here](https://github.com/ros2-rust/ros2_rust/issues/449#issuecomment-2550279764) that aims to keep the Minimal CI stable even as new versions of dependencies (including transitive dependencies) are released.

The lockfiles are admittedly somewhat annoyingly bulky, especially since each example crate needs one. We should be able to reduce that to a single lockfile if we double this up with [the suggestion here](https://github.com/ros2-rust/ros2_rust/pull/428#issuecomment-2551232392) to put a workspace at the root of the repo.

I understand that this PR may be controversial, so I'm opening it to test whether it even works and then facilitate discussion about whether or not this is how we'd like to resolve #449, or whether we should consider another approach.


mxgrey:
It seems like colcon-cargo might be ignoring the lockfiles that are present inside of each crate directory, although it's hard to tell for sure since the workflow terminal output gives pretty limited information.

In any case the lockfile approach doesn't seem to be working so I'll try a different approach.

PR #454, opened 2025-01-28
refactor: moved subprojects to separate repositories

esteve:
Corresponding PRs are:

 - https://github.com/ros2-rust/rosidl_runtime_rs/pull/2
 - https://github.com/ros2-rust/rosidl_rust/pull/1


esteve:
I didn't move the docs folder, I thought it'd be good to have them alongside `rclrs`, but I'm also fine movinth that to a separate repo.

esteve:
I had to tweak the `repos` files so that they would point to the forks, but those will have to be reverted once the corresponding PRs are merged and before merging this one.

esteve:
Marked as draft to avoid merging by mistake, the `repos` files need to be updated once the other PRs are merged.

mxgrey:
Are we sure about moving examples into another repo? That will make it harder to ensure that the examples stay in sync with the latest version of rclrs, especially with the incoming API changes.

Iirc the main motivation for splitting the repo is so we can eventually release the crates to the build farm while managing their versions separately. I wouldn't expect us to release the example crates, so there shouldn't be any detriment to keeping them in the same repo as rclrs. It's also very common practice in the Rust ecosystem to keep example projects in the same repo as the crate that they're demonstrating.

esteve:
@mxgrey that's a good point, I was following the structure of the `rclcpp` and `rclpy` repositories. However, I wouldn't dismiss releasing the examples as ROS packages anyway, they are released for the other official clients (e.g. https://index.ros.org/p/examples_rclcpp_minimal_publisher/#humble https://index.ros.org/p/examples_rclpy_minimal_publisher/#humble) and I can see their usefulness for demonstrating the functionality of the client libraries.

maspe36:
Do we need to have 1 ROS package per repo? Couldn't we keep examples in-place and just package them separately? Its more "rusty" IMO to have the examples stay in tree

A few examples from the rust ecosystem
- https://github.com/tokio-rs/tokio/tree/master/examples
- https://github.com/bevyengine/bevy/tree/main/examples
- https://github.com/BurntSushi/ripgrep/tree/master/crates/grep/examples (The repo structure here is interesting, but the point is examples per crate).

mxgrey:
It's worth considering the possibility that we would release the examples on the build farm as well. 

I think even in that situation, it wouldn't be a bad thing to always re-release the examples with each release of rclrs and to make the version numbers of the examples always match the version number of rclrs.

The examples would be leaf nodes in the build tree and they'd need to be rebuilt with each rclrs release anyway since Rust crates can't be updated with a new version of a dependency without being rebuilt anyway.

esteve:
@mxgrey thinking about it I agree with not moving the examples folder. Especially given that it'd be counterintuitive to release binaries that have been compiled with an older version of `rclrs`, I'll keep the examples in this repo, but still move the other ones to their respective repositories.

@maspe36 no, we don't need one repo per package. See for example the rclcpp repository (https://github.com/ros2/rclcpp) which contains three packages. Moving rosidl_generator_rs and rosidl_runtime_rs make sense as they are mostly released at a different pace. Plus it'd help us get the generator in the buildfarm as soon as possible until rclrs is in a state that can be released by the buildfarm

esteve:
@maspe36 @mxgrey @nwn @jhdcs Now that https://github.com/ros2-rust/rosidl_rust/pull/1 and https://github.com/ros2-rust/rosidl_runtime_rs/pull/2 have been merged, this PR is now ready to be reviewed. The changes here should be straightforward to review, I just removed `rosidl_runtime_rs` and `rosidl_generator_rs`

jhdcs (approved):
These changes look good to me, though I feel a second opinion might be warranted as I wasn't super involved in the reviews for this before now.

mxgrey (approved):
Everything looks like it's in order 👍

maspe36 (approved):
LGTM

PR #455, opened 2025-02-02
Add support for the newer Rolling and Jazzy distros in our dev Dockerfile

maspe36:
- Change the pip command based on the ROS_DISTRO argument
- Swap to the PyPi version of [colcon-ros-cargo](https://pypi.org/project/colcon-ros-cargo/)
- Only install `colcon-ros-cargo` as it lists `colcon-cargo` and `cargo-ament-build` as dependencies ([link](https://github.com/colcon/colcon-ros-cargo/blob/main/setup.cfg#L29))

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 ARG ROS_DISTRO=humble
-FROM ros:$ROS_DISTRO as base
+FROM ros:$ROS_DISTRO AS base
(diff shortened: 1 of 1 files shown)


jhdcs (approved):
Tested using both jazzy and rolling repos files, and this appeared to work for me.

PR #457, opened 2025-02-03
Add Publisher::get_subscription_count

### LINKED ISSUE #456: get_subscription_count() missing for Publisher
Hi, to avoid wasting CPU if my publisher has no subscriber, I would prefer to prepare the message to send using something like [get_subscription_count()](https://docs.ros.org/en/ros2_packages/rolling/api/rcl/generated/function_publisher_8h_1a170852c103ae30e42835608f08cc772a.html) on the `Publisher` object but this function does not seem available in rclrs.

I will try to provide de PR.

romainreignier:
This should close #456 

This is my first contribution to ros2_rust and I am not sure about the return code handling and safety comment.
Any feedback is welcomed.

### DIFF
--- a/rclrs/src/publisher.rs
+++ b/rclrs/src/publisher.rs
@@ -146,6 +146,20 @@ where
         }
     }
 
+    /// Returns the number of subscriptions of the publisher.
+    pub fn get_subscription_count(&self) -> Result<usize, RclrsError> {
+        let mut subscription_count = 0;
+        // SAFETY: No preconditions for the function called.
+        unsafe {
+            rcl_publisher_get_subscription_count(
+                &*self.handle.rcl_publisher.lock().unwrap(),
+                &mut subscription_count,
+            )
+            .ok()?
+        };
+        Ok(subscription_count)
+    }
+
     /// Publishes a message.
     ///
     /// The [`MessageCow`] trait is implemented by any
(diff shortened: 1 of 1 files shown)


rclrs/src/publisher.rs
```diff
@@ -146,6 +146,20 @@ where
...
+        let ret = unsafe {
+            rcl_publisher_get_subscription_count(
+                &*self.handle.rcl_publisher.lock().unwrap(),
+                &mut subscription_count,
+            )
+        };
+        debug_assert_eq!(ret, 0);
+        subscription_count
+    }
+
```
jhdcs, line 162:
```suggestion
    pub fn get_subscription_count(&self) -> Result<usize, RclrsError> {
        let mut subscription_count = 0;
        // SAFETY: No preconditions for the function called.
        unsafe {
            rcl_publisher_get_subscription_count(
                &*self.handle.rcl_publisher.lock().unwrap(),
                &mut subscription_count,
            )
            .ok()?
        };

        Ok(subscription_count)
    }

```

We've set things up so that the return codes from `rcl` can be automatically mapped to specialized Rust error types (namely, the `RclrsError` type). So in theory this _should_ work, I think.

As for your safety comment, I think it is correct. We use safety comments to document why the usage of an `unsafe` block is unlikely to cause issues down the road and/or document what precautions must be followed if there IS a possibility of the unsafe block biting us. Since at this point the publisher should be created, and this function call isn't creating anything special besides a simple `usize`, I think that your comment is correct. Though I welcome other maintainers to correct me if I'm wrong on this point.

Thank you for the PR! Just needs this tiny tweak to make things more idiomatic, but this should hopefully be all that's needed!

jhdcs (changes requested):

rclrs/src/publisher.rs
romainreignier: Thanks for the review and your comment.
Your proposed version was actually my first try I had a second pass and though that the return code might be always 0 in out case.

I will update my branch.

jhdcs:
Looking good so far! I did think of one other thing that should probably be added. In the bottom of the file, there is a `test_publishers` unit test. Would you be able to add a test for your code to it? Since there aren't any subscribers created in the test, we should expect a count of 0, but this is good practice to include anyways.

Sorry for the extra task there, wish I had thought of it before...

romainreignier:
Thanks for the suggestion, I have added the tests.

romainreignier:
Sorry, I had missed a publisher to test. I have tried to add a to each topic and it works as well.

jhdcs (approved):

jhdcs:
Confirmed that this works on my setup. Thank you very much for your contribution!

Issue #431, opened 2024-11-21
Add examples for logging


mxgrey:
After #422 is merged we should open a follow-up PR to add logging examples.

Maintainers can feel free to assign this issue to me.

PR #459, opened 2025-02-21
Added a simple ros2 rust logger example

Guelakais:
added simple example for ros2 rust node with logging

added the manifest file for the example

added the ros2 manifest file

modechange


esteve (changes requested):
@Guelakais thanks for your contribution. Could you revert the file mode changes? All the files are marked as executable in your PR and it makes it more difficult to review the actual changes when there are more than 100 files changed

Guelakais:
> @Guelakais thanks for your contribution. Could you revert the file mode changes? All the files are marked as executable in your PR and it makes it more difficult to review the actual changes when there are more than 100 files changed

I think it's better now. I have split the whole thing into more commits and left out the modechanges.

.github/workflows/rust-minimal.yml
```diff
@@ -2,14 +2,14 @@ name: Rust Minimal
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
```
esteve, line 5: Please revert all the changes to the workflows, they are unrelated to adding a logging example.

esteve (changes requested):

Guelakais:
Then I'm a bit confused about the errors. For me, the example was built without any problems.

esteve:
@Guelakais what errors? The CI passes:

https://github.com/ros2-rust/ros2_rust/pull/459/checks

romainreignier:
If I may add a comment, I think that this new example is a lot of duplication of the [minimal_publisher](https://github.com/ros2-rust/ros2_rust/blob/main/examples/minimal_pub_sub/src/minimal_publisher.rs) example.
I think it would make more sense to add the simple log line to the minimal examples and instead, either add more examples in the rustdoc or add a specific example to show the advanced features of logging as done by @avanmalleghem in https://github.com/ros2-rust/ros2_rust/pull/460.

Guelakais:
> If I may add a comment, I think that this new example is a lot of duplication of the [minimal_publisher](https://github.com/ros2-rust/ros2_rust/blob/main/examples/minimal_pub_sub/src/minimal_publisher.rs) example. I think it would make more sense to add the simple log line to the minimal examples and instead, either add more examples in the rustdoc or add a specific example to show the advanced features of logging as done by @avanmalleghem in #460.

I understand your perspective. However, I disagree with the implication that we must choose between the two approaches. While it's technically feasible to augment all [examples](https://github.com/ros2-rust/ros2_rust/tree/main/examples) with logging, the core issue is not time, but runtime efficiency. Specifically, ROS 2 logging relies on publishing messages to the `\rosout` topic via DDS. Each logger instance effectively becomes a publisher, introducing overhead and potentially impacting real-time performance due to the added DDS communication. This is the primary constraint we need to address.

esteve:
> While it's technically feasible to augment all examples with logging, the core issue is not time, but runtime efficiency.

This point does not matter, these packages are examples on how to use the logging API, not fully fledged production-ready applications.

In going to pick one of the two PRs, whichever seems more complete. https://github.com/ros2-rust/ros2_rust/pull/460 seems to exercise more parts of the logging API, which is the end goal of these examples

In any case, thanks @Guelakais for your contribution

avanmalleghem:
Well... I didn't create my PR to replace the one from @Guelakais . It is more "Ok we have a small and very simple example to start with. Now let's show some other features". The issue https://github.com/ros2-rust/ros2_rust/issues/431 talks about creating logging "example**s**". Now we have 2 examples :+1:

Guelakais:
> In my point of view, examples in an available the documentation is less maintenance that a full example for a simple use case.

On the other hand, there are simply programmers who like to run the working examples on their computer first to get a feel for what the tool can do. In addition, with the rust compiler toolchain, the examples are always compiled at the same time and tested during compilation, which makes maintenance very easy. The Rust compiler usually provides very helpful explanations as to why something is not working and possible solutions.

Guelakais (commented):
I think, that's ok.

.github/workflows/rust-minimal.yml
Guelakais: done

PR #461, opened 2025-02-27
Add a job to build and deploy the main branch doc on Github pages

romainreignier:
As proposed in https://github.com/ros2-rust/ros2_rust/pull/460, I suggest to host the doc of the git version of rclrs on Github pages.

In order for it to work, you need to enable Github Pages on this repo:
image

I understand @esteve point in https://github.com/ros2-rust/ros2_rust/pull/460#issuecomment-2688224887 that people using the git version of rclrs can do `cargo doc` themselves because the doc is already published on doc.rs for the released version of the crate so feel free to decline my PR if you prefer.

But I my point was to ease the access to the documentation of the git version instead of adding simple examples that have the same content than the rustdoc snippets to reduce the maintenance effort on API changes.

The workflow was tested on my fork in the [feature/add_doc_job](https://github.com/romainreignier/ros2_rust/blob/feature/add_doc_job/.github/workflows/docs.yml) with the result exposed [here](https://romainreignier.github.io/ros2_rust/rclrs/index.html). I had to do a different branch for the PR to point to the `main` branch of this repo.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,87 @@
+name: Docs
+on:
+  push:
+    branches: [main]
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+concurrency:
+  group: deploy
+  cancel-in-progress: false
+env:
+  CARGO_TERM_COLOR: always
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    container: rostooling/setup-ros-docker:ubuntu-jammy-ros-rolling-ros-base-latest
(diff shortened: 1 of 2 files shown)


esteve:
@romainreignier fantastic, this is great! Thanks for the PR. I'll review it properly later today or tomorrow, but at first glance it looks good to me.

romainreignier:
The down side is that I have to build the full ros workspace to be able to run `cargo doc`. The job is then taking several minutes. Maybe adding this to another job would save some CPU time.

PR #481, opened 2025-05-14
Force backtrace version that is compatible with 1.75

mxgrey:
A recent version bump in `backtrace` is no longer compatible with Rust version 1.75, despite being a patch increase.

This PR forces a fixed version for several examples that have a transitive dependency on `backtrace` to keep it compatible with Rust 1.75. This should fix CI failures that recently started happening.


jhdcs (commented):
Seems a simple enough change, and CI is passing. Thank you!

maspe36 (approved):

PR #484, opened 2025-05-23
Seems like `colcon-ros-cargo` needs a `cargo-ament-build` extension

arjo129:
As stated above ran into this issue when building a clean workspace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ sudo apt install -y git libclang-dev python3-pip python3-vcstool # libclang-dev 
 # Install these plugins for cargo and colcon:
 pip install git+https://github.com/colcon/colcon-cargo.git
 pip install git+https://github.com/colcon/colcon-ros-cargo.git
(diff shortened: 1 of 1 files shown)


maspe36:
Hey thanks for the PR :) 

Taking a look, `cargo-ament-build` is also provided as a [pip package](https://pypi.org/project/cargo-ament-build/), and `colcon-ros-cargo` has it listed as an [install dependency](https://github.com/colcon/colcon-ros-cargo/blob/main/setup.cfg#L30).

This pip dep works for our [actions](https://github.com/ros2-rust/ros2_rust/actions/runs/15128160098/job/42523965400) or the [docker images](https://github.com/ros2-rust/ros2_rust/blob/main/Dockerfile#L21), so I don't think this is needed. Not sure what happened in your local environment.

arjo129:
OK That is indeed strange that pip didnt fetch the dependency.

PR #486, opened 2025-05-28
fix: add Cargo.lock to version control

esteve:
This should fix the issues in our CI with `backtrace` requiring Rust 1.82


esteve:
@maspe36 I don't understand why that's an issue on Windows, but since we agreed to commit Cargo.lock anyway, I thought this might help. Thanks for the review

PR #490, opened 2025-06-16
update CI with jazzy and noble

knmcguire:
An update to the CI

- Adding Jazzy Jalisco
- Changing Rolling tooling to Ubuntu 24.04 noble (so latest)

Do we want to keep supporting humble? and how about short term releases as iron and kilted

### DIFF
--- a/.github/workflows/rust-stable.yml
+++ b/.github/workflows/rust-stable.yml
@@ -20,14 +20,19 @@ jobs:
       matrix:
         ros_distribution:
           - humble
+          - jazzy
           - rolling
         include:
           # Humble Hawksbill (May 2022 - May 2027)
           - docker_image: rostooling/setup-ros-docker:ubuntu-jammy-ros-humble-ros-base-latest
             ros_distribution: humble
             ros_version: 2
+          # Jazzy Jalisco (May 2024 - May 2029)
+          - docker_image: rostooling/setup-ros-docker:ubuntu-noble-ros-jazzy-ros-base-latest
+            ros_distribution: jazzy
+            ros_version: 2
           # Rolling Ridley  (June 2020 - Present)
-          - docker_image: rostooling/setup-ros-docker:ubuntu-jammy-ros-rolling-ros-base-latest
+          - docker_image: rostooling/setup-ros-docker:ubuntu-noble-ros-rolling-ros-base-latest
             ros_distribution: rolling
             ros_version: 2
     runs-on: ubuntu-latest


jhdcs:
Considering we're trying to have this client library be a Tier 1, I think we should support Humble (and other versions) for as long as ROS 2 supports them. So if a version becomes EOL, our support becomes EOL.

I am open to being overruled by the other maintainers if they feel that would be an unrealistic ask, though.

knmcguire:
Alright. In that sense I can also add Kilted. Iron is of course already EOL.

Once the PR is ready to be reviewed I'll share it on Matrix

knmcguire:
I've removed Kilted again since it is not yet available in rostooling it seems: https://github.com/ros-tooling/setup-ros-docker

I'll start a issue/pr over there, but for now I've commented it out here

luca-della-vedova (commented):
Had a quick look upstream and it seems kilted is now available, can you uncomment it?

knmcguire:
Absolutely :) the rostooling peeps were really quick to respond after opening my ticket for kilted there! PR merged in record time!

knmcguire:
ah no it's too fresh. We need to wait for the new action-ros-ci (and setup-ros?) to be released with kilted. It's either that or always be on the @main version of those but don't think we should do that to keep the CI stable.

So, shall we wait for that for this PR or shall I remove Kilted again such that we can do another PR once the tooling has been released @luca-della-vedova ?

luca-della-vedova:
> ah no it's too fresh. We need to wait for the new action-ros-ci (and setup-ros?) to be released with kilted. It's either that or always be on the @main version of those but don't think we should do that to keep the CI stable.
> 
> So, shall we wait for that for this PR or shall I remove Kilted again such that we can do another PR once the tooling has been released @luca-della-vedova ?

It seems both have kilted support, `setup-ros` on a new `0.7` version (so it should be picked up automatically), `action-ros-ci` on [0.4.3](https://github.com/ros-tooling/action-ros-ci/releases/tag/0.4.3) so if we bump to 0.4 it should be picked up

.github/workflows/rust-minimal.yml
```diff
@@ -57,8 +67,11 @@ jobs:
       with:
         components: clippy, rustfmt
 
+    # Colcon can not be run in a venv which is required in Ubuntu Noble
+    # Removing the externally managed file 
     - name: Install colcon-cargo and colcon-ros-cargo
       run: |
+        sudo rm -f /usr/lib/python3.12/EXTERNALLY-MANAGED
```
luca-della-vedova, line 74: Very interesting, I guess this is an alternative to the `--break-system-packages` flag? I didn't know this could be done!

luca-della-vedova (approved):
Just a comment for curiosity but this looks great, thank you! This (especially jazzy) is very overdue

.github/workflows/rust-minimal.yml
knmcguire: ah yes I read about that flag too. But I got this idea from my co-maintainer at Crazyswarm2 (see this [line](https://github.com/IMRCLab/crazyswarm2/blob/c08db30a187858de16cbea7fa387ebe0370b5dd0/.github/workflows/ci-ros2.yml#L47)). But seems to work out!

knmcguire:
Everything is building so feel free to merge if it's all good!

jhdcs (approved):
These changes appear good to me. Going to see if I can re-run the CI...

jhdcs:
All checks have passed. Merging.

PR #501, opened 2025-08-08
build: rename generate_docs feature to use_ros_shim

esteve:
This PR is the `rclrs` counterpart of https://github.com/ros2-rust/rosidl_runtime_rs/pull/14, given that we'll use this feature for publishing the crate onto crates.io in addition to docs.rs, `use_ros_shim` might make more sense.


esteve:
@jhdcs thanks for the prompt review

PR #502, opened 2025-08-08
build: add release-plz for automating releases

esteve:
This adds [release-plz](https://release-plz.dev/) for automating releases. For a similar PR for `rosidl_runtime_rs`, check https://github.com/ros2-rust/rosidl_runtime_rs/pull/14

PR #504, opened 2025-08-11
feat: moving examples to its own repo

esteve:
This PR removes the examples , to be merged after https://github.com/ros2-rust/examples/pull/6 is resolved.


mxgrey:
Can I suggest that we just put a workspace `Cargo.toml` at the root of this repo instead of moving examples into their own repo?

It's much easier to keep the examples in sync with the library when they're collocated.

esteve:
@mxgrey at first I tried that, but for some reason I couldn't get `colcon-cargo` to build a `rclrs` workspace (see https://github.com/colcon/colcon-cargo/issues/65), and following the discussion at https://github.com/ros2-rust/ros2_rust/pull/502#discussion_r2265616682 , this would all align (release-plz running, rclrs workspace and mimic rclcpp/rclpy repo structure).

mxgrey:
At the working group meeting last night there was a consensus that we all prefer a cargo workspace solution, but we should prioritize the ability to keep releases happening. So we'll move the examples to their own repo for now, with the expectation that we might move them back after the colcon infrastructure is ready to support it.

In the meantime, @luca-della-vedova 's suggestion to add the examples to the `.repos` file is interesting. My only hesitation is that it's not a strictly necessary member of the colcon workspace. A middle ground could be to simply add a step to the CI that clones the new examples repo.

But if our strongest preference is to have the examples hosted in this repo as siblings of `rclrs` (making them mandatory workspace members for all users building from source) then it makes perfect sense to also include them in the repos file that we encourage for users who are building from source.

luca-della-vedova:
Also happy with a change in the CI to manually clone examples, but as is this introduces potential error surface of introducing breaking changes without fixing the examples

esteve:
I actually thought I had added the examples to the repos files :sweat_smile:, I just pushed a change (https://github.com/ros2-rust/ros2_rust/pull/504/commits/34cb2ee30ba05787bcbb53485395c8932d3482e2) that adds them. My goal is to keep the CI functionally the same, but moving the examples to another repository to avoid the issues with Cargo workspaces.

We can reevaluate later on if we want the examples to be part of the ros2-rust workspace or not when a user clones the repos files, but for now I prefer to not disrupt the dev workflow too much.

luca-della-vedova (approved):

esteve:
@luca-della-vedova thanks for the review!

PR #505, opened 2025-08-12
Wake up wait set when adding a new waitable

mxgrey:
This PR fixes a subtle bug that could cause subscriptions, clients, and services to stall out where new messages fail to wake up the wait set because the subscription, client, or service itself has not been added to the wait set yet.

The fix itself is very simple:

```rust
pub(crate) fn add_to_wait_set(&self, waitable: Waitable) {
    self.channel.add_to_wait_set(waitable);
    let _ = self.wakeup_wait_set.trigger(); // <--- We were forgetting to do this
}
```

This was an oversight that was missed in the original PR that introduced async execution. The result was:
* We send off an instruction to add a new waitable to the wait set
* The wait set is never told to wake up, so it keeps waiting on its current set of primitives
* Since the wait set doesn't wake up, the new waitable never gets added to it
* It becomes a matter of luck whether something currently in the wait set ever wakes it up so that the wait set can notice that a new thing has been added

And so the fix is simply to explicitly trigger the wait set to wake up whenever we add a new waitable to it.

I've also added a test that will fail (at least most of the time, it's a bit of a race condition) if it gets ported to `main` and run without the fix to `add_to_wait_set`.

As a bonus, I've added a `Publisher::notify_on_subscriber_ready` method since it was needed for the new test. Almost all of the changed lines of code are actually related to the new test.

### DIFF
--- a/rclrs/src/executor.rs
+++ b/rclrs/src/executor.rs
@@ -241,7 +241,7 @@ impl ExecutorCommands {
     }
 
     pub(crate) fn add_to_wait_set(&self, waitable: Waitable) {
-        self.async_worker_commands.channel.add_to_waitset(waitable);
+        self.async_worker_commands.add_to_wait_set(waitable);
     }
 
     #[cfg(test)]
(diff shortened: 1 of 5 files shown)


esteve (approved):
@mxgrey LGTM, thanks!

PR #508, opened 2025-08-13
added example to RclrsError

Guelakais:


luca-della-vedova (changes requested):
Sorry it's a no for me, the example really shows how to build a variant and write a `match` statement

Guelakais:
> Sorry it's a no for me, the example really shows how to build a variant and write a `match` statement

Then you haven't really looked at it. The `match` statement is merely a means to an end, ensuring that the evaluation works properly. The point is that the error is created correctly and can be evaluated using an `assert_eq` call. In this case, a match works quite well for this purpose.

luca-della-vedova:
Let me paraphrase it for `Option`.

```
let val: Option<u32> = Some(123);
match val {
    Some(v) => assert_eq!(v, 123),
    None => panic!();
}
```

Would you say this is a useful example?

Guelakais:
according to the documentation of an `Option` from the [standard library](https://doc.rust-lang.org/stable/std/option/enum.Option.html), yes:
```rust
let mut x = Some(2);
match x.as_mut() {
    Some(v) => *v = 42,
    None => {},
}
assert_eq!(x, Some(42));
```
Do you want to argue now against the standard library of rust?

luca-della-vedova:
> Do you want to argue now against the standard library of rust?

I suppose we have different interpretations, agree to disagree

mxgrey:
@Guelakais Many of the PRs that you open contain far more noise than substance. In my view, this PR is strong example of noise. At best this example is teaching the reader the syntax of a match statement, which is not a good use of a docstring for this library. If every docstring of every library was needlessly demonstrating language syntax, it would become much harder for users to find the information that they actually need from each library's documentation.

It's important that you understand, @Guelakais, that no one is entitled to have their contributions accepted to an open source project. Every time you open a PR, you are subjecting your work to public criticism, including criticism that you don't agree with. I would urge you to think carefully about whether each PR you open or each comment you leave on an open source project has substantial value. Are you creating something that will actually help people? Are you moving a conversation forward in a constructive way? If the answer to either of those questions is not a strong yes, then it's likely that you are actually wasting the time of the people who are managing that project. And their time is a precious resource that none of us in the community want to see wasted.

Don't be surprised if the kind of behavior that you're demonstrating eventually gets you banned from participating in an open source project for violating the code of conduct.

Guelakais:
I don't really understand what your problem is. Do you have something against documentation or against the `match` statement? You can also formulate this without the `match` statement. Then it would look like this:
```rust
use rclrs::{RclReturnCode,RclrsError};
 let error: RclrsError = RclrsError::RclError { code: RclReturnCode::Timeout, msg: None };
 if let RclrsError::RclError { msg, .. } = &error {
     assert_eq!(msg, &None);
}
```

mxgrey:
Standard library documentation, especially for a module as fundamental as `Option`, is largely targeting brand new users who are likely to need an introduction to the basics of the language.

Users of rclrs are not expected to be learning the Rust language from the documentation of rclrs. There are much better resources for that, such as the documentation of the standard library.

esteve:
@Guelakais nobody is arguing against the standard library or the `match` keyword, but that the "example" you've added has no value. I agree with @mxgrey and @luca-della-vedova

PR #509, opened 2025-08-13
feat: vendorize messages so that cargo update works

esteve:
This PR vendorizes the `example_interfaces` and `test_msgs` packages so that `cargo update` can work during the release-plz job (see https://github.com/ros2-rust/ros2_rust/actions/runs/16942715079/job/48016392561)

### DIFF
--- a/rclrs/src/client.rs
+++ b/rclrs/src/client.rs
@@ -162,7 +162,9 @@ where
     /// # let node = Context::default()
     /// #   .create_basic_executor()
     /// #   .create_node("test_node")?;
-    /// use test_msgs::srv::{Empty, Empty_Request, Empty_Response};
+    /// # The following line of code is only needed for rclrs internal purposes,
+    /// # please use the test_msgs crate instead.
+    /// use vendor::test_msgs::srv::{Empty, Empty_Request, Empty_Response};
     ///
     /// async fn print_hello(_response: Empty_Response) {
     ///     print!("Hello!");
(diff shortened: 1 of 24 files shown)


esteve:
@Guelakais thanks for trying to help, but please do not send any more LLM-generated responses.

rclrs/src/client.rs
```diff
@@ -159,17 +159,17 @@ where
     /// signatures and which returns a `()` (a.k.a. nothing).
     /// ```
     /// # use rclrs::*;
-    /// # use rclrs::vendor::test_msgs::srv::{Empty, Empty_Request, Empty_Response};
+    /// # use crate::rclrs::vendor::test_msgs;
```
mxgrey, line 162: I'm a bit confused about the leading `crate::` here since I thought doctests give a compilation error when you do `use crate::`. But I guess if the tests are passing then :shrug:

mxgrey (approved):
If it makes you feel any better about the vendoring, r2r is currently vendoring many more interface packages than we are: https://docs.rs/r2r/latest/r2r/#modules

Since the only interface packages we're adding here are for examples and tests, I don't think there's any substantial risk that users will run into the diamond dependency problem with them, the way they currently do with `builtin_interfaces::msg::Time`. So I don't think this PR will make the vendoring problem any worse.

esteve:
@mxgrey thanks for the review. I'd really like to find a different path, but for now it is what it is and vendorizing the messages gets the job done. We might end up just uploading all the standard messages to `crates.io`, each as a separate crate, perhaps that could pave the way for the different Rust bindings to iteroperate.

rclrs/src/lib.rs
```diff
@@ -189,7 +193,7 @@ mod service;
 mod subscription;
 mod time;
 mod time_source;
-mod vendor;
+pub mod vendor;
```
maspe36, line 196: Oh this needs to be pub because doctest [only links to public stuff](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#include-items-only-when-collecting-doctests)... Jeez, I can see this causing some confusion. I think we need to re-evaluate our message consumption more and more everyday :sweat_smile:

rclrs/src/client.rs
maspe36: I don't think we even need `crate::`. Locally, doctests seem to be passing with out it

maspe36 (approved):

rclrs/src/lib.rs
esteve: Yeah, and I had to modify the vendorize script to not complain about missing docs for messages because now they are public 😅

PR #525, opened 2025-09-01
chore: Fix Generate Bindings CI

maspe36:
Noticed two issues in our `Generate bindings` action.

1. We are trying to run `git diff` in `rclrs/src` (which is failing)
```
warning: Not a git repository. Use --no-index to compare two paths outside a working tree
```

2. The repo needs to be marked as a safe directory
```
fatal: detected dubious ownership in repository at '/__w/ros2_rust/ros2_rust'
```

---

Not quite sure how to test this since the generate bindings action is on a timer :thinking:

### DIFF
--- a/.github/workflows/generate-bindings.yml
+++ b/.github/workflows/generate-bindings.yml
@@ -73,7 +73,8 @@ jobs:
 
     - name: Submit PR
       run: |
-        pwd
+        cd $GITHUB_WORKSPACE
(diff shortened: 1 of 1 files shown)


esteve:
@maspe36 thanks! We can trigger a run after merging, no need to wait until the timer runs it.

esteve (approved):

PR #533, opened 2025-09-07
Implement the TryInto trait on rclrs::Time for std::time::Duration

maspe36:
Handles the awkward conversion from ros time (i64, nanosecs), to Duration which is constrained to positive values only. I decided to throw this together as it seems like it could be a worthwhile thing to provide. Technically, a user can still do this safely with either [i64::cast_unsigned](https://doc.rust-lang.org/stable/std/primitive.i64.html#method.cast_unsigned) (as of rust 1.87, but it wraps), or [i64::TryInto<u64>](https://doc.rust-lang.org/stable/std/primitive.i64.html#impl-TryInto%3CU%3E-for-T).

So you could also do this, it just adds a little bit of indirection which could get annoying over time.
```rust
if let Ok(pos_t) = good_time.nsec.try_into() {
    Duration::from_nanos(pos_t);
}
```

This difference in time representation was brought up by @romainreignier in the [matrix chat](https://matrix.to/#/!AiYXdTWExqamTdatBl:matrix.org/$zm8vdD2uGjLuFGEELrHTl9p1W2uFaB-jsbHIpF6IU6E?via=matrix.org&via=uni-osnabrueck.de&via=attendees.fosdem.org).

### DIFF
--- a/rclrs/src/time.rs
+++ b/rclrs/src/time.rs
@@ -1,5 +1,6 @@
 use crate::{rcl_bindings::*, vendor::builtin_interfaces};
 use std::{
+    fmt,
     num::TryFromIntError,
     ops::{Add, Sub},
     sync::{Mutex, Weak},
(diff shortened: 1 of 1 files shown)


nwn:
Both ROS and Rust have distinct notions of points in time (e.g. [`rclcpp::Time`](https://docs.ros.org/en/rolling/p/rclcpp/generated/classrclcpp_1_1Time.html) / [`std::time::Instant`](https://doc.rust-lang.org/stable/std/time/struct.Instant.html)) versus durations of time (e.g. [`rclcpp::Duration`](https://docs.ros.org/en/rolling/p/rclcpp/generated/classrclcpp_1_1Duration.html) / [`std::time::Duration`](https://doc.rust-lang.org/stable/std/time/struct.Duration.html)). Is having a conversion like this conflating the two concepts?

maspe36:
Yup, you're right @nwn. Thanks for catching that!

PR #548, opened 2025-11-04
ci: only run scheduled python bindings generation on main repo

knmcguire:
To prevent the generate python bindings to also run on our forks. It is quite invasive of creating pull requests on all of our forked repos as well.

If this works nicely, let's also add this to all the other CI as well.

But the manual triggering of actions should still work though

### DIFF
--- a/.github/workflows/generate-bindings.yml
+++ b/.github/workflows/generate-bindings.yml
@@ -38,6 +38,10 @@ jobs:
             ros_distribution: rolling
             ros_version: 2
     runs-on: ubuntu-latest
+    if: |
+        github.event_name == 'workflow_dispatch' || (
+          github.event_name == 'schedule' && github.repository == 'ros2-rust/ros2_rust'
+        )
     permissions:
       contents: write
       pull-requests: write


knmcguire:
The build CI issues are due to some connections issues of pulling some of the packages. Either you can run it again after a while or trust the process and merge already.

we won't know if it works until next tuesday anyway

knmcguire:
Could this be merged? Feel free to run the CI again as those fails were due to a connection issue to the Ubuntu's package archives

esteve:
@knmcguire I couldn't find the button to rerun the failing CI jobs, I think it's because it had been too long, I don't know. Anyway, I force pushed to your fork to trigger a new CI build, but I didn't change anything.

knmcguire:
Ah... unfortunately we are hitting that async-lock issue (#561) and also the windows binary download issue (#562) ... 

Tomorrow I'll try to figure out if infra has fixed the windows binary issue and send out a fix for that. For the async-lock issue that is something that needs to be discussed within the group (as that requires a rust version update for minimal)

esteve:
@knmcguire thanks. CI passes finally! :partying_face:

Issue #554, opened 2025-11-27
qos() method is not working correctly.


alexzeit:
Hi,
rclrs 0.6.0! The [IntoPrimitiveOptions] trait's .qos() method is not working correctly.

I want use qos BestEffort. 
So the Rust code is reading BestEffort correctly, but the subscription is still being created with Reliable QoS 

The bug location is in: [subscription.rs] lines 245-251:

`impl<'a, T: IntoPrimitiveOptions<'a>> From<T> for SubscriptionOptions<'a> {
    fn from(value: T) -> Self {
        let primitive = value.into_primitive_options();
        let mut options = Self::new(primitive.name);
        primitive.apply_to(&mut options.qos);  // <-- This SHOULD work but doesn't!
        options
    }
}`

The [apply_to()](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) method should be overriding the QoS, but something in the conversion to rcl is broken.

alexzeit:
this works for me: ` // SAFETY: No preconditions for this function.
        let mut rcl_subscription_options = unsafe { rcl_subscription_get_default_options() };
        
        // PATCH FIX: Manually construct rmw_qos_profile_t and assign each field explicitly
        // This works around a bug where the From<QoSProfile> conversion doesn't properly apply
        use crate::rcl_bindings::*;
        rcl_subscription_options.qos.history = qos.history.into();
        rcl_subscription_options.qos.depth = match qos.history {
            crate::QoSHistoryPolicy::SystemDefault { depth } => depth as usize,
            crate::QoSHistoryPolicy::KeepLast { depth } => depth as usize,
            crate::QoSHistoryPolicy::KeepAll => 0,
        };
        rcl_subscription_options.qos.reliability = qos.reliability.into();
        rcl_subscription_options.qos.durability = qos.durability.into();
        rcl_subscription_options.qos.deadline = qos.deadline.into();
        rcl_subscription_options.qos.lifespan = qos.lifespan.into();
        rcl_subscription_options.qos.liveliness = qos.liveliness.into();
        rcl_subscription_options.qos.liveliness_lease_duration = qos.liveliness_lease.into();
        rcl_subscription_options.qos.avoid_ros_namespace_conventions = qos.avoid_ros_namespace_conventions;
`

esteve:
[...]

PR #555, opened 2025-11-27
Add support for best available QoS

mxgrey:
While investigating #554 I noticed that we are currently missing options for users to set "best available" QoS for various settings. The "best available" option was introduced in Iron, so I have suppressed the new feature when compiling for humble.

I also added a unit test to validate that rcl is seeing the correct QoS as set by the user. I have not been able to corroborate the report in #554.. it seems QoS is working just fine.


esteve (approved):
LGTM, thanks

Issue #394, opened 2024-04-19
Prevent message packages from crates.io to be included during the build


esteve:
As a side effect of https://github.com/ros2-rust/ros2_rust/pull/392, a set of message packages are now published on crates.io (see https://crates.io/crates/builtin_interfaces for example), although they've been yanked, this exposes an issue in our build system that anyone could register a crate named like a message package and potentially cause issues.

I've thought of something like this:

- rosidl_generator_rs stays the same
- a new base ros2_messages crate
- the ros2_messages crate can accept a list of features, each feature is the name of a message package
- the ros2_messages crate uses ament_rs to find the paths of the generated Rust files for each message package and uses `include!()` on those files
- downstream packages can access the messages via `ros2_messages::buildin_interfaces`, `ros2_messages::std_msgs`, etc.

I'm going to work on this urgently as the current situation needs to be fixed before it causes any breakage.

@jhdcs @maspe36 @mxgrey @luca-della-vedova thought?

mxgrey:
This is an interesting idea. But it's such an unconventional use of crates.io and cargo that I'd need to think very carefully about potentially bad side effects. 

Would we also publish a `ros2_messages` crate to crates.io?
* If so, what would it include? I'm not even sure how `include!()` works for published packages.
* If not, what stops someone from publishing one and breaking this new approach?

esteve:
@mxgrey `ros2_messages` would be published, that'd be our way of gatekeeping anyone from abusing crates.io to inject any dependencies. I'd argue that our combination depending on message packages whose version is "`*`" and using `colcon` is far more unconventional.

> If so, what would it include? I'm not even sure how include!() works for published packages.

It'd have code to read features from `Cargo.toml` and call `include!(...)` the appropriate messages. We already publish `rclrs` to crates.io and it contains a call to `include!(...)`

jhdcs:
The problem is that no matter what, the interaction between `colcon` and `cargo` has _always_ forced us to make unconventional approaches. They're both opinionated systems, and it feels like their opinions often don't match. @esteve already mentioned the black magic we're currently doing in the post above.

Personally, I feel like this fix is the right way to go. Even if it only turns out to be short-term, it gets people building again. Once it's out we all can try and brainstorm if a better solution exists.

esteve:
Also, a nice bonus of this solution is that I think we'd no longer need to vendorize messages.

jhdcs:
[...]

PR #556, opened 2025-12-09
feat!: Re-Export Generated Messages in rclrs

maspe36:
## Summary

Working example of re-exporting of the generated interface crates found on the AMENT_PREFIX_PATH env var. With this change generated interface symbols are available directly from `rclrs::`, so for example, `rclrs::std_msgs::msg::String` is how you'd access the String type of the generated rust code for the std_msgs package. 

This depends on 
- https://github.com/ros2-rust/rosidl_rust/pull/12 for fully relocatable symbols.
- https://github.com/ros2-rust/cargo-ament-build/pull/35 to have generated message crates omitted from patching w/ colcon-ros-cargo
- https://github.com/ros2-rust/examples/pull/19
- https://github.com/ros2-rust/rosidl_rust/pull/21

The way we find the generated code is as follows
1. For all entries in `AMENT_PREFIX_PATH`, look for a `share/` directory.
2. For all entries in `share/` look for `<package>/rust/`
3. If there is a `Cargo.toml` within that entry, with the metadata to re-export symbols, `include!` all source files within `src/` such that all symbols resolve the same. 

i.e. `std_msgs::msg::String` for a standalone crate build, is effectively the same symbol as `rclrs::std_msgs::msg::String`.

I believe this to be a viable alternative to patching for message packages, and (could) effectively eliminate the need for vendoring some messages.

## Open questions
1. How does `AMENT_PREFIX_PATH` get populated? If its populated as colcon is building packages (which I believe to be the case), then this will also re-export symbols in your own workspace as well as sourced overlays. 

2. Should this actually go into `rclrs` or a separate crate (i.e. ros_msgs?) By going into rclrs we eliminate the need for vendoring.

3. Its important to know that if we re-build rclrs, and the `AMENT_PREFIX_PATH` env var has changed, then the build.rs will be re-built, which will then re-build all of the generated code. 

Honestly, I'm not sure if we can really get around this. Rust just really wants you to recompile the code you're using each time you invoke `cargo`. We can and should have everything in a colcon workspace share a [target directory](https://doc.rust-lang.org/cargo/reference/build-cache.html), but this won't stop all rebuilds. Its _possible_ we expect compiled `.rlib`s in the share directory and we generate a file which has `extern crate` calls to force rustc to find the dependencies, but this will likely not work with any intellisense, and its overall pretty fragile IMO.

4. Can we get better error messages? If a user forgot to source something or expects an interface to exists but one doesn't, they are met with errors like this
```rust

  error[E0433]: failed to resolve: could not find `action_msgs` in the crate root
   --> rclrs/src/action/action_client.rs:5:6
    |
  5 |     {action_msgs::srv::CancelGoal_Response, builtin_interfaces::msg::Time},
    |      ^^^^^^^^^^^ could not find `action_msgs` in the crate root
```


rclrs/build.rs
```diff
@@ -54,6 +71,91 @@ fn main() {
...
+
+            format!("pub mod {package} {{ {internal_mods} }}")
+        })
+        .collect();
+
+    let out_dir = env::var("OUT_DIR").expect("OUT_DIR not set ");
+    let dest_path = PathBuf::from(out_dir).join("interfaces.rs");
+
+    // TODO I would like to run rustfmt on this generated code, similar to how bindgen does it
+    fs::write(dest_path, content.clone()).expect("Failed to write interfaces.rs");
```
maspe36: interface.rs ends up looking like this

```rust

pub mod builtin_interfaces { ... };

pub mod std_msgs {
  pub mod msg {
    use crate::builtin_interfaces;
    // use crate::...; etc
    
    include!("path/to/generated/code");
    
    pub mod rmw {
      use crate::builtin_interfaces;
      // use crate::...; etc
    
      include!("path/to/generated/rmw/code");
    }
  }
}
```

Where all generated interface code is placed into a module based on the paths found in the `AMENT_PREFIX_PATH` share directories. 

--- 

Interestingly enough, I also learned that the `#[path = "..."]` [attribute for modules](https://doc.rust-lang.org/reference/items/modules.html#the-path-attribute) works with absolute paths. So that also could work, but I didn't have proper intellisense with that approach so I stuck with [include!()](https://doc.rust-lang.org/std/macro.include.html)

rclrs/src/lib.rs
```diff
@@ -205,6 +204,8 @@ mod test_helpers;
 
 mod rcl_bindings;
 
+include!(concat!(env!("OUT_DIR"), "/interfaces.rs"));
```
maspe36: This generates a _lot_ of warnings right now. Will need to look at how to hygienically do that (or move to another crate and suppress there)

maspe36:
@esteve I believe this is essentially an implementation of the thought you had back in https://github.com/ros2-rust/ros2_rust/issues/394. We could also use feature flags instead of embedding metadata in the generated crates like I did. That may cut down on some rebuilds?

EDIT: Future me, we can only use feature flags for base ROS 2 messages (i.e. `builtin_interfaces` etc), we can't use features for user defined message packages.

rclrs/build.rs
luca-della-vedova: Is there a way to change the location of the `interfaces.rs` to be a bit easier to find, perhaps in the install folder?
I find its current location in an obscure build subdirectory fairly confusing and it makes it hard to inspect what is going on:

```
lucadv@noble:~/ws_rclrs$ find . -name "interfaces.rs"
./src/ros2_rust/target/debug/build/rclrs-d6fb61407f6fcb8c/out/interfaces.rs
./src/ros2_rust/target/debug/build/rclrs-e547ca5eac3bba00/out/interfaces.rs
```

luca-della-vedova (commented):
Very interesting approach! And that's the kind of diff that gets me excited :smile:
I gave this a try and I have some comments.

First of all, [removing `colcon-ros-cargo`](https://github.com/ros2-rust/ros2_rust/pull/556/commits/448c44e13d47bc995cfac10d009b8cb9ddaf2621) to fix CI is _not_ a fix and actually breaks CI, so it should be reverted and the CI issues should be solved.
Removing `colcon-ros-cargo` will actually skip all the ros-cargo packages, including `rclrs`, if you look at the logs from the green CI you will notice that these warnings are printed:

```
[0.461s] WARNING:colcon.colcon_core.verb:No task extension to 'build' a 'ros.ament_cargo' package
```

And the build log shows that `rclrs` was actually not built and not tested, so that change effectively made the CI a noop.
There are two test failures, one is a very underwhelming `cargo fmt` run, the other is a failure to compile a unit test (seems to me that line was added by accident?):

```
error[E0423]: expected value, found struct `crate::rcl_interfaces::msg::IntegerRange`
   --> rclrs/src/subscription.rs:537:17
    |
537 |           let _ = crate::rcl_interfaces::msg::IntegerRange;
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use struct literal syntax instead: `crate::rcl_interfaces::msg::IntegerRange { from_value: val, to_value: val, step: val }`
    |
   ::: ~/ws_rclrs/install/rcl_interfaces/share/rcl_interfaces/rust/src/msg.rs:52:1
    |
52  | / pub struct IntegerRange {
53  | |     pub from_value: i64,
54  | |     pub to_value: i64,
55  | |     pub step: u64,
56  | | }
    | |_- `crate::rcl_interfaces::msg::IntegerRange` defined here

For more information about this error, try `rustc --explain E0423`.
error: could not compile `rclrs` (lib test) due to 1 previous error
```

Removing the line and running cargo fmt makes `colcon test` succeed on my machine.
Now to the questions

> ## Open questions
> 1. How does `AMENT_PREFIX_PATH` get populated? If its populated as colcon is building packages (which I believe to be the case), then this will also re-export symbols in your own workspace as well as sourced overlays.

It's... Interesting... I gave this a try and it seems that the content on the variable contains (together with the environment variable) the prefixes of the package that the current package depends on. My guess (and what I hope `colcon` would do) is that behind the scenes it looks for all the packages specified in `package.xml` and sources them before building the required package, so the prefix contains all the required dependencies but not more than that, which is pretty good behavior.

> 2. Should this actually go into `rclrs` or a separate crate (i.e. ros_msgs?) By going into rclrs we eliminate the need for vendoring.

I'm very partial about red diffs but what is imho even more great about bundling is removing these [confusing vendoring issues](https://github.com/ros2-rust/ros2_rust/issues/385).
A question still stands though, I thought some of the reason for vendoring was distributing the crate standalone on crates.io. How would we go about that with this approach?

> 3. Its important to know that if we re-build rclrs, and the `AMENT_PREFIX_PATH` env var has changed, then the build.rs will be re-built, which will then re-build all of the generated code.

Yeah this is not ideal, other people might disagree but personally I don't feel it is a definitive blocker, if the behavior is robust and developer friendly we can accept less than ideal compile times for now.
Probably a noob question but how does the script detect that the environment variable changed? It seems to work and I expected to see a [rerun-if-env-changed](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-env-changed) directive but I couldn't find any.

### Other general comments

I also tried to build the [message_demo example](https://github.com/ros2-rust/examples/tree/main/rclrs/message_demo) and noticed that this doesn't work. I think the issue is that you are not [including feature flags](https://github.com/ros2-rust/examples/blob/e7c18ef53a55081bdc489f6cf7217ee42f52f04c/rclrs/message_demo/Cargo.toml#L15) for message packages.

```
error[E0277]: the trait bound `rclrs::rclrs_example_msgs::msg::rmw::VariousTypes: serde::Deserialize<'de>` is not satisfied
    --> src/message_demo.rs:110:28
     |
110  |     let rmw_deserialized = serde_json::from_str(&rmw_serialized)?;
     |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `serde_core::de::Deserialize<'_>` is not implemented for `rclrs::rclrs_example_msgs::msg::rmw::VariousTypes`
     |
     = note: for local types consider adding `#[derive(serde::Deserialize)]` to your `rclrs::rclrs_example_msgs::msg::rmw::VariousTypes` type
```

And finally, I noticed that you implemented parsing of the environment variable into paths / splitting manually, have you considered using [ament_rs](https://github.com/ros2-rust/ament_rs/) to cut down on that boilerplate?

rclrs/build.rs
```diff
@@ -54,6 +71,91 @@ fn main() {
...
+                .to_str()
+                .unwrap();
+
+            // Find all dependencies for this crate that have a `*` version requirement.
+            // We will assume that these are other exported dependencies that need symbols
+            // exposed in their module.
+            let dependencies: String = Manifest::from_path(package_dir.clone().join("Cargo.toml"))
+                .iter()
+                .map(star_deps_to_use)
+                .collect();
```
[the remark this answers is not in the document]
maspe36: > require message packages to always have a * version requirement? 

Not necessarily, this is just what I did in the first pass as this is actually how we're specifying dependencies in the generated crates. We could evolve this to be an intersection of dependencies found in both the Cargo.toml and package.xml, or perhaps some other mechanism like metadata in the Cargo.toml.

> On the other hand, how do we make sure we are only re-exporting message packages? 

I don't think we really can fully prevent this because we do want users to be able to generate and install their own rust crates representing their custom message packages. 

> What happens if a user has a Rust library installed instead?

I have gone ahead and removed the logic to create the marker file for crates with the expected metadata. `colcon-ros-cargo` won't find the generated rust crates for message packages anymore.
https://github.com/ros2-rust/cargo-ament-build/pull/35

If we decide on another approach then this will need to be revisited.

maspe36:
> First of all, [removing colcon-ros-cargo](https://github.com/ros2-rust/ros2_rust/pull/556/commits/448c44e13d47bc995cfac10d009b8cb9ddaf2621) to fix CI is not a fix and actually breaks CI

Yup, fixed this by making a change to `cargo-ament-build` to no longer install the `rust_packages` marker file for crates that have this reexport metadata
https://github.com/ros2-rust/cargo-ament-build/pull/35

Verified that `colcon build` and `colcon test` now actually are running :D 

> Removing the line and running cargo fmt makes colcon test succeed on my machine.

Ah yeah, whoops. Fixed!

> I thought some of the reason for vendoring was distributing the crate standalone on crates.io. How would we go about that with this approach?

This is an extremely valid point, thank you for bringing this up. I would propose supplying the absolute bare minimum required to satisfy compilation and having that be conditionally included if we're doing a build for docs.rs. I'll see how feasible this really is.

> Probably a noob question but how does the script detect that the environment variable changed? 

Not a noob question! In my early changes I was using `env!`, and by simply using the `env!` (or `option_env!`) macro build scripts will automatically rerun when the value is changed. I have since moved away from the macro as a first step towards being able to build for docs.rs.

> As of 1.46, using [env!](https://doc.rust-lang.org/std/macro.env.html) and [option_env!](https://doc.rust-lang.org/std/macro.option_env.html) in source code will automatically detect changes and trigger rebuilds. rerun-if-env-changed is no longer needed for variables already referenced by these macros.

https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-env-changed

> I also tried to build the [message_demo example](https://github.com/ros2-rust/examples/tree/main/rclrs/message_demo) and noticed that this doesn't work. I think the issue is that you are not [including feature flags](https://github.com/ros2-rust/examples/blob/e7c18ef53a55081bdc489f6cf7217ee42f52f04c/rclrs/message_demo/Cargo.toml#L15) for message packages.

We need the `serde` feature on for `rclrs` in that example I believe. Enabled the feature, and everything built (well, with warnings still). Went ahead and put up a PR with the required changes.
https://github.com/ros2-rust/examples/pull/19

> And finally, I noticed that you implemented parsing of the environment variable into paths / splitting manually, have you considered using [ament_rs](https://github.com/ros2-rust/ament_rs/) to cut down on that boilerplate?

I didn't find much value from this crate honestly, but I have gone ahead and used it. I could see improvements being made to it in the future.

maspe36:
I have enabled cargo docs builds in https://github.com/ros2-rust/ros2_rust/pull/556/commits/1ff04a1f5409062cdfe8c62f1978c7a7a556fa8b

I believe all outstanding comments have been fully addressed now

mxgrey:
I haven't given this a test run yet, but conceptually the more I think about this approach the more I like it. It resolves every technical issue that I can think of, and doesn't introduce any new problems besides just being unconventional (or you could also say creative or resourceful).

> Should this actually go into rclrs or a separate crate (i.e. ros_msgs?) By going into rclrs we eliminate the need for vendoring.

This would be my only feedback at this time. I believe very strongly that this should go into its own crate, maybe named `ros_messages` or `rosidl` (I would avoid `ros_msgs` because too many message packages end with `_msgs`, so the name might make users think this is actually a message package itself).

As a separate crate it would be feasible for r2r or other Rust-based ROS client library implementations to share the same message package pipeline. I think that would be a huge win for the ecosystem. As far as I can figure this would still fix the need for vendoring messages within rclrs.

knmcguire:
About the failing test, if you'd sync the branch with main than it should be resolved

maspe36:
> I believe very strongly that this should go into its own crate, maybe named `ros_messages` or `rosidl` (I would avoid `ros_msgs` because too many message packages end with `_msgs`, so the name might make users think this is actually a message package itself).

I think this would be a positive change. We can discuss more tomorrow in the WG meeting, but I am thinking perhaps something like `ros_registry` or maybe `rosidl_registry` to hopefully keep the name on the shorter end.

EDIT: I have done some noodling and gone ahead with [ros-env](https://github.com/ros2-rust/ros-env)

esteve:
@maspe36 I've triggered another CI build for Humble on Linux. I don't know why the Windows build fails, seems related to the new `ros-env` crate. Do you know why it's failing?

maspe36:
@esteve I don't know about the previous failure, but the humble one was due to an outdated `.repos` file. Now the windows one is failing due to windows paths escaping rust strings (classic). I'll get a fix out for `ros-env` and link here.

.github/workflows/rust-minimal.yml
```diff
@@ -69,11 +69,12 @@ jobs:
 
     # Colcon can not be run in a venv which is required in Ubuntu Noble
     # Removing the externally managed file
-    - name: Install colcon-cargo and colcon-ros-cargo
+    - name: Install colcon-cargo, colcon-ros-cargo, and cargo-ament-build
       run: |
         sudo rm -f /usr/lib/python3.12/EXTERNALLY-MANAGED
         sudo pip3 install git+https://github.com/colcon/colcon-cargo.git
         sudo pip3 install git+https://github.com/colcon/colcon-ros-cargo.git
+        cargo install --git https://git@github.com/maspe36/cargo-ament-build.git --branch feature/skip_reexport_marker_file
```
esteve: This will need to be removed before merging. Unfortunately a new release of `cargo-ament-build` depends on https://github.com/ros2-rust/cargo-ament-build/pull/37

esteve (changes requested):

.github/workflows/rust-minimal.yml
esteve: There's a new release of the `cargo-ament-build` plugin (https://crates.io/crates/cargo-ament-build). We now also have Debian packages (https://github.com/ros2-rust/cargo-ament-build/releases/tag/v0.1.11)

esteve:
```suggestion
```

.github/workflows/rust-stable.yml
```diff
@@ -69,11 +69,12 @@ jobs:
 
     # Colcon can not be run in a venv which is required in Ubuntu Noble
     # Removing the externally managed file
-    - name: Install colcon-cargo and colcon-ros-cargo
+    - name: Install colcon-cargo, colcon-ros-cargo, and cargo-ament-build
       run: |
         sudo rm -f /usr/lib/python3.12/EXTERNALLY-MANAGED
         sudo pip3 install git+https://github.com/colcon/colcon-cargo.git
         sudo pip3 install git+https://github.com/colcon/colcon-ros-cargo.git
+        cargo install --git https://git@github.com/maspe36/cargo-ament-build.git --branch feature/skip_reexport_marker_file
```
esteve:
```suggestion
```

.github/workflows/rust-win.yml
```diff
@@ -40,6 +40,8 @@ jobs:
         # prerequisites and fixes for windows build ros2_rust:
         #     * Libclang has to be added (from the ros2_rust instructions) and the dll has to be renamed
         #     * colcon-ros-cargo and colcon-cargo have to be added as PyPI packages
+        #     * cargo-ament-build is manually installed from github for now to include the change to skip
+        #         installing the marker file if the package should be reexported
```
esteve:
```suggestion
```

.github/workflows/rust-win.yml
```diff
@@ -48,6 +50,7 @@ jobs:
           pixi add --pypi "colcon-ros-cargo@git+https://github.com/colcon/colcon-ros-cargo.git" --manifest-path C:\pixi_ws\pixi.toml
           pixi add --pypi "colcon-cargo@git+https://github.com/colcon/colcon-cargo.git" --manifest-path C:\pixi_ws\pixi.toml
           pixi upgrade colcon-core --manifest-path C:\pixi_ws\pixi.toml
+          cargo install --git https://git@github.com/maspe36/cargo-ament-build.git --branch feature/skip_reexport_marker_file
```
esteve:
```suggestion
```

.github/workflows/rust-minimal.yml
```diff
@@ -69,11 +69,12 @@ jobs:
 
     # Colcon can not be run in a venv which is required in Ubuntu Noble
     # Removing the externally managed file
-    - name: Install colcon-cargo and colcon-ros-cargo
+    - name: Install colcon-cargo, colcon-ros-cargo, and cargo-ament-build
```
esteve:
```suggestion
    - name: Install colcon-cargo and colcon-ros-cargo
```

.github/workflows/rust-stable.yml
```diff
@@ -69,11 +69,12 @@ jobs:
 
     # Colcon can not be run in a venv which is required in Ubuntu Noble
     # Removing the externally managed file
-    - name: Install colcon-cargo and colcon-ros-cargo
+    - name: Install colcon-cargo, colcon-ros-cargo, and cargo-ament-build
```
esteve:
```suggestion
    - name: Install colcon-cargo and colcon-ros-cargo
```

esteve (changes requested):
@maspe36 ping 🙂 I've added a bunch of suggestions reverting the changes to CI now that we have a new release of `cargo-ament-build` with the marker changes.

esteve:
@maspe36 I've rebased this PR on top of `main` so that it'd pick up the CI workaround for `ament_mypy`, I hope that's ok with you.

esteve (approved):
It's ok if rolling fails for now, CI for the other supported distros pass.

PR #569, opened 2026-01-12
Temporary url for the windows binary download in CI

knmcguire:
Until the nightly binary releases of windows binaries on rolling have a fixed place to be found.


knmcguire:
CI has been going for 10 minutes so that means that it retrieved the binary from the temporary source :)

knmcguire:
The CI now hits the same issue as with the minimal build. I'll propose an update to pixi.toml to the ros2/ros2 repo

PR #591, opened 2026-01-24
feat(dynamic_message): add DynamicSequence length helpers

bresilla:
## What

Adds small ergonomic helpers to dynamic message sequences:
- `len()` / `is_empty()` for `DynamicSequence`, `DynamicBoundedSequence`, `DynamicSequenceMut`, `DynamicBoundedSequenceMut`
- `iter()` for immutable sequences

## Why

Makes it easier to traverse dynamic message sequences (especially sequences of nested messages) without reaching into internal storage.

## Notes

- No behavior change; purely additive API.
- `cargo test` passes.

### DIFF
--- a/rclrs/src/dynamic_message/field_access/dynamic_sequence.rs
+++ b/rclrs/src/dynamic_message/field_access/dynamic_sequence.rs
@@ -1,6 +1,8 @@
-use std::fmt::{self, Debug};
-use std::marker::PhantomData;
-use std::ops::{Deref, DerefMut};
+use std::{
+    fmt::{self, Debug},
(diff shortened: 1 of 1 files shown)


luca-della-vedova (commented):
I _believe_ but please correct me if I'm wrong that since all the types implement `Deref` which just returns the inner slice, we can just deref the sequences through `*` then we should be able to use all of Rust's slice APIs (i.e. [is_empty](https://doc.rust-lang.org/std/primitive.slice.html#method.is_empty)) so we wouldn't need to implement any ourselves

PR #592, opened 2026-01-24
feat: add serialized pub/sub APIs

bresilla:
## What

- Re-export `rcl_bindings` (`pub mod rcl_bindings`) to enable low-level extensions.
- Add first-class serialized messaging types:
  - `SerializedMessage`
  - `SerializedSubscription`
  - `SerializedPublisher`
- Add `Node::create_serialized_subscription(...)` and `Node::create_serialized_publisher(...)`.

## Why

Enables native tools (e.g. rosbag MCAP record/play) to work with raw serialized CDR bytes without needing generated Rust message crates.

## Notes

- Uses existing dynamic message typesupport resolution (`DynamicMessageMetadata`) so it works for any installed message type.
- Does not change existing typed or dynamic pub/sub behavior.


mxgrey:
Is this possibly a duplicate of the feature introduced by #492 ?

luca-della-vedova (commented):
I haven't done Rust in a few months (life!) but I have a few comments

bresilla:
@luca-della-vedova Yes, that makes sense conceptually. The serialized and dynamic paths already share the same runtime type-support lookup, and the main behavioral difference is whether the publish/take path goes through `rcl_publish`/dynamic decoding or `rcl_publish_serialized_message`/`rcl_take_serialized_message` on raw bytes. I kept this version separate because I wanted the smallest API that exposes serialized transport directly for tool-style use cases like rosbag/MCAP, without first redesigning the dynamic pub/sub abstractions. If maintainers prefer, I can rework this toward a shared abstraction, but I was treating that as a follow-up refactor rather than a prerequisite for exposing the capability at all.

bresilla:
@mxgrey Not a duplicate of #492. #492 adds dynamic pub/sub where the payload is decoded into `DynamicMessage` and delivered via the dynamic message path. This PR adds serialized pub/sub for raw CDR bytes, while reusing the same runtime type-support lookup. So they are related, but the use case here is different: this is meant for tool-style scenarios like rosbag/MCAP record/play where I need raw serialized transport rather than decoded dynamic message views.

Issue #590, opened 2026-01-24
Kilted: log!() macros do not publish to "/rosout" topic


EmeraldWither:
Hello!

When using
```rust
log!(node.error(), "This is a log message");
```
it is expected that a message will be published to the `/rosout` topic with the message contents, however, this behavior does not appear to exist.

I confirmed it by running the minimal project below, while having `$ ros topic echo /rosout` active in another terminal session, alongside having another package subscribed to the `/rosout` topic, and verifying that nothing is published to `/rosout` on a log message. 

Here is a minimal project that I setup to test this: 

`src/minimal_logger.rs`
```rust
use anyhow::{Error, Result};
use rclrs::*;

fn main() -> Result<(), Error> {
    let context = Context::default_from_env()?;
    let mut executor = context.create_basic_executor();
    let node = executor.create_node("log_test_node")?;
    log!(node.error(), "This is a log message!");
    executor.spin(SpinOptions::default());
    Ok(())
}
```

`cargo.toml`
```toml
[package]
name = "minimal_pub_sub_rust"
version = "0.5.0"
edition = "2021"
[[bin]]
name = "minimal_logger"
path = "src/minimal_logger.rs"

[dependencies]
anyhow = {version = "1", features = ["backtrace"]}
rclrs = "0.7.0"
```

`package.xml`
```xml
<?xml version="1.0"?>
<?xml-model
   href="http://download.ros.org/schema/package_format3.xsd"
   schematypens="http://www.w3.org/2001/XMLSchema"?>
<package format="3">
  <name>minimal_pub_sub_rust</name>
  <version>0.5.0</version>
  <description>Package</description>
  <license>Apache License 2.0</license>
  <maintainer email="68785503+EmeraldWither@users.noreply.github.com">EmeraldWither</maintainer>

  <build_depend>rclrs</build_depend>
  <exec_depend>rclrs</exec_depend>

  <export>
    <build_type>ament_cargo</build_type>
  </export>
</package>
```

esteve:
@EmeraldWither thanks for filing this ticket, it seems that distros after Humble changed the behavior of the `rosout` initialization, I've submitted https://github.com/ros2-rust/ros2_rust/pull/593, can you check if it fixes the issues for you? Thanks.

PR #593, opened 2026-01-28
fix: initialize rosout publisher on distros newer than Humble

esteve:
This PR initializes the `rosout` publisher for distros after Humble.

Fixes https://github.com/ros2-rust/ros2_rust/issues/590

### DIFF
--- a/rclrs/src/logging.rs
+++ b/rclrs/src/logging.rs
@@ -667,4 +667,205 @@ mod tests {
     fn test_function_macro() {
         assert_eq!(function!(), "rclrs::logging::tests::test_function_macro");
     }
+
+    #[test]
+    fn test_rosout_publishing_default() -> Result<(), RclrsError> {
+        use crate::{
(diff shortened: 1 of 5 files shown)


esteve:
@EmeraldWither friendly ping :slightly_smiling_face: Can you check if this change fixes your issue with `rosout`? I believe the test covers it and I've been able to reproduce it for distros newer than Humble. Thanks.

EmeraldWither:
Hi, thanks for taking a look at this. 

This change does work on Kilted, and using the following code snippet does result in the expected behavior: 
```rust
use anyhow::{Error, Result};
use rclrs::*;

    let context = Context::default_from_env()?;
    let mut executor = context.create_basic_executor();
    let node = executor.create_node("log_test_node")?;
    log!(node.error(), "This is a log message!");
    executor.spin(SpinOptions::default());
    Ok(())
}
```

esteve:
@EmeraldWither thanks for the feedback :slightly_smiling_face:

PR #609, opened 2026-02-17
Replace try_next with try_recv

balthasarschuess:
Usage of try_next is deprecated in futures crate.
This causes issues for users with `-Dwarnings` enabled.


mxgrey (approved):
Thanks for updating this, looks good!

esteve (approved):
@balthasarschuess thanks!

PR #610, opened 2026-02-18
docs: add missing cargo-ament-build to installation prerequisites

lncosinx:
Description:
The current installation instructions missed the cargo-ament-build dependency. Without installing it globally via cargo install cargo-ament-build, running colcon build will fail with a fatal error stating "cargo ament-build was not detected".

Changes:
Added cargo install cargo-ament-build to the prerequisites block in README.md.
Added the same command to the "Option 1: Installing the dependencies" section in docs/building.md.

This will help new users successfully build the workspace without encountering the missing cargo plugin error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,9 @@ sudo apt install -y git libclang-dev python3-pip python3-vcstool
 
 # Install colcon plugins for Rust
 pip install --break-system-packages colcon-cargo colcon-ros-cargo
+
+# Install cargo plugin for ament
+cargo install cargo-ament-build
 ```
 
 Because of an issue in `rclrs` (https://github.com/ros2-rust/ros2_rust/issues/557), as a workaround, the following two packages need to be installed:
--- a/docs/building.md
+++ b/docs/building.md
@@ -45,6 +45,7 @@ sudo apt install -y git libclang-dev python3-pip python3-vcstool # libclang-dev 
 # Install these plugins for cargo and colcon:
 pip install git+https://github.com/colcon/colcon-cargo.git
 pip install git+https://github.com/colcon/colcon-ros-cargo.git
+cargo install cargo-ament-build
 ```
 
 ### Option 2: Using the Docker image


esteve:
@lncosinx the `colcon-ros-cargo` PyPI package has a dependency on the `cargo-ament-build` package (see https://github.com/colcon/colcon-ros-cargo/blob/main/setup.cfg#L30). Did you install `colcon-ros-cargo` via PyPI? Could you please paste the output of `pip show colcon-ros-cargo` here? Thanks.

lncosinx:
> @lncosinx the `colcon-ros-cargo` PyPI package has a dependency on the `cargo-ament-build` package (see https://github.com/colcon/colcon-ros-cargo/blob/main/setup.cfg#L30). Did you install `colcon-ros-cargo` via PyPI? Could you please paste the output of `pip show colcon-ros-cargo` here? Thanks.

Hi @esteve, thanks for the review and for pointing that out!
You are completely right. Here is the output of pip show colcon-ros-cargo:

Plaintext
Name: colcon-ros-cargo
Version: 0.2.0
Summary: Extension for colcon to support Ament Cargo packages.
Location: /home/lncosinx/.local/lib/python3.10/site-packages
Requires: cargo-ament-build, colcon-cargo, colcon-core, colcon-library-path, colcon-ros, toml

So pip did indeed install cargo-ament-build as a dependency.
However, during colcon build, the build system couldn't find it. This was likely because pip installed the executable into my ~/.local/bin directory, which wasn't in my active $PATH at the time. Because it wasn't detected, the ament_cargo CMake script threw this specific fatal error:

ament_cargo package found but cargo ament-build was not detected. Please install it by running: $ cargo install cargo-ament-build

Since the error message generated by the build system explicitly suggests using cargo install cargo-ament-build, I followed that instruction (which put the binary in ~/.cargo/bin and fixed the build). That's why I proposed adding it to the documentation.

Given that this was ultimately a $PATH issue on my end (missing ~/.local/bin) combined with a potentially misleading error message from the build script, how would you like to proceed?

1. Close this PR, as the PyPI dependency handles it automatically.

2. Modify this PR to instead add a tip about ensuring ~/.local/bin is in the $PATH when installing via pip.

I'm completely happy to close the PR or update it based on your preference!

PR #611, opened 2026-03-08
feat!: add parameter callbacks

azerupi:
Adds three new capabilities to the parameter API:

1. `ParameterBuilder::validate()` to register a callback that can reject parameter changes. Called during declaration (for the initial value) and on every subsequent set, both programmatic and via service calls. Runs after range checks, before the value is written.
2. `MandatoryParameter::on_change()` / `OptionalParameter::on_change()` to register a callback invoked after a parameter value is successfully changed. Intended for side effects (reconfiguring algorithms, logging, etc.). Callbacks are invoked outside the parameter map lock to avoid deadlocks.
3. `MandatoryParameter::subscribe()` / `OptionalParameter::subscribe()` that return a `ParameterSubscription` for async notification of changes via `tokio::sync::watch`. Supports `changed()`, `get()`, and `wait_for(predicate)`. The naming might be slightly confusing with subscribers already being a foundational ROS 2 concept, let me know if there are better naming ideas.

Also adds `ValidationFailed` and `InitialValueRejected` error variants to `ParameterValueError` and `DeclarationError` respectively.

Examples can be found here: https://github.com/azerupi/ros2_rust_examples/tree/parameter_callback_v2

---- 
***Disclaimer** AI was used to help write this PR*

### DIFF
--- a/rclrs/src/parameter.rs
+++ b/rclrs/src/parameter.rs
@@ -19,6 +19,7 @@ use std::{
     marker::PhantomData,
     sync::{Arc, Mutex, RwLock, Weak},
 };
+use tokio::sync::watch;
 
 // This module implements the core logic of parameters in rclrs.
 // The implementation is fairly different from the existing ROS 2 client libraries. A detailed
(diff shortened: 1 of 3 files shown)


esteve (approved):
@azerupi awesome work, thanks!

PR #612, opened 2026-03-09
Add structured parameters

azerupi:
This is an alternative implementation of #516.

  The goal of this PR is **not** to dismiss or negate the work done in #516. But while reviewing that PR, because I'm excited about this feature, I felt like there was a better approach and it felt easier to explore an alternative design by trying another implementation from scratch (with AI assistance) rather than posting a bunch of review comments on the initial PR. 

Anyone working on or reviewing #516 should feel free to dismiss this PR entirely or incorporate any ideas from it into the original PR if desired.

  ### Key differences from #516

| | #516 (`StructuredParameters`) | This PR (`ParameterSet`) |
|---|---|---|
| **Code generation approach** | Trait-based indirection — the macro generates a `StructuredParametersMeta<DefaultForbidden>` trait impl that calls `declare_structured_()` passing all builder options as function arguments. The actual builder logic lives in hand-written trait impls on `MandatoryParameter<T>`, `OptionalParameter<T>`, etc. | Direct code generation — the macro emits `ParameterBuilder` method chains (`.default().description().mandatory()`) directly, with no intermediate traits. |
| **Intermediate types** | Requires a `DefaultForbidden` marker type with `unreachable!()` impls to satisfy the type system for nested structs. | No intermediate types — nested structs are detected by type name and generate recursive `ParameterSet::declare()` calls. |
| **Leaf vs nested detection** | No distinction — all fields (leaf and nested) use the same `declare_structured()` call with all builder options passed as arguments. For nested structs, options like `default`, `description`, `range` are silently ignored. Leaf-only attributes on nested fields are not caught at compile time. | The macro distinguishes leaf from nested at expansion time and generates different code for each. Leaf-only attributes on nested fields produce compile errors. |
| **Namespacing** | Parameter names are composed from field names and an optional user-provided prefix. E.g. `node.declare_parameters("")` on a `Robot` struct with a nested `sensors: SensorConfig` field produces `speed`, `sensors.rate`. With `node.declare_parameters("robot")`: `robot._speed`, `robot._sensors._rate`. | The top-level struct name is converted to snake_case and used as namespace prefix by default. `node.declare_parameter_set::<Robot>()` produces `robot.speed`, `robot.sensors.rate`. Overridable with `#[parameters(namespace = "...")]` or `#[parameters(flatten)]`. |
| **Flatten** | Not supported. | `#[param(flatten)]` on fields skips the field name in the namespace; `#[parameters(flatten)]` on structs skips the struct name. Flatten is not recursive — it only affects the level it's applied to. |
| **Range syntax** | `range = rclrs::ParameterRange { lower: Some(1.0), ..Default::default() }` | `range(lower = 1.0)` — parsed into `ParameterRange` by the macro. |
| **Macro crate structure** | Single `impl.rs` file (~160 lines). | Modular: `struct_attrs.rs`, `field_attrs.rs`, `field_info.rs`, `codegen.rs` + `mod.rs` (~550 lines). |
| **rclrs-macros dependency** | Added as `dev-dependencies` in rclrs and not re-exported, the derive macro is only usable by adding the proc-macro crate as a dependency in downstream crates. | Added as `dependencies` and re-exported via `pub use rclrs_macros::ParameterSet` — downstream crates use `#[derive(ParameterSet)]` from `rclrs` directly |

  ### What this PR adds

  - **`rclrs-macros` crate** — proc-macro crate with `#[derive(ParameterSet)]`
  - **`ParameterSet` trait** in rclrs with `default_namespace()` and `declare()` methods
  - **`NodeState` convenience methods** — `declare_parameter_set()` and `declare_parameter_set_with_prefix()`
  - **Struct-level attributes** — `#[parameters(flatten)]`, `#[parameters(namespace = "...")]`
  - **Field-level attributes** — `#[param(default, description, constraints, range(...), ignore_override, discard_mismatching_prior_value,
  discriminate, mandatory, optional, read_only, flatten)]`
  - **10 integration tests** covering all features

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,6 @@
 [workspace]
 members = [
   "rclrs",
+  "rclrs-macros",
 ]
 resolver = "2"
--- a/rclrs-macros/Cargo.toml
+++ b/rclrs-macros/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "rclrs-macros"
+version = "0.1.0"
+edition = "2021"
+license = "Apache-2.0"
+description = "Procedural macros for the rclrs ROS 2 client library"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+proc-macro2 = "1.0"
+quote = "1.0"
+syn = { version = "2.0", features = ["full", "extra-traits"] }
--- a/rclrs-macros/package.xml
+++ b/rclrs-macros/package.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0"?>
+<?xml-model
+   href="http://download.ros.org/schema/package_format3.xsd"
+   schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>rclrs-macros</name>
+  <version>0.5.0</version>
+  <description>Procedural macros for the rclrs ROS 2 client library.</description>
+  <maintainer email="esteve@apache.org">Esteve Fernandez</maintainer>
+  <license>Apache License 2.0</license>
+  <export>
+    <build_type>ament_cargo</build_type>
+  </export>
+</package>
--- a/rclrs-macros/src/lib.rs
+++ b/rclrs-macros/src/lib.rs
@@ -1,0 +1,15 @@
+use proc_macro::TokenStream;
+
+/// Derive macro for declaring ROS 2 parameters from struct definitions.
+///
+/// See `rclrs::ParameterSet` for full documentation.
+#[proc_macro_derive(ParameterSet, attributes(parameters, param))]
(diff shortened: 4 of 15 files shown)


balthasarschuess:
I think we should add a blanket implementation for ParameterVariant for serde::ser::Serialize and serde::de::Deserialize we should guard it behind a feature gate though to not pull in the dependency for all users. 

But that could also happen in a subsequent PR

azerupi:
I have a better approach to this I think, similar direction but with many improvements. I've done some exploration with AI and ended up with this design if you are interesed: https://github.com/azerupi/ros2_rust/tree/params/e-keyed-maps

I have it structured in a few PRs that build on each other but I have to go through each of them and clean them up before I can submit them.

PR #618, opened 2026-03-27
Interface for serialized messages

atimin:
We're developing a bridge that will receive and store serialised data in its original form in a database. This patch enables that functionality. I didn't put much effort into it as it's generated code. However, if the team sees value in this feature, I can develop it further.  

Thank you for the awesome library!


atimin:
Looks like there is another PR doing the same: https://github.com/ros2-rust/ros2_rust/pull/592.   Please, ignore this one.

esteve:
@atimin no problem! Were you able to try the approach in #592 ? Does it work for you?

atimin:
Hey @esteve, I tried to adopt the code from #592. It doesn't work in my environment. I need more time to investigate why. It may be the way I've integrated it. 

However, I can see that my patch adheres to the project API style and uses callbacks. I've already tested it in a production environment, processing thousands of messages per second with no issues. Would you prefer to use my work? 

In any case, I'm open to collaborating on both approaches.

PR #622, opened 2026-04-10
feat: add server_is_available to ActionClientState

patrick-deschambault:
## Description
This PR adds the `server_is_available` method to `ActionClientState`. 

Currently, `rclrs` action clients does not have a built-in mechanism to verify the availability of an action server on the network. This makes it difficult to safely send goals upon initialization without risking dropped requests.

By exposing the `rcl_action_server_is_available` binding, this PR enables developers to easily build reliable waiting logic. 

## Example Use Case
This enables the implementation of async timeout loops in user code, such as:

```rust
async fn wait_for_server(&self, timeout: std::time::Duration) -> Result<(), rclrs::RclrsError> {
    tokio::time::timeout(timeout, async {
        loop {
            if self.client.server_is_available()? {
                return Ok(());
            }
            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
        }
    }).await.unwrap_or(Err(rclrs::RclrsError::Custom("Timeout".into())))
}

### DIFF
--- a/rclrs/src/action/action_client.rs
+++ b/rclrs/src/action/action_client.rs
@@ -169,6 +169,34 @@ pub struct ActionClientState<A: Action> {
 }
 
 impl<A: Action> ActionClientState<A> {
+    /// Checks if the action server is available and ready to receive requests.
+    ///
+    /// This method polls the underlying DDS network to determine if the action
+    /// server associated with this client is currently active. This is particularly
+    /// useful for implementing a `wait_for_server` timeout loop before dispatching goals.
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`RclrsError`] if there is an issue communicating with the
+    /// underlying `rcl` C layer.
+    pub fn server_is_available(&self) -> Result<bool, RclrsError> {
+        let mut is_available = false;
+        
+        unsafe {
+            let handle = self.board.handle.lock();
+            let node_handle = self.board.handle.node_handle.rcl_node.lock().unwrap();
+            
+            rcl_action_server_is_available(
+                &*node_handle,
+                &*handle,
+                &mut is_available
+            )
+        }
+        .ok()?;
+
+        Ok(is_available)
+    }
+
     /// Request the action server to execute a goal. You will receive a
     /// [`RequestedGoalClient`] which you can use to wait for the reply from the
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: ros2_rust_humble.repos [hunk 1]
```diff
@@ -23,6 +23,10 @@ repositories:
     type: git
     url: https://github.com/ros2/unique_identifier_msgs.git
     version: humble
+  ros2/geometry2:
```
esteve: @patrick-deschambault thanks for the PR, can you elaborate why you needed this? Could you remove it if you don't need it? Thanks.

esteve (changes requested):

ros2_rust_humble.repos
patrick-deschambault: My mistake ! Should be erased ( 3cf276ece1a7460b5afe1ce260c5ba273c23a081 )

esteve:
@patrick-deschambault CI is failing because the code is not formatted properly, could you run cargo fmt and fix the warnings? Thanks

rclrs/src/action/action_client.rs
```diff
@@ -169,6 +169,34 @@ pub struct ActionClientState<A: Action> {
...
+    ///
+    /// This method polls the underlying DDS network to determine if the action
+    /// server associated with this client is currently active. This is particularly
+    /// useful for implementing a `wait_for_server` timeout loop before dispatching goals.
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`RclrsError`] if there is an issue communicating with the
+    /// underlying `rcl` C layer.
+    pub fn server_is_available(&self) -> Result<bool, RclrsError> {
```
esteve, line 182: It'd be really useful to provide tests for this function, could you add them?

esteve (changes requested):

rclrs/src/action/action_client.rs
patrick-deschambault: Yes sure, I'll put the PR on Draft and bring it Under Review when ready. Thanks for your input !

esteve: Thanks!

esteve (approved):

PR #623, opened 2026-04-11
Windows ros env test

maspe36:


maspe36:
Not 100% sure what the issue is with my approach for regular ubuntu builds, but I was able to verify that window builds work now and locally I can still build on ubuntu. I will release a new version of ros-env to crates.io that fixes the build for https://github.com/ros2-rust/ros2_rust/pull/556

PR #624, opened 2026-04-12
fix: gate deprecated liveliness variant by distro (https://github.com/ros2/rmw/pull/414)

esteve:
This PR removes `RMW_QOS_POLICY_LIVELINESS_MANUAL_BY_NODE` from the list of available QoS options for any distro that's not Humble, Jazzy or Kilted (i.e. Lyrical, Rolling and newer).

See https://github.com/ros2/rmw/pull/414


maspe36 (approved):
LGTM

PR #626, opened 2026-04-15
fix(logging): avoid overriding global rcutils log handler

esteve:
This PR makes the logging tests run sequentially. Logging in ROS is not thread-safe as it uses global state, when `cargo test` runs these tests in parallel, sometimes cargo crashes with a SIGSEGV.

### DIFF
--- a/rclrs/src/logging.rs
+++ b/rclrs/src/logging.rs
@@ -335,56 +335,36 @@ pub unsafe fn impl_log(
         static FORMAT_STRING: OnceLock<CString> = OnceLock::new();
         let format_string = FORMAT_STRING.get_or_init(|| CString::new("%s").unwrap());
 
-        let severity = severity.as_native();
+        let severity = severity.as_native() as i32;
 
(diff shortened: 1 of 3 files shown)


esteve:
Perhaps a more complete approach would be to protect such global state inside `rclrs`, but these changes should suffice.

esteve:
The issues with the rolling job are caused by `action-ros-ci`, I think the APT repos are not updated with the latest rolling sync.

esteve:
Here's an example of a failing CI job crashing:

https://github.com/ros2-rust/ros2_rust/actions/runs/24464336823/job/71487081210?pr=625

```
error: test failed, to rerun pass `-p rclrs --lib`
  
  Caused by:
    process didn't exit successfully: `/__w/ros2_rust/ros2_rust/ros_ws/build/rclrs/debug/deps/rclrs-051c2c557ee4ff13 --color=never --quiet` (signal: 11, SIGSEGV: invalid memory reference)
```

PR #630, opened 2026-04-18
fix: correct pointer dereference in resize_unchecked function

TonyWelte:
# Description

Resizing a dynamic sequence would crash with the following error
```
free(): invalid pointer
```

This PR fixes resizing dynamic sequences and adds tests 

# Disclaimer

Github Copilot was used to generate the tests and then to find the issue and fix it


esteve (approved):
@TonyWelte thanks!

PR #635, opened 2026-05-07
fix: install ament_cmake_mypy for rolling CI builds

esteve:
test_msgs from rcl_interfaces (rolling) recently added ament_mypy as a test dependency in its CMakeLists.txt, but rosdep may fail to install it when building from source. Explicitly install the package before the build step to avoid 'Unknown CMake command ament_mypy' errors.


esteve:
Merging as a workaround so that other PRs can pass

PR #649, opened 2026-06-14
feat: re-export rosidl_runtime_rs from rclrs

esteve:


mxgrey (approved):
This is a good idea. It can save dependency management trouble for downstream packages.

PR #650, opened 2026-06-14
ci: pin Windows runner to VS 2022

esteve:
See https://github.com/colcon/colcon-cmake/pull/161 and https://github.blog/changelog/2026-02-05-github-actions-early-february-2026-updates/#windows-server-2025-with-visual-studio-2026-image-now-available-for-github-hosted-runners

### DIFF
--- a/.github/workflows/rust-win.yml
+++ b/.github/workflows/rust-win.yml
@@ -16,7 +16,8 @@ env:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


esteve:
The rolling issues in CI are fixed in #641 , but this PR needs to be merged first to make sure that the Windows CI job can build.

PR #651, opened 2026-06-20
fix: action server clock use-after-free

azerupi:
While working on something else I discovered the action server had two bugs related to teardown. These bugs have been narrowed down and fixed using Claude code. 

### Bug 1: Action server dereferences a freed clock on goal teardown

`rcl_action_server_init` is given a raw pointer to the node's `rcl_clock` for the action server's expired-goals timer. `rcl_action_notify_goal_done`, called whenever a goal terminates or is dropped, dereferences that clock.

But `ActionServerHandle` only kept the node handle alive, not the clock (which is owned by the node's `NodeState`). A goal holds `Arc<ActionServerHandle>`, so it kept the `rcl_action_server` and `rcl_node` alive but not the clock. When a node's `NodeState` is dropped while a goal still lives (e.g. a goal-handler future still being torn down at executor/context shutdown), the clock is freed and the goal's `Drop` calls `rcl_action_notify_goal_done` against it.

**Fix:** `ActionServerHandle` now holds the `Clock` (it's `Arc`-refcounted), so the clock lives as long as the server and any goal handle that references it.

### Bug 2 — `rcl_action_server_fini` is never called (resource leak)

`ActionServerHandle` had no `Drop`, so `rcl_action_server_fini` was never invoked. The action server and its underlying services, publishers, and expired-goals timer leaked on every drop. The action client (`ActionClientHandle::drop`) already finalizes itself the standard way (as do `ServiceHandle`/`ClientHandle`). The server was simply missing it.

**Fix:** add a `Drop` that finalizes the server, mirroring the action client. It runs before the struct's fields are dropped, so the clock kept alive by Bug 1's fix is still valid when `fini` touches the expired-goals timer, which is why the two fixes belong together.

### Testing

- **Regression test added**: asserts that creating an action server takes a strong reference to the clock it borrows by raw pointer. Fails deterministically without Bug 1's fix, passes with it. I chose this invariant assertion over a crash-based test because a UAF doesn't crash reliably, so a crash test would be flaky, whereas this fails the instant the clock reference is dropped.
- The full single-process lib suite — which `SIGSEGV`'d ~25–33% of runs before ran clean across 32 runs under gdb after the fix.

### DIFF
--- a/rclrs/src/action/action_server.rs
+++ b/rclrs/src/action/action_server.rs
@@ -319,7 +319,7 @@ impl<A: Action> ActionServerState<A> {
         let handle = Arc::new(ActionServerHandle {
             rcl_action_server: Mutex::new(rcl_action_server),
             node_handle: Arc::clone(&node.handle()),
-            _clock: clock,
+            clock: clock,
             goals: Default::default(),
         });
 
@@ -687,8 +687,8 @@ pub(crate) struct ActionServerHandle<A: Action> {
     /// `Arc<ActionServerHandle>` and, on drop, calls `rcl_action_notify_goal_done`
     /// (which touches this clock). Without holding it here, a goal dropped after
     /// the node has been torn down would dereference a freed clock.
-    #[allow(dead_code)]
-    _clock: Clock,
+    #[allow(unused)]
+    clock: Clock,
     /// Ensure the `impl_*` of the action server goals remain valid until they
     /// have expired or until the rcl_action_server_t gets fini-ed.
     goals: Mutex<HashMap<GoalUuid, Arc<ActionServerGoalHandle<A>>>>,
--- a/rclrs/src/action/action_server.rs
+++ b/rclrs/src/action/action_server.rs
@@ -1,9 +1,9 @@
 use super::empty_goal_status_array;
 use crate::{
     action::GoalUuid, error::ToResult, rcl_bindings::*, ActionGoalReceiver, CancelResponseCode,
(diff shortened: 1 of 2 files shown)


rclrs/src/action/action_server.rs
```diff
@@ -677,6 +681,14 @@ pub(crate) struct ActionServerHandle<A: Action> {
...
     #[allow(unused)]
     node_handle: Arc<NodeHandle>,
+    /// The clock the `rcl_action_server` borrows by raw pointer for its
+    /// expired-goals timer. We must keep it alive for the lifetime of the server,
+    /// including for any goal handle that outlives the node: a goal holds an
+    /// `Arc<ActionServerHandle>` and, on drop, calls `rcl_action_notify_goal_done`
+    /// (which touches this clock). Without holding it here, a goal dropped after
+    /// the node has been torn down would dereference a freed clock.
+    #[allow(dead_code)]
+    _clock: Clock,
```
mxgrey: In other places in rclrs where we add something to a struct to keep its reference count alive, we use `#[allow(unused)]`. That would allow us to name this field as simply `clock` without the leading underscore.

mxgrey (approved):
Thanks for this fix, looks good! I just left one minor remark on a field attribute.

rclrs/src/action/action_server.rs
azerupi: Absolutely, done.

esteve (approved):
@azerupi LGTM, thanks!

PR #656, opened 2026-06-27
fix: make the test log handler coexist with rosout

azerupi:
The test-only custom logging output handler (`set_logging_output_handler`, used by the rclrs logging tests) installed itself via `rcutils_logging_set_output_handler`, *replacing* the default `rcl_logging_multiple_output_handler`. While installed this had two process-global side effects:

1. `/rosout` (and console/file) publishing was disabled, because the default handler that fans out to them was no longer called. A test running concurrently that relied on `/rosout` (e.g. the rosout publishing test) could silently lose its messages.
2. Every log record emitted anywhere in the process was captured into the handler's buffer, so concurrent tests' logs polluted assertions such as `last_message()`.

Under cargo's parallel test execution these made the logging tests intermittently fail.

Fix: the custom handler now
- chains to `rcl_logging_multiple_output_handler`, so console/file/rosout keep working while it is installed, and
- captures only records emitted on the thread that installed it (the test thread), so concurrently-running tests no longer pollute the captured output.

To keep the forwarded record printf-safe, `impl_log` now always calls `rcutils_log` with the `"%s"` + message form (instead of passing the message as the format string) and stashes the finished message in a thread-local that the handler reads back. This avoids parsing the C `va_list` in Rust while letting the record be forwarded to the default handler unchanged.

All changes are within `#[cfg(test)]` code, production logging is unchanged.

### DIFF
--- a/rclrs/src/logging/logging_configuration.rs
+++ b/rclrs/src/logging/logging_configuration.rs
@@ -69,6 +69,36 @@ pub(crate) mod log_handler {
     };
 
     use crate::{rcl_bindings::*, LogSeverity, ENTITY_LIFECYCLE_MUTEX};
+    use std::{cell::RefCell, ffi::CString, thread::ThreadId};
+
+    thread_local! {
+        /// The fully-formatted message for the log record currently being emitted
+        /// on this thread. `impl_log` stashes it here right before calling
+        /// `rcutils_log` so the capture handler can read the finished string
+        /// without parsing the C `va_list` (the messy problem noted above). Only
+        /// read on the capture thread; cleared by `impl_log` after each record.
+        static CURRENT_MESSAGE: RefCell<Option<CString>> = const { RefCell::new(None) };
+    }
+
+    /// The thread that installed the capture handler (the logging test's thread).
+    /// The custom handler captures only records emitted on this thread, so logs
+    /// from concurrently-running tests don't pollute the captured output.
+    static CAPTURE_THREAD: OnceLock<ThreadId> = OnceLock::new();
+
+    /// Stash the formatted message for the current log record (called by
+    /// `impl_log` before `rcutils_log`).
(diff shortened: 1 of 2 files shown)


rclrs/src/logging/logging_configuration.rs
```diff
@@ -225,6 +264,26 @@ pub(crate) mod log_handler {
         message: *const std::os::raw::c_char,
         logging_output: *mut va_list,
     ) {
+        // Chain to the default rcl handler so console/file/`/rosout` output keeps
+        // working while our capture handler is installed. Our custom handler used
+        // to *replace* this, which silently disabled `/rosout` publishing
+        // process-wide for the duration — breaking concurrently-running tests that
+        // rely on `/rosout`. The format string and `va_list` are passed through
```
maspe36, line 271: This change seems fine in of itself, 

> breaking concurrently-running tests that rely on `/rosout`

but to me this is the real issue. _Unit_ tests that depend on system level properties are actually integration tests and I argue shouldn't be run concurrently. 

Do you have a list of these tests that would break concurrently?

azerupi: I believe it is only `test_logging_macros` because it replaces the logging out handler by calling `set_logging_output_handler` and this then intermittently breaks the other tests that subscribe to `/rosout`

rclrs/src/logging/logging_configuration.rs
maspe36, line 60: Something good to know, in Rust 1.99 `c-variadic` is getting stabilized! 

https://github.com/rust-lang/rust/issues/162306

rclrs/src/logging/logging_configuration.rs
maspe36: I still don't love depending on system level properties in concurrent unit tests. However, whenever we can move to Rust 1.99 I believe we will have the tools needed to reassess this.

maspe36 (approved):

PR #640, opened 2026-05-17
add lyrical to CI jobs and rcl_bindings.rs

ottojo:
Lyrical is about to be released, the `rostooling/setup-ros-docker:ubuntu-resolute-ros-lyrical-ros-base-latest` docker image already exists. I successfully ran the `../generate_bindings.py rcl_wrapper.h lyrical .` command and ran the subscriber example on lyrical.

PR #658, opened 2026-07-20
feat: add support for lyrical.

jpace121:
## Description

Replacement for https://github.com/ros2-rust/ros2_rust/pull/640. 

This MR rebases all the commits from the linked MR on top of main and adds the lyrical bindings file which was missing from the above MR. 

## Verification

Builds locally. The min pub sub example works in my local workspace.


jpace121:
> Caused by:
>    process didn't exit successfully: `/__w/ros2_rust/ros2_rust/ros_ws/build/rclrs/debug/deps/rclrs-9634ec9615184451 --color=never --quiet` (signal: 11, SIGSEGV: invalid memory reference)

Well that's interesting. Not really sure what's happening here...

jpace121:
It appears the segfault has something to do with Actions. Not sure what though.

jpace121:
I'll be honest, I have a bad feeling the seg fault is something that also happens in rolling given how similar the bindings are.

esteve:
@jpace121 thanks for submitting this PR, I'll close the other one as the original submitter hasn't addressed my feedback since submitting #640 . This segfault might be related to #659 , but needs further debugging, though it's strange that Rolling doesn't seem to be affected in CI.

jpace121:
I made a first pass at seeing if the linked MR fixed the issue locally, but it appears to test that, I'd need to rebuild basically every msg package, which is beyond what I'm interested in biting off tonight.

I'll look at that later, I think its going to require some thinking on how to get that to work.

jpace121:
Commenting here to note I'm still interested in this. 

My understanding from reading the issue tracker traffic is that esteve's hypothesis about the segfault in CI being related related to another that already has a in PR fix is very likely. 

The reason things are working in rolling is the Docker images for rolling weren't being updated, which is fixed in https://github.com/ros2-rust/ros2_rust/pull/668.

azerupi:
Correct yes, we need to merge https://github.com/ros2-rust/rosidl_runtime_rs/pull/22 ~and https://github.com/ros2-rust/rosidl_rust/pull/36 but the second PR actually needs to go on the ROS build-farm and then we need to wait on a sync from the ROS project for each distro which can take some time.~

In the meantime I think @maspe36 is investigating an avenue where we would not depend on the ROS build-farm and sync schedule.

maspe36 (dismissed):
LGTM

esteve (approved):
@jpace121 thanks!

azerupi (approved):
Thanks @jpace121 rolling and lyrical passed in CI already! 
I just added one small commit for consistency but everything should be good to merge.

PR #660, opened 2026-07-27
Fix WaitSetRunner cleanup before propagating callback panics

### LINKED ISSUE #613: Got `WaitSetRunner unexpectedly dropped`
I'm low on time but I can create a minimal example later. I got

```
[FATAL] [1773330122.816267516] [rclrs.basic_executor]: WaitSetRunner unexpectedly dropped. This should never happen. Please report this to the rclrs maintainers with a minimal reproducible example.
```

The thing that caused this was `panic!()`ing in inside a subscription created using a `Worker`. The node continues to run even after it hits this error, though the topic cannot receive anymore. This suppresses the panic, which is bad. When using a non-`Worker` subscription, this causes the application to panic as expected. Let me know how I can provide more help.

coolcmyk:
Fixes #613.

  A panic inside a callback was caught by its `WaitSetRunner`, but the basic
  executor immediately resumed unwinding when it received the panic. This
  allowed executor and ROS entities to be destroyed while other runner threads
  were still active, causing RMW cleanup errors and potentially a segmentation
  fault.

  This change:

  - Returns callback panics from `WaitSetRunner` as a distinct outcome.
  - Saves the first panic payload and the runner that encountered it.
  - Signals all remaining runners to stop and wakes runners blocked in the wait set.
  - Waits for the runners to finish.
  - Restores executor-owned state before resuming the original panic.

  The panic still propagates to the caller, but worker shutdown and ROS resource
  cleanup now happen first.

  ### Validation

  - Ran `cargo check -p rclrs --lib`.
  - Reproduced the issue using example 613 with an intentional callback panic.
  - Confirmed that the panic is propagated without the previous RMW cleanup
    errors or segmentation fault.
  - Confirmed the panic can be caught by the caller with `catch_unwind`.
  
  ```
  use rclrs::*;
use ros_env::std_msgs;
use std::time::Duration;

fn main() -> Result<(), RclrsError> {
    let mut executor = Context::default_from_env()?.create_basic_executor();
    let node = executor.create_node("worker_demo")?;

    let publisher = node.create_publisher("output_topic")?;

    // Keep a separate runner blocked in rcl_wait when the subscription callback
    // panics. The executor must stop and reclaim this runner before propagating
    // the panic.
    let idle_worker = node.create_worker(());
    let _idle_timer = idle_worker.create_timer_inert(Duration::from_secs(60))?;

    let worker = node.create_worker(String::new());
    let _subscription = worker.create_subscription(
        "input_topic",
        move |data: &mut String, msg: std_msgs::msg::String| {
            *data = msg.data;
            panic!();
        },
    )?;

    // Use this timer-based implementation when timers are available instead
    // of using std::thread::spawn.
    let _timer =
        worker.create_timer_repeating(Duration::from_secs(1), move |data: &mut String| {
            let msg = std_msgs::msg::String { data: data.clone() };

            publisher.publish(msg).ok();
        })?;

    println!(
        "Beginning repeater... \n >> \
        Publish a std_msg::msg::String to \"input_topic\" and we will periodically republish it to \"output_topic\".\n\n\
        To see this in action run the following commands in two different terminals:\n \
        $ ros2 topic echo output_topic\n \
        $ ros2 topic pub input_topic std_msgs/msg/String \"{{data: Hello}}\""
    );
    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
        executor.spin(SpinOptions::default());
    }));

    assert!(result.is_err(), "callback panic was not propagated");

    Ok(())
}
  ```
image
image

### DIFF
--- a/rclrs/src/action/action_goal_receiver.rs
+++ b/rclrs/src/action/action_goal_receiver.rs
@@ -6,6 +6,7 @@ use rosidl_runtime_rs::Action;
 use std::{
     future::Future,
     ops::{Deref, DerefMut},
+    panic::UnwindSafe,
     sync::Arc,
 };
 use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver};
@@ -47,7 +48,7 @@ impl<A: Action> ActionGoalReceiver<A> {
     #[must_use]
     pub fn into_action_server<Task>(
         self,
-        callback: impl FnMut(RequestedGoal<A>) -> Task + Send + Sync + 'static,
+        callback: impl FnMut(RequestedGoal<A>) -> Task + Send + Sync + UnwindSafe + 'static,
     ) -> ActionServer<A>
     where
         Task: Future<Output = TerminatedGoal> + Send + Sync + 'static,
--- a/rclrs/src/action/action_server.rs
+++ b/rclrs/src/action/action_server.rs
@@ -14,6 +14,7 @@ use std::{
     collections::HashMap,
     ffi::CString,
     future::Future,
+    panic::UnwindSafe,
     sync::{Arc, Mutex, MutexGuard, Weak},
     time::Duration,
 };
@@ -204,7 +205,7 @@ impl<A: Action> ActionServerState<A> {
     /// Change the callback for this action server.
     pub fn set_callback<Task>(
         &self,
-        mut callback: impl FnMut(RequestedGoal<A>) -> Task + Send + Sync + 'static,
+        mut callback: impl FnMut(RequestedGoal<A>) -> Task + Send + Sync + UnwindSafe + 'static,
     ) where
         Task: Future<Output = TerminatedGoal> + Send + Sync + 'static,
     {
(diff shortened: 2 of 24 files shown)


mxgrey (commented):
Let's add [`UnwindSafe`](https://doc.rust-lang.org/std/panic/trait.UnwindSafe.html) as a trait bound for all callbacks to help ensure that catching the unwind will be safe.

coolcmyk:
Hi mxgrey, i've added `UnwindSafe` across the callbacks in this [commit](https://github.com/ros2-rust/ros2_rust/pull/660/changes/9c88d64e839bdb86c8faa1bce3ca0e631dbaf996). Tests has passed with

```bash
cargo test -p rclrs --lib
```

```bash
test result: ok. 119 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.45s
```

> Let's add [`UnwindSafe`](https://doc.rust-lang.org/std/panic/trait.UnwindSafe.html) as a trait bound for all callbacks to help ensure that catching the unwind will be safe.

PR #663, opened 2026-08-06
Several improvements around parameters

azerupi:
This PR has 3 changes that improve parameter handling.

#### Problem
Right now on main, users of rclrs can implement the **public** `ParameterVariant` trait for arbitrary Rust types. This is for example very useful to implement string-backed enums.

However the code in rclrs that writes the parameter only checks that the kind is correct before storing the value. However the functions that read the value (`MandatoryParameter::get`, `OptionalParameter::get`, `ReadOnlyParameter::get`, both `subscribe()` closures, and both `on_change` wrappers) assume it can infallibly convert to the Rust type.

```rust
pub fn get(&self) -> T {
    self.value.read().unwrap().clone().try_into().ok().unwrap()
}
```

So let's imagine we have this user code

```rust
    #[derive(Clone, Debug, PartialEq)]
    enum Switch {
        On,
        Off,
    }

    impl From<Switch> for ParameterValue {
        fn from(value: Switch) -> Self {
            ParameterValue::String(
                match value {
                    Switch::On => "on",
                    Switch::Off => "off",
                }
                .into(),
            )
        }
    }

    impl TryFrom<ParameterValue> for Switch {
        type Error = ParameterValueError;

        fn try_from(value: ParameterValue) -> Result<Self, Self::Error> {
            match value {
                ParameterValue::String(s) => match s.as_ref() {
                    "on" => Ok(Switch::On),
                    "off" => Ok(Switch::Off),
                    other => Err(ParameterValueError::Invalid(format!(
                        "unknown Switch '{other}', expected one of: on, off"
                    ))),
                },
                _ => Err(ParameterValueError::TypeMismatch),
            }
        }
    }

    impl ParameterVariant for Switch {
        type Range = ();

        fn kind() -> ParameterKind {
            ParameterKind::String
        }
    }
```

It will only work if the string value is `"on"` or `"off"`, anything else will result in a panic since we are only validating the kind. Any string can be stored in the parameter.

```
$ ros2 param set /my_node switch banana
Set parameter successful
```

This is pretty bad because any remote client on the ROS graph could crash a node that implements this.

#### Fix 7c9cac1cd1b021efe171a0a411d01ee1276a9e9c

To fix this, we are now checking the conversion before storing the value. Any value that doesn't pass the conversion is does not get stored. In the same way other built-in checks run (e.g. ranges).

```
$ ros2 param set /my_node switch banana
Setting parameter failed: Parameter value is not valid for this parameter's type: unknown Switch 'banana', expected one of: on, off
```

and `param.get()` still returns the previous value because the new value was never stored.

#### Gap
ROS 2 parameter descriptors carry constraints in `IntegerRange`, `FloatingPointRange` or `additional_constraints` (free text).

Right now, for custom fields, for example the previously defined `Switch` enum the constraints have to be passed at each  declaration of a parameter using this type. There is also nothing that keeps those multiple constraint declarations in sync with the type, adding a new enum variant will silently make the declared constraints drift unless all of them are updated. And if you forget to declare the constraints, `ros2 param describe` gives you no information.

#### Fix 3f27dccdf88eb884e50973baeabae81d9df39859

We add a new method on `ParameterVariant` that allows the the constraints to be declared where the type is defined as a ROS parameter. It has a default implementation that returns `None`.

This allows a type to specify its constraints. If the user declaring the parameter adds explicit constraints they will be used otherwise we default to the type's constraints.

#### Gap

A parameter could only be declared with the Rust type that most directly represents one of the nine ROS 2 types: bool, i64, f64, Arc<str>, Arc<[u8]>, Arc<[bool]>, Arc<[i64]>, Arc<[f64]>, Arc<[Arc<str>]>. This is convenient for the parameter machinery (rclrs) but inconvenient for the application (user).

#### Fix 4f098d45114b152eaab24b467b5a47cc92340299

We implement `ParameterVariant` for many common Rust types to allow them to be used as parameters.

| Category | Types | ROS 2 kind |
|---|---|---|
|Strings | `String`, `PathBuf` | String |
| Narrow integers | `i8`, `i16`, `i32`, `u8`, `u16`, `u32` | Integer |
| Narrow float | `f32` | Double |
| Durations | `DurationSecs`, `DurationMillis` (new types around `Duration`) | Double / Integer | 
| Arrays | `Vec<T>` of every scalar above, plus `Vec<i64>`, `Vec<f64>`, `Vec<bool>`, `Vec<u8>`, `Vec<String>` | matching array kind |

**Note:**
- `Vec<u8>` is a byte array, not an integer array. ROS 2 has a dedicated byte array type. It's the one Rust array type that doesn't follow its scalar (u8 alone is an Integer).
- `u64`, `usize`, `i128` and `u128` are deliberately absent. ROS 2's integer type is i64 and conversions from those types could silently produce values that did not match the intent.
- I decided to create newtypes around `Duration` to be able to differentiate between seconds (Float) and milliseconds (Integer). 

This was only possible after fix n°1 because now we have many Rust types that can't represent all values that could be stored in the ROS type. But since we now check the conversion before storing, this is totally fine!

The narrow integers define an `IntegerRange` that allow to describe the parameter constraints. 

**Before:**
```rust
let wheels: MandatoryParameter<Arc<[Arc<str>]>> = node
    .declare_parameter("wheels")
    .default_string_array(["left", "right"])
    .mandatory()?;
let timeout_s: MandatoryParameter<f64> = /* ... and convert by hand at every use */;
```

**After:**
```rust
let wheels:  MandatoryParameter<Vec<String>>  = node.declare_parameter("wheels")
    .default(vec!["left".to_string(), "right".to_string()]).mandatory()?;
let device:  MandatoryParameter<PathBuf>      = node.declare_parameter("device")
    .default(PathBuf::from("/dev/ttyUSB0")).mandatory()?;
let timeout: MandatoryParameter<DurationSecs> = node.declare_parameter("timeout")
    .default(DurationSecs(Duration::from_millis(2500))).mandatory()?;
let port:    MandatoryParameter<u16>          = node.declare_parameter("port")
    .default(8080).mandatory()?;
```

Beyond the ergonomics provided here, this is the groundwork for future PRs that provide a derive proc-macro based API similar to clap.

This PR was assisted by Claude code (opus) to start and reviewed and iterated on manually.

### DIFF
--- a/rclrs/src/parameter.rs
+++ b/rclrs/src/parameter.rs
@@ -1,11 +1,13 @@
 mod override_map;
 mod range;
 mod service;
+mod std_types;
 mod value;
 
 pub(crate) use override_map::*;
 pub use range::*;
 use service::*;
+pub use std_types::*;
 pub use value::*;
 
 use ros_env::rcl_interfaces::msg::rmw::{ParameterType, ParameterValue as RmwParameterValue};
--- a/rclrs/src/parameter.rs
+++ b/rclrs/src/parameter.rs
@@ -50,9 +50,17 @@ struct ParameterOptionsStorage {
 
 impl<T: ParameterVariant> From<ParameterOptions<T>> for ParameterOptionsStorage {
     fn from(opts: ParameterOptions<T>) -> Self {
+        // A declaration that says nothing about constraints inherits whatever the parameter's
+        // type has to say about itself, so that introspection is useful without every
+        // declaration site restating the type's own rules.
+        let constraints = if opts.constraints.is_empty() {
+            T::type_constraints().unwrap_or(opts.constraints)
+        } else {
+            opts.constraints
+        };
         Self {
             description: opts.description,
-            constraints: opts.constraints,
+            constraints,
             ranges: opts.ranges.into(),
         }
     }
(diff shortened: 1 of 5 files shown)


rclrs/src/parameter.rs
```diff
@@ -622,6 +668,15 @@ impl ParameterMap {
...
                     {
+                        // The kind matching is not sufficient. The parameter was declared with a
+                        // concrete Rust type, and a value of the right kind can still be
+                        // unrepresentable in it. Reject those here rather than letting a later
+                        // read of the parameter fail.
+                        if let Err(reason) = (storage.type_check)(&value) {
+                            return Err(format!(
+                                "Parameter value is not valid for this parameter's type: {reason}"
+                            ));
+                        }
```
balthasarschuess, line 679: I think this is the crux of this PR and a really nice improvement!
This would allow us to arbitrary user values and even support blanked support `serde` serializable/deserializable values and keep nice error messages.

I champion this approach and I'd say go for the more idiomatic https://github.com/ros2-rust/ros2_rust/pull/612 over my hacky version https://github.com/ros2-rust/ros2_rust/pull/516
I'll already close my PR :)

balthasarschuess (commented):
Looks good to me and I think this is the way forward for ergonomic user code integration.

rclrs/src/parameter.rs
azerupi: Thank you for looking into this!
Indeed, closing that gap allowed me to implement the parameter trait for most basic Rust types which is already a very good ergonomic improvement on its own.

I am working on a successor to #612 with similar ideas but even better design in my opinion. I still need to do a lot of cleanup, but stay tuned for follow up PRs :slightly_smiling_face:

azerupi:
Superseded by the stack of PRs #687, #688, #690 and #691

Issue #664, opened 2026-08-07
`proxy_elems` checks the C buffer's length against the size of the Rust proxy type — small message arrays panic in release


espressolee:
`rclrs/src/dynamic_message/field_access/dynamic_sequence.rs`, `TypeErasedSequence::proxy_elems` and `proxy_elems_mut`.

The guard on these two functions compares two quantities that are not related to each other, and the mismatch makes rclrs panic while reading valid messages.

```rust
// dynamic_sequence.rs:196-211
let element_size = T::size_in_memory(metadata);
let sequence_data =
    std::slice::from_raw_parts(self.data as *const u8, self.size * element_size);
check::<T>(sequence_data);
```

`T` here is the **proxy** type, not the element type. So `check` (field_access.rs:18-22)

```rust
fn check<T>(bytes: &[u8]) {
    assert!(bytes.len() >= std::mem::size_of::<T>());
    let align = std::mem::align_of::<T>();
    assert_eq!(bytes.as_ptr().align_offset(align), 0);
}
```

asserts `self.size * element_size >= size_of::<DynamicMessageView>()`.

`DynamicMessageView<'msg>` is `{ &'msg MessageStructure, &'msg [u8] }` — a thin reference plus a fat one, **24 bytes** on 64-bit. That number is a property of a Rust struct made of two references. `sequence_data` is the C-owned message buffer. The assertion relates the size of the former to the length of the latter, and there is no reason for those to be comparable.

The module's own header comment already says as much (`dynamic_sequence.rs:17-18`): *"The memory layout of the proxy type doesn't match that of T"*.

## The consequence, on valid input

Any `message[]` field whose **total payload** is under 24 bytes trips the assertion. It is `assert!`, not `debug_assert!`, so it fires in release.

`geometry_msgs/msg/Polygon` carries `Point32[] points`, and `Point32` is three `float32` = 12 bytes. **A Polygon with one point panics.** Two points is 24 and passes.

Others in the same shape: `builtin_interfaces/Time[]` (8 bytes/element) with one or two elements, and any small-message array under 24 bytes total.

The threshold is incidental rather than chosen — adding one field to `DynamicMessageView` would move it from 24 to 32 and silently widen the set of messages that panic.

## Reproduced, std-only

No ROS 2, no rclrs, no FFI. `check` and the call site transcribed verbatim; `MessageStructure` reduced to the one field that is read, layout preserved.

```rust
fn check<T>(bytes: &[u8]) {
[...]

PR #665, opened 2026-08-08
docs: say what actually guards the dynamic_message publish/take casts

espressolee:
Comments only. No code changes — every changed line starts with `//`, and stripping comments
from both files before and after gives byte-identical text.

### What it fixes

`DynamicPublisher::publish` carries this:

```rust
// SAFETY: The message type is guaranteed to match the publisher type by the type system.
```

It is a verbatim copy of the comment on `Publisher::publish`, misspelling of "explictly"
included. On `Publisher<T>` the sentence is true — the generic parameter ties the message to
`<T as Message>::RmwMsg`. On `DynamicPublisher`, whose whole point is that the type is a runtime
value, the compiler cannot see anything of the sort. What actually guards the cast is the
`message_type` comparison two lines above, and that comparison is by **name**
(`MessageTypeName { package_name, type_name }`).

### The obligation neither comment mentions

Two type support libraries are involved:

| | library | role |
|---|---|---|
| `DynamicMessageMetadata` | `rosidl_typesupport_introspection_c` | lays out `message.storage` |
| the publisher / subscription | `rosidl_typesupport_c` | what `rcl_publish` / `rcl_take` serialise with |

So the name comparison rests on those two describing the same layout. Both constructors already
say the two libraries exist, in prose a few lines up — *"This loads the introspection type
support library … However, we also need the regular type support library"* — so this is known,
just missing from the place where the `unsafe` block is justified.

`DynamicSubscription`'s `take()` is the mirror: `metadata.create()` lays out the storage by
introspection, `rcl_take` writes it through the C type support, and the comment discusses
pointer validity only.

### Both are sound as written

Each constructor resolves the metadata and the C type support from the same package name in the
same function, through one `ament.find_package`, so they come from one install prefix and agree
by construction. This PR records that, rather than changing it. The point is that a reader
auditing these blocks is currently told the compiler is checking something it never sees.

Happy to reword or drop either hunk.

---

### Generative AI disclosure

This PR was opened on 2026-08-08, before this repository adopted the OSRF AI policy in #672 on
2026-09-01. Adding the disclosure now so the record is complete:

```text
Assisted-by: Codex:gpt-5.6-sol
```

`AGENTS.md` also asks for the trailer in the commit message body. I have left the commit as it
is rather than rewriting it, since amending would invalidate the existing CI results and any
references to the current SHA. Happy to amend and force-push if you would prefer the trailer on
the commit itself.

### DIFF
--- a/rclrs/src/dynamic_message/dynamic_publisher.rs
+++ b/rclrs/src/dynamic_message/dynamic_publisher.rs
@@ -125,9 +125,19 @@ impl DynamicPublisherState {
         }
         let rcl_publisher = &mut *self.handle.rcl_publisher.lock().unwrap();
         unsafe {
-            // SAFETY: The message type is guaranteed to match the publisher type by the type system.
+            // SAFETY: Unlike `Publisher<T>`, nothing here is enforced by the type system --
+            // a `DynamicMessage`'s type is a runtime value. What guarantees the match is the
+            // `message_type` comparison at the top of this function.
+            //
+            // That comparison is by name (`MessageTypeName`), so it stands on the two type
+            // support libraries for a given package describing the same layout: the
+            // introspection one laid out `message.storage` (via `DynamicMessageMetadata`),
+            // while `rcl_publish` serialises those bytes through the `rosidl_typesupport_c`
+            // one. `DynamicPublisher::create` resolves both from the same package name, so
+            // they come from one install prefix.
+            //
             // The message does not need to be valid beyond the duration of this function call.
-            // The third argument is explictly allowed to be NULL.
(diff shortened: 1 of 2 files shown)


espressolee:
Gentle review ping: this remains a comments-only safety-contract clarification at `15b5548b24b0db4161d46a901aab36c6324f01f0`.

- Rust Stable: success
- Rust Minimal: success
- Rust Windows: success
- mergeable: true
- no runtime, ABI, or generated-code change

The issue it documents (#664) is still open; this PR only states the existing ownership/lifetime guard around the casts and does not claim to fix #664. Happy to reword or drop either comment hunk.

PR #668, opened 2026-08-14
ci: update rolling docker image

azerupi:
In #658 it was noted that there is a segfault occurring in Lyrical. Rolling is actually similarly broken. It wasn't caught because the CI is using a very old version of some of the rolling packages. Because the docker image already contains some of the rolling packages and because we don't do an apt dist-upgrade those packages are actually never updated to the latest. So we are testing against packages from 2025-07-30 on rolling. 

This forces an update of the packages in the CI images and prevents testing on stale packages. At the same time we are updating to ubuntu resolute instead of noble for rolling as packages are not being built for noble anymore since April.

**Expected:** This PR should start making rolling failing

### DIFF
--- a/.github/workflows/generate-bindings.yml
+++ b/.github/workflows/generate-bindings.yml
@@ -34,7 +34,7 @@ jobs:
             ros_distribution: kilted
             ros_version: 2
           # Rolling Ridley  (June 2020 - Present)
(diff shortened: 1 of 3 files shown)


maspe36 (approved):
LGTM Thanks!

Issue #666, opened 2026-08-14
BasicExecutorRuntime::spin causes high CPU usage with SpinOptions::default() (busy-waiting)


Kimhlo:
Description
When using `SpinOptions::default() `(i.e. `timeout: None`), the executor consumes abnormally high CPU (~28%) even when there is no work to do.
Steps to Reproduce

```
executor
    .spin(SpinOptions::default())
    .first_error()
    .inspect_err(|e| logger_error!("Executor spin failed: {:#}", e))?;
```
Expected Behavior
With `timeout: None`, the executor should block (sleep) when there is no task available, resulting in near-zero CPU usage.
Actual Behavior
The executor enters a busy-wait loop. perf profiling shows:
```
# Children      Self  Command          Shared Object                    Symbol                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
# ........  ........  ...............  ...............................  ...........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
#
    99.81%     0.00%  xxxxxx_ros_node  [xxxxxxxx]                        [k] xxxxxxxxxxxxx
            |
            ---0xaaaaffffffff
               <rclrs::executor::Executor>::spin
               |          
               |--28.39%--<rclrs::executor::basic_executor::BasicExecutorRuntime as rclrs::executor::ExecutorRuntime>::spin
               |          
               |--22.45%--<std::sync::mpmc::list::Channel<alloc::sync::Arc<rclrs::executor::basic_executor::Task>>>::send
               |          |          
               |           --0.58%--__libc_calloc
               |          
               |--13.82%--<std::sync::mpmc::list::Channel<alloc::sync::Arc<rclrs::executor::basic_executor::Task>>>::start_recv
               |          
               |--12.43%--<std::sync::mpmc::list::Channel<alloc::sync::Arc<rclrs::executor::basic_executor::Task>>>::read
               |          |          
               |           --0.72%--cfree
               |          
               |--8.90%--futures_task::waker::wake_by_ref_arc_raw::<rclrs::executor::basic_executor::Task>
               |          
               |--3.13%--<xxxxxx_ros_node::xxxxxxxxx::xxxxxx::xxxxxxx>::xxxxxxx_service::{closure#0}::{closure#1}
               |          
[...]

PR #669, opened 2026-08-17
Coalesce duplicate BasicExecutor task wake-ups

Dylan-Gallagher:
## Summary

- track whether a `BasicExecutorRuntime` task already has a ready-queue entry
- coalesce repeated wake-ups into one pending poll
- preserve wake-ups delivered while the task is being polled
- add regressions for burst wake-ups and wake-during-poll ordering

## Context

The profile in #666 points at the task send/receive/waker path. The blocking
receiver does sleep when the ready queue is empty in a minimal idle executor,
but `Task::wake_by_ref` currently enqueues the same task on every notification.
A burst of notifications can therefore keep redundant task entries in the
queue and cause repeated polls after the future has already observed the latest
state.

Rust's waker contract allows multiple wake-ups before the next poll to be
coalesced. This change uses an atomic queued flag to allow only one outstanding
ready entry per task. The flag is cleared before `Future::poll`, which ensures
that a wake delivered during the poll still schedules another poll.

I could not reproduce the issue's exact 28% CPU result without the reporter's
full async service workload, so this PR makes the narrower claim: it removes a
deterministically reproduced queue-amplification mechanism in the profiled hot
path. A two-second minimal idle spin was already near zero CPU before the
change.

## Testing

- `cargo test -F default,serde executor::basic_executor::tests -- --test-threads=1`
  - 2 passed
- `cargo test -F default,serde -- --test-threads=1`
  - 121 unit tests and 51 doc tests passed
- `colcon build --packages-select rclrs`
  - passed in an isolated ROS 2 Jazzy container workspace
- `cargo fmt --all -- --check`
- `rustfmt +nightly --edition 2021 --check rclrs/src/executor/basic_executor.rs`
- `git diff --check`

The burst-wake regression demonstrates that 1,000 wake notifications result in
one outstanding ready entry. The second regression verifies that a synchronous
self-wake during the first poll still produces the required second poll.

Relates to #666.

### DIFF
--- a/rclrs/src/executor/basic_executor.rs
+++ b/rclrs/src/executor/basic_executor.rs
@@ -136,20 +136,7 @@ impl ExecutorRuntime for BasicExecutorRuntime {
         }));
 
         while let Ok(task) = self.next_task(&workers_finished) {
-            // SAFETY: If the mutex is poisoned then we have unrecoverable situation.
-            let mut future_slot = task.future.lock().unwrap();
-            if let Some(mut future) = future_slot.take() {
-                let waker = waker_ref(&task);
-                let task_context = &mut TaskContext::from_waker(&waker);
-                // Poll the future inside the task so it can do some work and
-                // tell us its state.
-                if future.as_mut().poll(task_context).is_pending() {
-                    // The task is still pending, so return the future to its
-                    // task so it can be processed again when it's ready to
-                    // continue.
-                    *future_slot = Some(future);
-                }
-            }
+            task.poll();
         }
 
         let (runners, new_worker_receiver, errors) = worker_result_receiver.recv().expect(
@@ -349,6 +336,7 @@ impl TaskSender {
         let task = Arc::new(Task {
             future: Mutex::new(Some(f)),
             task_sender: self.task_sender.clone(),
+            queued: AtomicBool::new(true),
         });
 
         if let Err(_) = self.task_sender.send(task) {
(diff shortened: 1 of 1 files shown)


azerupi:
Hey @Dylan-Gallagher, thank you for the PR! 

I looked into #666 and with the additional example provided by the reported I was able to reproduce and find out the cause. You can check out the issue for the explanation. As it stands, this PR would not solve the issue (I have tested), it's not really something we can solve easily in rclrs because it is a tokio behavior.

espressolee (commented):
Reviewed exact head `a0a85afb6159486b71450c130ef40c34fe53a707` against exact base `7ec55378e123df42f3671751161b9fbd046c4214`. Current `main` is now `4606ed41a212fb6c65ab01d1e79ddabde4e7cb05`, but `basic_executor.rs` is unchanged across that base drift and the merge tree is clean.

I agree with the existing discussion that this does not solve #666's Tokio cooperative-budget loop. I modeled the submitted `Task`/`ArcWake`/`std::mpsc` state machine independently under the project's Rust 1.85 MSRV and checked the narrower queue-coalescing claim:

- 1,000 baseline burst wakes produce 1,000 queue entries; this head produces 1
- 16 threads × 1,000 wakes still produce one outstanding entry
- a cross-thread wake delivered while `Future::poll` is blocked schedules the required second poll
- a synchronous wake immediately before `Poll::Ready` leaves one harmless no-op entry, which drains cleanly
- a stale waker remains coalesced after completion
- receiver disconnect resets `queued` after the failed send
- an 8-thread, 16,000-generation wake storm eventually observes the final generation without a lost wake

The seven hostile tests passed 100 consecutive runs (700/700 total executions), with Rust 1.85 `fmt` and `clippy -D warnings` clean for the isolated model. The exact-head CI is green on Humble, Jazzy, Kilted, Rolling, and Windows; the stable Linux log reports 121 unit tests and 51 doc tests passing on each ROS distribution.

The important high-level limit is also deterministic: a future that self-wakes once on every poll and returns `Pending` still executes 1,000/1,000 polls with the ready queue remaining non-empty. Coalescing cannot affect the one-wake-per-poll Tokio coop-budget behavior described in #666.

I found no correctness blocker in the narrow queue-entry invariant. The remaining decision is whether the extra atomic on every wake is justified by a real burst-wake workload independent of #666. A representative benchmark or queue-depth/poll-count measurement would establish that value; otherwise this is a correct but currently unquantified optimization. The upstream regressions are synchronous, so a cross-thread wake-during-poll test would also strengthen the concurrency contract.

Local limitation: the repository's documentation-only ROS shim does not provide the generated message types needed to compile the full `rclrs` test target, and the local Docker daemon was unavailable. I therefore rely on the exact-head remote ROS matrix for full-crate qualification and keep the local claim bounded to the changed task state machine.

PR #670, opened 2026-08-22
Remove sync requirement for action server task

mxgrey:
This is a teeny tiny PR that removes an unnecessary restriction on action servers.

When I initially implemented the action server API, I naively put both `Send + Sync` requirements everywhere to ensure that objects can be used safely across threads. This is more restrictive than necessary for the async task that needs to be produced by an action server. Async tasks really only need to be `Send`, not `Sync`. It's worth noting that [`BoxFuture`](https://docs.rs/futures/latest/futures/future/type.BoxFuture.html), which gets used to encapsulate many async tasks across many different Rust async execution engines, only requires `Send` and not `Sync`.

This change is backwards compatible for all existing uses of this API, because we are only removing a constraint.

### DIFF
--- a/rclrs/src/action/action_server.rs
+++ b/rclrs/src/action/action_server.rs
@@ -245,7 +245,7 @@ impl<A: Action> ActionServerState<A> {
         mut callback: impl FnMut(RequestedGoal<A>) -> Task + Send + Sync + 'static,
     ) -> Result<ActionServer<A>, RclrsError>
     where
-        Task: Future<Output = TerminatedGoal> + Send + Sync + 'static,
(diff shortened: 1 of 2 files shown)


azerupi:
There are a few more places with the same where clause 

```rust
where Task: Future<Output = TerminatedGoal> + Send + Sync + 'static
```

Should we change all of them?

- `drain_receiver_into_callback`
- `set_callback`
- `into_action_server`  in `rclrs/src/action/action_goal_receiver.rs`

mxgrey:
@azerupi thanks for catching that, yes! I've updated this in 9d321ee

azerupi (approved):

PR #676, opened 2026-08-29
ci: guard the Windows libclang rename against an existing target

azerupi:
The Windows job fails in "Install ros2_rust prerequisites" before any Rust is compiled:

    Added libclang >=21.1.0,<22
    Rename-Item: Cannot create a file when that file already exists.

The step renames libclang-13.dll to libclang.dll so that bindgen can find it. The conda libclang package now resolves to 21.1.0, and that build ships both libclang-13.dll and libclang.dll, so the rename collides with a file that is already there. GitHub runs pwsh with $ErrorActionPreference = 'Stop', which turns that non-terminating error into a failed step and aborts the job.

The guard only tested the source path. $dst was computed and never used, so the missing half of the check was already visible in the source. Test both paths and skip the rename when the target exists, which keeps the workaround in place for older package builds that ship only the versioned DLL.

Assisted-by: Claude:claude-opus-5 [Claude Code]

### DIFF
--- a/.github/workflows/rust-win.yml
+++ b/.github/workflows/rust-win.yml
@@ -39,13 +39,16 @@ jobs:
 
       - name: Install ros2_rust prerequisites 
         # prerequisites and fixes for windows build ros2_rust:
-        #     * Libclang has to be added (from the ros2_rust instructions) and the dll has to be renamed
+        #     * Libclang has to be added (from the ros2_rust instructions). Older builds of the
+        #       conda package only shipped a versioned libclang-13.dll, which bindgen does not
+        #       find, so it gets renamed. Newer builds ship libclang.dll themselves, in which
+        #       case there is nothing to rename.
         #     * colcon-ros-cargo and colcon-cargo have to be added as PyPI packages
         run: |
           pixi add libclang --manifest-path C:\pixi_ws\pixi.toml
           $src = "C:\pixi_ws\.pixi\envs\default\Library\bin\libclang-13.dll"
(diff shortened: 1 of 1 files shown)


azerupi:
Windows CI started failing in main, see this run: https://github.com/ros2-rust/ros2_rust/actions/runs/33256371360/job/99110720086?pr=663

This PR fixes it. Of course we still have the CI for rolling that is broken until we pull in the fixes for Lyrical.

knmcguire:
So libclang is also now part of the .lock file in ros2 repo so it gets installed already

https://github.com/ros2/ros2/blob/266da106e27445f50426d3973252bc552fe7a6f4/pixi.lock#L109

Would the install and rename step of libclang even be necessary at this point?

azerupi:
@knmcguire I'm not super familiar with how pixi works but that entry in the lockfile you linked to seems to be linux-64 only and not win-64?

knmcguire:
Ahh yes you are right. I didnt notice that they added Linux as an environment now for pixi before so hence the mix up.

I am not a maintainer so I can't approve the PR but the fix looks good to me!

knmcguire:
Fyi, once ros-ci-actions has windows support working again, you can switch to that. I won't be able to work on that until after ROSCon though but hope to be able to help out there soon.

But it's good to keep the clang issue in mind. This is not the first time I've seen it happening for pixi for windows...

https://github.com/ros-tooling/setup-ros/issues/893

PR #679, opened 2026-08-31
Update setup-pixi action version to v0.10.2

knmcguire:
Seems like the latest windows CI is broken, and that is probably due to that the pixi.toml file for rolling has been updated to only work with the latest pixi version. This should hopefully solve the issue


azerupi:
Hey @knmcguire !
See https://github.com/ros2-rust/ros2_rust/pull/676

knmcguire:
@azerupi ah thanks! My fix didn't do anything anyway so I'll close it

PR #680, opened 2026-08-31
[CLOSED] Accept standard Rust ranges for parameters

### LINKED ISSUE #675: Accept standard Rust ranges in `ParameterBuilder::range()`
Declaring a bounded parameter currently requires building a `ParameterRange` struct by hand:

```rust
node.declare_parameter("speed")
    .default(0.5)
    .range(ParameterRange { lower: Some(0.0), upper: Some(1.0), step: None })
    .mandatory()?;
```

Most parameters have no step, so this is a lot of ceremony for a simple bound. Changing the builder to accept `impl Into<T::Range>` and adding a few conversion impl between Rust range types and `ParameterRange` would allow native Rust range syntax:

```rust
node.declare_parameter("speed").default(0.5).range(0.0..=1.0).mandatory()?;
node.declare_parameter("count").default(5).range(0..10).mandatory()?;
node.declare_parameter("offset").default(-3).range(..=0).mandatory()?;
```

We would need to add `From` impls into `ParameterRange<T>`  for:

- `RangeInclusive<T>`, `RangeFrom<T>`, `RangeToInclusive<T>`, `RangeFull`, generic over `T: ParameterVariant + PartialOrd`
- `Range<i64>` and `RangeTo<i64>`, converting the exclusive end into the inclusive bound end - 1

`Range<f64>` and `RangeTo<f64>` should not be implemented. `ParameterRange::upper` is inclusive and no floating point value sits just below an exclusive bound, so `.range(0.0..1.0)` should remain a compile error rather than silently becoming an inclusive bound.

anishfyi:
### **Reverted to original issue text, author removed the part were the AI agent admitted it didn't even run the tests.**

Replaces #678, which GitHub closed automatically when its head fork was accidentally deleted. This branch restores the exact reviewed commit.

Closes #675.

## Summary

- make `ParameterBuilder::range` accept `impl Into<T::Range>`
- convert inclusive, lower-bounded, upper-bounded, and full standard Rust ranges into `ParameterRange`
- support exclusive upper bounds for integer ranges by converting them to the preceding inclusive value
- preserve empty exclusive integer ranges at `i64::MIN` as invalid parameter ranges
- add conversion and builder regression tests

Exclusive floating-point ranges intentionally remain unsupported, since they cannot be represented by `ParameterRange`'s inclusive upper bound.

## Testing

- `cargo fmt --check`

The `rclrs` tests require a sourced ROS installation; one is not available in this environment. The added conversion tests are covered by the repository's ROS-enabled CI.

### DIFF
--- a/rclrs/src/parameter.rs
+++ b/rclrs/src/parameter.rs
@@ -148,8 +148,12 @@ impl<'a, T: ParameterVariant> ParameterBuilder<'a, T> {
     }
 
     /// Sets the range for the parameter.
-    pub fn range(mut self, range: T::Range) -> Self {
-        self.options.ranges = range;
+    ///
+    /// Standard Rust ranges can be used for integer parameters. Floating-point
+    /// ranges must use inclusive bounds because no single value immediately
+    /// precedes an exclusive floating-point bound.
+    pub fn range(mut self, range: impl Into<T::Range>) -> Self {
+        self.options.ranges = range.into();
         self
     }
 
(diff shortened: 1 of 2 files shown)


esteve:
@anishfyi As per our AI policy, please update your contribution to reflect the use of AI tools.

> Testing
>    cargo fmt --check
>
> The rclrs tests require a sourced ROS installation; one is not available in this environment. The added conversion tests are covered by the repository's ROS-enabled CI.

This is upsetting, you didn't even run this, moreover you don't even have installed ROS, why should we review this PR?

Issue #32, opened 2020-11-05
Should we use Corrosion for CMake?


jhdcs:
Rust has a crate called [corrosion](https://github.com/AndrewGaspar/corrosion#installation) that is supposed to simplify the build process when CMake is required. Would the CMake in our current project benefit from using this crate?

If we decide to use this, we will need to raise the minimum required CMake version to at least 3.12. However, using that version of CMake means that we won't need to include corrosion within the project - we could instead use CMake's FetchContent module (which came into being in version 3.11).

gbiggs:
I think it's better to work on fixing `colcon-cargo` so that we can build `rclrs` more "natively" within a colcon workspace. This is something I hope to work on over the next few weeks.

jhdcs:
Ah, perhaps I misunderstood, but I thought `colcon-cargo` was going to be used for importing non `rclrs` projects? If we are going to use `colcon-cargo` to build `rclrs`, though, I can stop working on the `corrosion` branch on my fork. I was getting stuck on getting the generated outputs into the `install` folder anyways...

nnarain:
I think the concept is that `colcon` (and extensions such as `colcon-cargo`) are not necessarily ROS2 build tools. `colcon-cargo`  doesn't appear to do anything specific to ROS2. It integrates `cargo` into `colcon`. For example `colcon-ros` appears to be the plugin that actually builds cmake ROS packages.

So if a ROS2 Rust project is build-able via `cargo`. `colcon-cargo` can be used to integrate that in a normal `colcon build` command.

If `corrosion` is used, `colcon-cargo` is not necessary since `colcon-ros` will build it as a `cmake` package.

gbiggs:
colcon is completely unspecific to ROS. It's an entirely separate project - it isn't even an OSRF project.

Like any other ROS client library, it's perfectly fine to build `rclrs` with `colcon-cargo`. Absolutely everything in ROS 2 is built using `colcon`, even colcon itself is built using `colcon`. That would eliminate an extra translation step that `corrosion` would introduce.

jhdcs:
Ah, excellent. Thank you both for the information.

[...]

PR #684, opened 2026-09-03
ci: restore blocking Rust package checks

### LINKED ISSUE #682: CI: per-package lint/test/doc loops match `(ament_cargo)` but colcon prints `(ros.ament_cargo)`, so they run on zero packages
## Summary

The per-package loops in `rust-stable.yml`, `rust-minimal.yml` and `rust-win.yml` select packages with `awk '$3 == "(ament_cargo)"'`, but `colcon list` prints `(ros.ament_cargo)`. The loops match nothing, so the `fmt --check`, `clippy -D warnings`, `cargo test` and `rustdoc -D warnings` steps report success while processing zero packages.

The selector has read `(ament_cargo)` in every revision of `rust-stable.yml` and `rust-minimal.yml` since 2024-11-15 (`b4e975cd`), and in `rust-win.yml` since it was added on 2025-06-03 (`205b84be`). CI has installed colcon-ros-cargo from unpinned git `main` throughout. colcon-ros-cargo [#32](https://github.com/colcon/colcon-ros-cargo/pull/32) changed the type string on 2024-12-06, so with those two facts together the steps have selected zero packages since then.

The problem is the false green, not the diagnostics behind it. PRs merged with these workflows received passing lint, test and doc checks that had not run.

## Why the string changed

`colcon_ros/package_identification/ros.py` sets `desc.type = f'ros.{build_type}'`. colcon-ros-cargo #32 removed its own identification extension and re-keyed its task entry points from `ament_cargo` to `ros.ament_cargo`. CI installs colcon-ros-cargo from git `main` (`rust-stable.yml:83`), so it runs the post-#32 code.

## Evidence

Measured in a container matching the humble leg (`ros:humble`, colcon-ros-cargo at `dc26ba5f`), with the repo and every `ros2_rust_humble.repos` entry checked out:

```
$ colcon list
rclrs   src/ros2_rust/rclrs   (ros.ament_cargo)
```

`(ament_cargo)` appears on 0 lines. Same result with and without colcon-ros-cargo installed, since the type string comes from `colcon_ros`.

In the CI logs, the loop body's own `echo "Running clippy in $path"` never appears as runtime output. In run [33553913917](https://github.com/ros2-rust/ros2_rust/actions/runs/33553913917) (main, 2026-09-01) and run [27496259196](https://github.com/ros2-rust/ros2_rust/actions/runs/27496259196) (main, 2026-06-14), each of the humble, jazzy and kilted legs shows `Running clippy in`, `Running cargo test in` and `Running rustdoc check in` exactly once, inside the step's own script listing, and never again. As a positive control the same legs print 107, 114 and 118 runtime `Compiling` lines. (The rolling leg is `continue-on-error` and did not reach these steps in either run.)

Affected lines: `rust-stable.yml:95,114,130,146`, `rust-minimal.yml:95,114,130,146`, `rust-win.yml:79`. The same stale literal is in `rust-win.yml`; I have not run the Windows leg myself.

The staleness is also user-facing in `docs/building.md:110-116`, which tells a reader that seeing

```
rclrs   src/ros2_rust/rclrs   (ament_cargo)
```

means "the `colcon-ros-cargo` plugin works as expected". Someone following that today sees `(ros.ament_cargo)` and has reason to think their setup is broken when it is correct.

## What turning them back on surfaces

Running the workflow's commands verbatim, including the per-package feature branches and the three package exclusions, in a fully built workspace on that container:

| step | packages | red |
| --- | --- | --- |
| clippy | 9 | 4 |
| cargo test | 6 | 1 |
| rustdoc | 6 | 1 |

I did not run the `fmt` step (it needs the nightly toolchain).

Nearly all of it is style debt on `rclrs`. One item is worth a decision rather than a fix: `clippy::let_underscore_future` fires on six `let _ = ...` sites, four of them calls to `ExecutorCommands::run`, whose docstring tells callers to do exactly that ("You can safely discard the promise ... Use `let _ = promise;` to suppress the warning"). The returned `Promise` is a oneshot receiver for a task that `run` has already queued, so the lint is not reporting a dropped task, but `-D warnings` will reject the documented pattern. Either the docstring should suggest a different spelling or the lint needs an `allow` with that explanation.

One red I am not claiming: `cargo test` on `rosidl_runtime_rs` fails to link its test binary with `cannot find -lrosidl_runtime_c`. That may well be my container rather than the project, and CI is the place to find out.

<details>
<summary>Full diagnostic inventory (rclrs, Rust 1.85.0)</summary>

clippy, 137 diagnostics across 32 lint kinds on `rclrs`: 23 elidable lifetimes (`'msg`), 20 very-complex-type, 15 useless conversion, 14 needless deref, 11 needless borrow, 5 redundant pattern matching, 5 `let_underscore_future` (2 of them in test modules), 4 unit let-binding, 4 elidable lifetimes (`'a`), 4 doc list indentation, and 22 kinds with one or two hits each (among them two unannotated `transmute`s, an `unsafe fn` without a `# Safety` section, and a 9-argument function). Three other packages carry one clippy lint each (`unwrap_or` constructing a default, an unnecessary raw-pointer cast, and the sixth `let_underscore_future` in `examples/rclrs/rust_pubsub`).

`let_underscore_future` sites:

```
rclrs/src/action/action_server.rs:376   let _ = self.board.node.commands().run(f);
rclrs/src/action/action_server.rs:425
rclrs/src/node/node_options.rs:410
rclrs/src/subscription.rs:588           (test module)
rclrs/src/worker.rs:820                 (test module)
examples/rclrs/rust_pubsub/src/simple_subscriber.rs:36
```

rustdoc, 8 errors on `rclrs`: unresolved intra-doc links (`crate::spin`, `crate::Node::create_action_client`, `NodeState::create_action_server`, `NodeState::create_action_goal_receiver`) and public docs for `begin` / `begin_cancelling` linking to the private item `CancellingGoal`.

CI uses `dtolnay/rust-toolchain@stable`, so today's stable will report a different, probably larger, set than 1.85.0 does.

</details>

## Proposed next step

I can open a PR that updates the nine selectors and the `docs/building.md` example. Because that will expose the failures above, I am raising it here first so you can decide whether the revived steps should start blocking or `continue-on-error`. Happy to take any part of the cleanup that would be useful.

Environment: `ros:humble` on arm64, Rust 1.85.0 (the version this repo's `Dockerfile` pins), clippy 0.1.85.

Assisted-by: Claude Code:claude-fable-5-1

espressolee:
Fixes #682.

Since colcon-ros-cargo #32 (December 2024), `colcon list` has reported Rust packages as `(ros.ament_cargo)`. The workflows still filtered for `(ament_cargo)`, so their per-package loops selected zero packages and passed without running their checks.

This PR restores those checks as blocking checks:

- selects `ros.ament_cargo` packages and fails explicitly if the selector returns no target packages;
- computes each package-path list once and reuses it;
- limits Clippy and rustdoc to packages owned by this repository with `--packages-select`, rather than linting source imported from separate repositories by the `.repos` file;
- sources the built `install/setup.sh` overlay so generated ROS message crates are available to Clippy and rustdoc;
- keeps `action-ros-ci` as the Linux build/test runner instead of running the same tests twice;
- propagates individual `cargo test` failures in the Windows batch loop; and
- excludes the generated `ros_ws/` directory from the root Cargo workspace.

The restored Clippy step exposed the existing `rclrs` lint backlog. This PR fixes that backlog in a separate automated-fix commit followed by reviewed manual fixes. It also repairs the broken intra-doc links exposed by the restored rustdoc step. The Rolling matrix entry remains allowed to fail at the job level because current Rolling is broken by the upstream change discussed in #658; all non-Rolling matrix entries remain blocking.

Local validation on Ubuntu 22.04 aarch64 with ROS 2 Humble:

- selector fixtures: new selector 1 target; old selector 0 targets; repository target selection 1; dependency closure 2;
- `actionlint` 1.7.7: all three edited workflows pass;
- nightly rustfmt 1.10.0: 1/1 target package passes;
- Rust 1.85.0 and Stable 1.98.1: blocking Clippy and rustdoc both pass for the selected target package;
- Rust 1.85.0: 120/120 unit tests and 51/51 doc tests pass;
- Stable 1.98.1: 120/120 unit tests and 51/51 doc tests pass on the final successful run; and
- colcon build: 13/13 packages pass.

One local parallel Stable run intermittently lost `/rosout` messages while the test capture handler was active. That race is fixed on current `main` by #656. A narrower mutex workaround was tested and then removed in a follow-up commit so this PR does not duplicate the upstream fix. Merge-commit CI against current `main` is therefore the authority for the final cross-distro result.

Assisted-by: Claude Code:claude-fable-5-1

Assisted-by: Codex:gpt-6 cargo-clippy:1.85.0 cargo-clippy:1.98.1 actionlint:1.7.7

### DIFF
--- a/.github/workflows/rust-minimal.yml
+++ b/.github/workflows/rust-minimal.yml
@@ -14,6 +14,10 @@ on:
 env:
   CARGO_TERM_COLOR: always
 
+defaults:
+  run:
+    shell: bash
+
 jobs:
   build:
     strategy:
--- a/.github/workflows/rust-minimal.yml
+++ b/.github/workflows/rust-minimal.yml
@@ -82,6 +82,20 @@ jobs:
         sudo pip3 install git+https://github.com/colcon/colcon-cargo.git
         sudo pip3 install git+https://github.com/colcon/colcon-ros-cargo.git
 
+    - name: Search Rust packages in this repository
+      id: list_repo_rust_packages
+      run: |
+        mapfile -t package_paths < <(colcon list | awk '$3 == "(ros.ament_cargo)" { print $2 }')
+        if (( ${#package_paths[@]} == 0 )); then
+          echo "::error::No ros.ament_cargo packages found in the repository"
+          exit 1
+        fi
+        {
+          echo 'package_paths<<EOF'
+          printf '%s\n' "${package_paths[@]}"
+          echo EOF
+        } >> "$GITHUB_OUTPUT"
+
     # test_msgs recently added ament_mypy as a test dependency, but rosdep
     # may fail to install it when building from source on rolling
     - name: Install ament_cmake_mypy for rolling
@@ -91,13 +105,16 @@ jobs:
         sudo apt-get install -y ros-rolling-ament-cmake-mypy
 
     - name: Check formatting of Rust packages
+      env:
+        PACKAGE_PATHS: ${{ steps.list_repo_rust_packages.outputs.package_paths }}
       run: |
-        for path in $(colcon list --packages-up-to $(echo "${{ steps.list_packages.outputs.package_list }}") | awk '$3 == "(ros.ament_cargo)" { print $2 }'); do
-        cd $path
         rustup toolchain install nightly
-        cargo +nightly fmt -- --check
-        cd -
-        done
+        while IFS= read -r path; do
+          [ -n "$path" ] || continue
+          cd "$path"
+          cargo +nightly fmt -- --check
+          cd - > /dev/null
+        done <<< "$PACKAGE_PATHS"
 
     - name: Build and test
       id: build
@@ -107,48 +124,53 @@ jobs:
         target-ros2-distro: ${{ matrix.ros_distribution }}
         vcs-repo-file-url: ros2_rust_${{ matrix.ros_distribution }}.repos
 
-    - name: Run clippy on Rust packages
-      continue-on-error: true  # revived check: non-blocking until the pre-existing rclrs backlog is paid down
+    - name: Search target Rust packages in the built workspace
+      id: list_workspace_rust_packages
+      env:
+        PACKAGE_NAMES: ${{ steps.list_packages.outputs.package_list }}
       run: |
         cd ${{ steps.build.outputs.ros-workspace-directory-name }}
-        . /opt/ros/${{ matrix.ros_distribution }}/setup.sh
-        for path in $(colcon list --packages-up-to $(echo "${{ steps.list_packages.outputs.package_list }}") | awk '$3 == "(ros.ament_cargo)" { print $2 }'); do
-        cd $path
-        echo "Running clippy in $path"
-        # Run clippy for all features except use_ros_shim (needed for docs.rs)
-        if [ "$(basename $path)" = "rclrs" ]; then
-          cargo clippy --no-deps --all-targets -F default -- -D warnings
-        else
-          cargo clippy --no-deps --all-targets --all-features -- -D warnings
+        mapfile -t package_names <<< "$PACKAGE_NAMES"
+        mapfile -t package_paths < <(colcon list --packages-select "${package_names[@]}" | awk '$3 == "(ros.ament_cargo)" { print $2 }')
+        if (( ${#package_paths[@]} == 0 )); then
+          echo "::error::No target ros.ament_cargo packages found in the built workspace"
+          exit 1
         fi
-        cd -
-        done
+        {
+          echo 'package_paths<<EOF'
+          printf '%s\n' "${package_paths[@]}"
+          echo EOF
+        } >> "$GITHUB_OUTPUT"
 
-    - name: Run cargo test on Rust packages
-      continue-on-error: true  # revived check: non-blocking until the pre-existing rclrs backlog is paid down
+    - name: Run clippy on Rust packages
+      env:
+        PACKAGE_PATHS: ${{ steps.list_workspace_rust_packages.outputs.package_paths }}
       run: |
         cd ${{ steps.build.outputs.ros-workspace-directory-name }}
         . install/setup.sh
-        for path in $(colcon list --packages-up-to $(echo "${{ steps.list_packages.outputs.package_list }}") | awk '$3 == "(ros.ament_cargo)" && $1 != "examples_rclrs_minimal_pub_sub" && $1 != "examples_rclrs_minimal_client_service" && $1 != "rust_pubsub" { print $2 }'); do
-        cd $path
-        echo "Running cargo test in $path"
-        # Run cargo test for all features except use_ros_shim (needed for docs.rs)
-        if [ "$(basename $path)" = "rclrs" ]; then
-          cargo test -F default,serde
-        else
-          cargo test --all-features
-        fi
-        cd -
-        done
+        while IFS= read -r path; do
+          [ -n "$path" ] || continue
+          cd "$path"
+          echo "Running clippy in $path"
+          # Run clippy for all features except use_ros_shim (needed for docs.rs)
+          if [ "$(basename "$path")" = "rclrs" ]; then
+            cargo clippy --no-deps --all-targets -F default -- -D warnings
+          else
+            cargo clippy --no-deps --all-targets --all-features -- -D warnings
+          fi
+          cd - > /dev/null
+        done <<< "$PACKAGE_PATHS"
 
     - name: Rustdoc check
-      continue-on-error: true  # revived check: non-blocking until the pre-existing rclrs backlog is paid down
+      env:
+        PACKAGE_PATHS: ${{ steps.list_workspace_rust_packages.outputs.package_paths }}
       run: |
         cd ${{ steps.build.outputs.ros-workspace-directory-name }}
-        . /opt/ros/${{ matrix.ros_distribution }}/setup.sh
-        for path in $(colcon list --packages-up-to $(echo "${{ steps.list_packages.outputs.package_list }}") | awk '$3 == "(ros.ament_cargo)" && $1 != "examples_rclrs_minimal_pub_sub" && $1 != "examples_rclrs_minimal_client_service" && $1 != "rust_pubsub" { print $2 }'); do
-        cd $path
-        echo "Running rustdoc check in $path"
-        cargo rustdoc -- -D warnings
-        cd -
-        done
+        . install/setup.sh
+        while IFS= read -r path; do
+          [ -n "$path" ] || continue
+          cd "$path"
+          echo "Running rustdoc check in $path"
+          cargo rustdoc -- -D warnings
+          cd - > /dev/null
+        done <<< "$PACKAGE_PATHS"
--- a/.github/workflows/rust-minimal.yml
+++ b/.github/workflows/rust-minimal.yml
@@ -108,6 +108,7 @@ jobs:
         vcs-repo-file-url: ros2_rust_${{ matrix.ros_distribution }}.repos
 
     - name: Run clippy on Rust packages
+      continue-on-error: true  # revived check: non-blocking until the pre-existing rclrs backlog is paid down
       run: |
         cd ${{ steps.build.outputs.ros-workspace-directory-name }}
         . /opt/ros/${{ matrix.ros_distribution }}/setup.sh
@@ -124,6 +125,7 @@ jobs:
         done
 
     - name: Run cargo test on Rust packages
+      continue-on-error: true  # revived check: non-blocking until the pre-existing rclrs backlog is paid down
       run: |
         cd ${{ steps.build.outputs.ros-workspace-directory-name }}
         . install/setup.sh
@@ -140,6 +142,7 @@ jobs:
         done
 
     - name: Rustdoc check
+      continue-on-error: true  # revived check: non-blocking until the pre-existing rclrs backlog is paid down
       run: |
         cd ${{ steps.build.outputs.ros-workspace-directory-name }}
         . /opt/ros/${{ matrix.ros_distribution }}/setup.sh
--- a/.github/workflows/rust-minimal.yml
+++ b/.github/workflows/rust-minimal.yml
@@ -92,7 +92,7 @@ jobs:
 
     - name: Check formatting of Rust packages
       run: |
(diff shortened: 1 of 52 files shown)


espressolee:
The first CI run exposed three follow-up issues:

1. `fmt` reported formatting differences in two `rclrs` import groups, so I applied rustfmt’s output.

2. `clippy` failed before reaching any lints. `action-ros-ci` creates `ros_ws/` inside the repository checkout, which places its packages beneath the root `Cargo.toml` workspace added in #502. Because those packages are not workspace members, Cargo aborts with `current package believes it's in a workspace when it's not`. The same issue affected the test and rustdoc steps.

   I added `exclude = ["ros_ws"]` to the root workspace to keep the generated ROS workspace outside it.

3. Once the loops actually ran, `clippy` also picked up packages from `ros2-rust/examples`, even though this job’s build step does not include them. This caused `message_demo` to fail with E0433.

   I updated the fmt, clippy, test, and rustdoc loops to use the same `--packages-up-to` closure as the build step, so all of these checks operate on the same package set.

With these fixes in place, the next CI run should reach the actual `rclrs` clippy backlog.

espressolee:
The two remaining failures appear to be a Rolling-only flake in in the existing `rclrs` test run.

At exact base `60e829b5`, Rust Stable and Rust Minimal both reached the 120-test `rclrs` binary and terminated with `SIGSEGV` on September 3; later Rolling checks on that same base commit passed. At this PR head, `65ec9ec2`, Stable terminates with `SIGSEGV` and Minimal with `SIGILL` at the same stage. Humble, Jazzy, Kilted, and, and Windows pass. The only Rust source changes in this PR are rustfmt output.

The corrected selectors are working: the logs show `rclrs`, and `rosidl_runtime_rs` where applicable, actually being selected and run. I cannot rerun the failed jobs without repository admin rights.

Could someone with write access rerun the two Rolling jobs? I don't think a code change is justified before seeing this exact head rerun.

.github/workflows/rust-win.yml
```diff
@@ -76,7 +76,7 @@ jobs:
           call C:\pixi_ws\ros2-windows\setup.bat
           cd C:\workspace
           for /f "tokens=1,2,3" %%A in ('pixi run --manifest-path C:\pixi_ws\pixi.toml colcon list') do (
```
azerupi, line 79: From the logs, cargo test fails but the step still succeeds and the failure is silently ignored

.github/workflows/rust-stable.yml
```diff
@@ -124,10 +125,11 @@ jobs:
         done
 
     - name: Run cargo test on Rust packages
```
azerupi: I believe the step "Build and test" lines 102-108 already run the tests. At least tests were being run while this was broken. This means that now we are running tests twice. We should check which one we want to keep and avoid running tests twice.

.github/workflows/rust-stable.yml
```diff
@@ -108,10 +108,11 @@ jobs:
         vcs-repo-file-url: ros2_rust_${{ matrix.ros_distribution }}.repos
 
     - name: Run clippy on Rust packages
+      continue-on-error: true  # revived check: non-blocking until the pre-existing rclrs backlog is paid down
       run: |
         cd ${{ steps.build.outputs.ros-workspace-directory-name }}
         . /opt/ros/${{ matrix.ros_distribution }}/setup.sh
-        for path in $(colcon list | awk '$3 == "(ament_cargo)" { print $2 }'); do
+        for path in $(colcon list --packages-up-to $(echo "${{ steps.list_packages.outputs.package_list }}") | awk '$3 == "(ros.ament_cargo)" { print $2 }'); do
```
azerupi: This is being repeated multiple times, is it worth extracting into a step that adds the list of packages in an output that other steps can reuse?

azerupi (commented):
Thank you for the PR!

My personal opinion is that we shouldn't merge this with `continue-on-error: true`. We should take the time in this PR to fix all of the clippy lints that make CI fail and land this together. Merging this with `continue-on-error: true` would now run the steps but hide the failures, this would be only marginally better than now.

We / I can help you get this complete if you need assistance. Some clippy lints can probably automatically be fixed by clippy itself with [`cargo clippy --fix`
](https://github.com/rust-lang/rust-clippy#automatically-applying-clippy-suggestions). To make review easier, make sure you commit automated fix passes like this separately.

@esteve @maspe36 what are your thoughts on this?

@espressolee about rolling CI, it is broken in main because of an upstream change see https://github.com/ros2-rust/ros2_rust/pull/658#issuecomment-5509343686. You will not get rolling CI to pass.

espressolee:
Thanks for the detailed review. I reworked the PR around the blocking checks rather than keeping the step-level `continue-on-error` flags.

- `9a5cdda` is the separate `cargo clippy --fix` pass.
- `d3b3144` resolves the remaining Clippy diagnostics under both Rust 1.85.0 and Stable 1.98.1, and fixes the rustdoc link failures.
- `71537a4` removes the duplicate Linux `cargo test` loops, keeps `action-ros-ci` as the Linux test runner, sources the built overlay for Clippy/rustdoc, computes package paths once, fails on an empty selector, and makes Windows `cargo test` failures propagate.
- The post-build checks use `--packages-select` for this repository's target packages. Using the dependency closure also linted the separately imported `rosidl_runtime_rs` repository, whose current main has an unrelated `unnecessary_cast` lint. That external source is still built/tested by `action-ros-ci`; it is not silently allowed in this PR.

The only remaining `continue-on-error` is the matrix-level expression for Rolling, following the known upstream breakage you pointed out. All non-Rolling matrix entries remain blocking.

Local Humble validation is recorded in the updated PR description. In particular, both Rust toolchains pass blocking Clippy and rustdoc; Rust 1.85.0 and Stable each completed 120/120 unit tests and 51/51 doc tests. `actionlint` passes all three edited workflows. The selector fixture gives 1 target for `(ros.ament_cargo)` and 0 for the old `(ament_cargo)` spelling.

Local stress also reproduced the `/rosout` race fixed on current main by #656. I briefly tested a mutex workaround, then fully reverted it in `a2948d1` so the final diff does not duplicate #656. The new CI run will provide the merge-commit result against current main.

.github/workflows/rust-win.yml
espressolee: Fixed in `71537a4`: the Windows batch loop now checks `errorlevel` immediately after each `cargo test`, returns a failing step on any package failure, and also fails if the selector matches zero target packages.

.github/workflows/rust-stable.yml
espressolee: Agreed. I removed the direct Linux `cargo test` loops. `action-ros-ci` remains the single Linux build/test runner; the separate per-package steps now cover only Clippy and rustdoc.

.github/workflows/rust-stable.yml
espressolee: Done in `71537a4`. Repository and post-build Rust package paths are each computed once into step outputs and reused. Both selector steps fail explicitly if they produce an empty target set.

.github/workflows/rust-win.yml
espressolee: Fixed in : the Windows batch loop now checks  immediately after each , returns a failing step on any package failure, and also fails if the selector matches zero target packages.

.github/workflows/rust-minimal.yml
```diff
@@ -14,6 +14,10 @@ on:
 env:
   CARGO_TERM_COLOR: always
 
+defaults:
+  run:
+    shell: bash
```
esteve, line 19: Why?

esteve (changes requested):

.github/workflows/rust-stable.yml
esteve: I prefer readability over reducing repetition if the solution is too complex or not easy to read, it's only a handful of packages.

esteve:
@espressolee please run the tests locally before pushing any more commits.

espressolee:
You're right. I did run the Humble checks locally on aarch64 (Rust 1.85 and Stable Clippy/rustdoc, 120 unit tests, 51 doc tests, and colcon build/test), but that did not cover GitHub's x86_64 `c_char` or the container job's default shell. I should not have treated that as sufficient coverage.

I will not push any more commits until I have reproduced the current PR merge against `main` in an x86_64 container and run the workflow checks locally. Sorry for the churn.

esteve:
@espressolee are we talking to a person or to an AI agent? Just don't be a [meat proxy](https://gruhn.me/blog/2026-08-03/), you're putting the burden of AI-generated code and issues on humans without even checking what you're submitting.

espressolee:
Local x86_64 results, with no further commits pushed after your request:

- I reconstructed the PR merge against current `main`. Its tree is `c443a9b8ad3f2b2492c1af1276388f85b41ab628`, which matches the tree of GitHub's merge commit `6dc9629` exactly.
- Environment: `linux/amd64`, Ubuntu 22.04, ROS 2 Humble, Rust 1.85.0 and Stable 1.98.1.
- `colcon build --packages-up-to rclrs`: 13/13 packages passed.
- Blocking Clippy and rustdoc: passed with both Rust 1.85.0 and Stable 1.98.1.
- Parallel `rclrs` unit tests on the PR merge: 0/5 passed; all five processes ended with SIGSEGV.
- The same parallel test command on current `main` (`2766eab`) in the same container setup: 0/5 passed with the same SIGSEGV.
- PR merge with `--test-threads=1`: 120/120 passed.
- `test_logging_macros` alone passed, the three logging tests passed 10/10, and excluding `test_logging_macros` still crashed in 4/5 runs.

So the native parallel crash is reproducible on current `main` and is not introduced by #684. I am keeping the PR at remote head `f4044d5` while I finish separating that base failure from the checks this PR restores.

espressolee:
This is my account, and I am using Codex as an AI coding assistant. Some of the code and text are AI-assisted, as disclosed in the commit messages and PR description. I am responsible for what was pushed.

Your criticism is fair. I allowed iterative generated fixes to reach CI before adequate x86_64 local validation, which shifted review and debugging burden onto maintainers. I have stopped pushing and will not ask you to diagnose it. I will review and validate the work offline; if I cannot reduce it to one minimal, locally verified update, I will close the PR.

PR #685, opened 2026-09-03
feat!: expose the QoS profile in TopicEndpointInfo

peterisho:
rmw_topic_endpoint_info_t contains the endpoint's QoS profile. rclrs drops it during conversion. This PR adds it to TopicEndpointInfo as a qos_profile field, using the existing From<&rmw_qos_profile_t> impl.

rclcpp and rclpy already expose this field. It lets a subscriber match a discovered publisher's QoS

Marked feat! because the new field breaks literal construction of the struct.


azerupi (approved):
Thank you! Looks good to me

esteve (approved):
@peterisho thanks!

PR #688, opened 2026-09-05
feat(parameters): parameter conversion

azerupi:
The code changes in this PR have been assisted by Claude Code but I have reviewed and iterated on the code.

## Problem

A parameter's conversion was defined only through trait resolution. `ParameterVariant` supplied the kind and conversion in both directions, and every read, write and declaration went through the type's conversion.

But this has 2 problems:

1. The orphan rule makes it hard to use types from third-party crates as parameters because the trait `ParameterVariant` lives in rclrs and only rclrs can implement it for foreign types. This means that any type from the std or other crates that doesn't have a `ParameterVariant` trait implementation in rclrs can't be used by the user.
2.  Some types can have multiple ways to convert them. For example `std::time::Duration` could be stored in the parameter as secodns (`f64`) or milliseconds (`i64`) or even another representation. But you can only have one implementation of `ParameterVariant`.

Both of those cases required the user to make newtypes and implement `ParameterVariant` for them. 

## Solution

The conversion becomes a value that a declaration carries, rather than a trait implementation the type must have. 

```rust
let seconds = ParameterConversion::double(
    Duration::as_secs_f64,
    Duration::try_from_secs_f64,
);

let timeout = node.declare_parameter_with("timeout", seconds)
    .default(Duration::from_millis(500))
    .mandatory()?;

let d: Duration = timeout.get();    // the handle is in the field's own type
```

Nothing about the existing API changes. `declare_parameter::<T>(name)` is now `declare_parameter_with(name, ParameterConversion::of_variant::<T>())`, and a type with a `ParameterVariant` impl declares exactly as before.

`get_with` and `set_with` give `use_undeclared_parameters()` the same reach, so a parameter whose type has no `ParameterVariant` is readable and writable there as well:

```rust
let undeclared = node.use_undeclared_parameters();
undeclared.set_with("timeout", Duration::from_secs(3), &seconds)?;
assert_eq!(undeclared.get_with("timeout", &seconds), Some(Duration::from_secs(3)));
```

## Bug fix

`ParameterKind` does not identify a Rust type. Several types share one, and a value of the right kind can still be unrepresentable in the declared type, for example `70000` is a perfectly good `Integer` but not a `u16`.

In current main the parameter write paths only compared kinds, while every getter assumes the stored value converts to the Rust type. This means that setting a bad parameter could succeed and make the node panic. It is even worse when you think a parameter can be changed externally to the process. 

So right now in main, a remote `SetParameters` call could store a value that passed validation and then panicked the node on the next read. And while this can't happen if we use the standard ROS types as parameters it is reachable today by anyone implementing the **public** `ParameterVariant` trait for a type whose `TryFrom` is partial. And this bug becomes unavoidable once narrow integers are used as parameters, which is a feature I want to add in a later PR.

The declaration's conversion *is* the check that needs to pass, so in the setters we call this function in addition to the kind check:

```rust
fn type_check_of<T: 'static>(conversion: ParameterConversion<T>) -> ValidateCallback {
    Arc::new(move |value| conversion.from_value(value.clone()).map(|_| ()))
}
```

`ParameterValueError::Invalid` carries the reason the conversion gave, and the parameter service puts it in the response, so a user running `ros2 param set` sees why a value was refused.


azerupi:
Superseded by #707. GitHub auto-closed this PR as merged while the stack was being reordered to move it to the bottom (its head branch briefly became an ancestor of its then-base `azerupi/params/rust-ranges`), and a merged PR can't be reopened.

#707 is the same branch with the same three commits, now based directly on `main` and rebased onto current `main` so CI is green. The stack is now #708: #707 → #687 → #690 → #691 → #692 → #693 → #699 → #700.

Sorry for the noise — the discussion above is still the reference, and I'll carry any outstanding points over to #707.

PR #711, opened 2026-09-20
feat!: take the ROS shim interfaces from ros-env

azerupi:
The stubs in `rclrs/src/vendor.rs` were added when the crate moved to `ros-env` for its messages, and in that same change every call site switched from `crate::vendor::` to `ros_env::`. Nothing has referred to them since. When building the with `use_ros_shim` it fails to resolve `ros_env::rcl_interfaces` at all.

Drop the vendored stubs and let `ros-env` supply the interfaces under its own `use_ros_shim` feature, so the module paths are the same whether they come from the shim or from a sourced workspace.

This depends on either https://github.com/ros2-rust/ros-env/pull/14 or https://github.com/ros2-rust/ros-env/pull/15 and a new version of ros-env (0.3.0) being released.

Assisted-by: Claude:claude-opus-5 [Claude Code]


esteve (approved):

esteve:
I've triggered anoter CI run, I'll merge this PR once CI passes.

